PR #14, opened 2015-11-27
Features/whatsnew

uvchik:
I would like to add a "what's new" section to the documentation. I got the idea from the pvlib.

Now everybody can add a line to the file of the next version, if something is added, changed enhanced,... 

@oemof/oemof-main : What do you all think about it?


uvchik:
closed in favour of #15

Issue #36, opened 2015-12-23
outputlib - basic plots


ckaldemeyer:
Uwe has already started with his outputlib and created a method which creates a dataframe with all component timeseries arround a given bus.

He started with basic matplotlib which has all configuration options but in my opinion sometimes to many for standard plots as it is overwhelming...

After some trying my idea would be to go with the pandas basic plotting functions (based on matplotlib) for the basic plotting functions and the plots for renpass-gis.

Here's a small example with a handfull of plots and some possible configuration options beyond the standard functionalities (needs matplotlib >= 1.4):

    import pandas as pd
    import numpy as np
    import matplotlib.pyplot as plt
    import matplotlib as mpl
    from datetime import datetime as dt
    mpl.style.use('ggplot')
    
    # Generate sample data
    sample_data = np.random.rand(24*365, 5)
    df = pd.DataFrame(sample_data,
                      index=pd.date_range('1/1/2015 00:00',
                                          periods=len(sample_data), freq='H'))
    
    # Select date range to plot
    date_from = dt(2015, 12, 22, 0, 0)
    date_to = dt(2015, 12, 22, 23, 0)
    df = df.loc[date_from:date_to]
    
    # Plotting
    # Formatting-tuple (title, colormap, xlabel, ...)
    # for matplotlib.axes.AxesSubplot object could
    # be passed by kwargs later
    df.plot(kind='line', colormap='Spectral', title='Line Plot', linewidth='2')
    [(ax.set_ylabel("Power in GW"),
     ax.set_xlabel("Date and Time"),
     ax.legend(('Wind', 'PV', 'Biomass', 'RoR', 'Demand'), loc='upper right'))
     for ax in plt.gcf().axes]
    
    df.plot(kind='bar', stacked=True, colormap='Greens', title='Bar Plot')
    [ax.legend(loc='upper right') for ax in plt.gcf().axes]
    
    df.plot(kind='barh', stacked=True, colormap='Oranges', title='H-Bar Plot')
    [ax.legend(loc='upper right') for ax in plt.gcf().axes]
    
    df.plot(kind='area', stacked=False, alpha=0.5, colormap='Spectral',
            title='Area Plot')
    [ax.legend(('Wind', 'PV', 'Biomass', 'RoR', 'Demand'),
               loc='upper right') for ax in plt.gcf().axes]
    
    df.plot(kind='box', colormap='Reds', title='Box Plot')
[...]

PR #46, opened 2016-01-08
Features/outputlib based on pandas uvchik

uvchik:
I started to make bus_plot to a default plot function as discussed in #36. I think stackplot is a good example to test if bus_plot can be used in higher plot functions.

The following comments are connected to the commits use the link in the header of the comment.

### DIFF
--- a/oemof/outputlib/to_pandas.py
+++ b/oemof/outputlib/to_pandas.py
@@ -175,7 +175,6 @@ class EnergySystemDataFrame:
         kwargs.setdefault('tick_distance', 24)
         kwargs.setdefault('subplots', False)
         kwargs.setdefault('colormap', 'Spectral')
-        kwargs.setdefault('mpl_style', 'ggplot')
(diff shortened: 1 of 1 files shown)


uvchik:
Please merge (partly merge) if you agree.

simonhilpert:
Can we soon merge outputlib into dev for use in other feature branches as well?

uvchik:
I think we should clarify the last things about the api (optional and obligate parameters) before we merge it into the master but you can merge this branch into your feature branch to use the plotting class.

I would like to make `date_to` and `date_from` an optional parameter and plot the whole timeset if no limit is given. What so you think?

ckaldemeyer:
I agree!

ckaldemeyer:
Oh, I have just read my mails and saw your pull request! Meanwhile I have removed some bugs and added some feature. See #36 or repo for changes.

Maybe a short web conference tomorrow morning would be faster to clear all remaining questions!

Issue #28, opened 2015-12-10
Creating a logging file


uvchik:
With commit 37afe31b one can use a helping function to initialise a logger within oemof.

Just import the logger module and initialise the logger at the very beginning of your app.

```python
# Import logging module
from oemof.tools import logger

# initialise the logger
logger.define_logging()
```

Now everything higher than info will be displayed on the screen but everything (including debug) will be written to a logfile in the folder: ~/.oemof/log_files/

Of course you are able to customise this behaviour. For example if want just warnings and errors but no infos on the screen.

The screen output looks like this now:
```python
11:40:24-INFO-Path for logging: /home/uwe/.oemof/log_files
11:40:24-INFO-Used oemof version: 18345077 @ features/two_region_example
11:40:25-INFO-Processing region: Landkreis Wittenberg (LanWit)
11:40:26-WARNING-Something odd happened...
```

The file output looks like this:
```python
2015-12-10 11:41:14,628 - INFO - logger - Path for logging: /home/uwe/.oemof/log_files
2015-12-10 11:41:14,628 - INFO - logger - Used oemof version: 18345077 @ features/two_region_example
2015-12-10 11:41:14,693 - DEBUG - tools - Getting polygon from DB
2015-12-10 11:41:14,694 - INFO - two_regions_example - Processing region: Landkreis Wittenberg (LanWit)
2015-12-10 11:41:15,591 - DEBUG - models - Retrieving cp values from cp_values.hf5
2015-12-10 11:41:16,987 - WARNING - models - Something odd happened...
```

@oemof/oemof-main : Do we agree that we want to use logging messages within the code? I like the fact that you have a protocol in a file. This can help a lot while debugging. But than we need meaningful logging messages.

uvchik:
No feedback? So I will close the issue and use. Reopen if anything is unclear.

PR #56, opened 2016-01-19
Replace print by logging.info

uvchik:
Therefore we have to be sure that  at least a basic logger is defined otherwise there won't appear anything on the screen. So I defined a very basic logger to see the message on the screen.

It is still possible to use the default oemof logger described  in issue #28.

The advantage is that you can see the message on the screen and(!) in the file. Furthermore it is configurable so that you can decide what level you want to see, and what level you want to log in a file.

 See the following link for more information: https://docs.python.org/3/howto/logging.html

@oemof/oemof-main : Is it okay to merge it?

### DIFF
--- a/doc/api/oemof.tools.rst
+++ b/doc/api/oemof.tools.rst
@@ -1,0 +1,62 @@
+oemof.tools package
+===================
+
+Submodules
+----------
+
+oemof.tools.config module
+-------------------------
+
+.. automodule:: oemof.tools.config
+    :members:
+    :undoc-members:
+    :show-inheritance:
+
+oemof.tools.db module
+---------------------
+
+.. automodule:: oemof.tools.db
+    :members:
+    :undoc-members:
+    :show-inheritance:
+
+oemof.tools.entities_from_csv module
+------------------------------------
+
+.. automodule:: oemof.tools.entities_from_csv
+    :members:
+    :undoc-members:
+    :show-inheritance:
+
+oemof.tools.helpers module
+--------------------------
+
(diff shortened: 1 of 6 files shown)


caro-rli:
I'm ok. Now I can't also see any logging from the solver, can I?

simonhilpert:
do you mean from solph module or solver information? You can print out solver information in 
the console by setting verbose=True in the simulation object. 

logging in solph is something that needs to be coded...

uvchik:
I know that I can use the verbose level to print out the solver message. I do not like the `print` command because everything is gone if I run anything else. With the logging command I can display the information and write them to a file. I would like to change the behaviour of the verbose mode. I would like `verbose=True` to lift up the message to a higher logging level.

`verbose=False`: logging.debug('Solver output.....')
`verbose=True`: logging.info('Solver output.....')

You can see the changes on commit b480241.

I am open for other suggestions.

simonhilpert:
I think thats good.

simonhilpert:
ok i will push some changes in a few minutes

simonhilpert:
Ok I pushed.

uvchik:
@simonhilpert wrote:
> Do we merge use logging for the next release? I have made some changes on the branch (that are actually not to be placed there, and I do not know why I did this - but I did)

I would merge it for the January release if there are no objections.

ckaldemeyer:
(y) Sounds good as we also want to use it in renpass-gis.

gnn:
Sounds good to me. :+1:

PR #65, opened 2016-01-25
Add a time-depended upper bound for the output of a component

uvchik:
For the investment I added an extra function block. I made some test and I couldn't see a time difference if I used the second block for all components. That means that I could remove all the if-clauses and just use it every time, but i didn't have the guts to do it. If it is okay for you I will. The code will be clearer and easier to read.

```python
           # constraint for additional capacity
            def add_output_rule(block, e, t):
                lhs = model.w[e, model.O[e][0], t]
                rhs = ub_out[e][model.O[e][0]][t] + block.add_out[e]
                return(lhs <= rhs)

            # constraint for additional capacity
            def add_output_rule_time_depended_bound(block, e, t):
                lhs = model.w[e, model.O[e][0], t]
                rhs = ub_out[e][model.O[e][0]][t] * (
                    1 + block.add_out[e] / out_max[e][model.O[e][0]])
                return(lhs <= rhs)

            if exist_ub_out:
                block.output_bound = po.Constraint(
                        block.indexset,
                        rule=add_output_rule_time_depended_bound)
            else:
                block.output_bound = po.Constraint(
                        block.indexset, rule=add_output_rule)
```


uvchik:
I merged the branch for another reason.

@simonhilpert : What do think we should about the constraints above?

simonhilpert:
If it's merged and it works as intended, I say we keep it this way. Did you exmaple_apps etc?

uvchik:
> If it's merged and it works as intended, I say we keep it this way.

I just thought we could remove one if clause and make the code cleaner but I'm okay with leaving it as it is.

> Did you example apps ?

So far you can find an example in the reegis_hp repository but examples and test is something that we need to discuss.

PR #73, opened 2016-01-29
Features/constraint tests

uvchik:
I added a draft how to test constraints built by a specific class. I do not know if it is a smart way, so I would like to get some feedback.

I create an isolated object, build the constraints, write out an lp-file and compare it with an existing lp-file.

As an example a add tests for a simple transformer and a fixed source.

### DIFF
--- a/oemof/solph/optimization_model.py
+++ b/oemof/solph/optimization_model.py
@@ -95,22 +95,7 @@ class OptimizationModel(po.ConcreteModel):
                   if not isinstance(c, cp.Sink)}
 
         # set attributes lists per class with objects and uids for opt model
-        logging.info("Building component constraints.")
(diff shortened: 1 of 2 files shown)


uvchik:
There is also a possibility to compare the output of the blocks using the method `pprint` or `display`.

```python
block.pprint()
```

Therefore I created separate methods. These methods could return the block and then it is possible to check the block within the tests. I do not know if this is necessary if we have the lp-file-tests.

uvchik:
The test command is `nosetests3 --with-doctest --rednose --doctest-options '+ELLIPSIS'`.

The parameter `--rednose` is optional.

uvchik:
I talked to @gnn and @simonhilpert so I merged the test to have them in January release.

PR #84, opened 2016-02-03
Features/no initial soc for storages

uvchik:
As I understand it the first and the last values of the state of charge of storages are already forced to be the same by a constraint. But it is still obligate to define an initial state of charge. I would like to make it optional.

### DIFF
--- a/oemof/solph/linear_constraints.py
+++ b/oemof/solph/linear_constraints.py
@@ -549,7 +549,7 @@ def add_storage_balance(model, block):
         if(t == 0):
             t_last = len(model.timesteps)-1
             expr += block.cap[e, t]
-            expr += - block.cap[e, t_last] #* (1 - cap_loss[e])
+            expr += - block.cap[e, t_last] * (1 - cap_loss[e])
             expr += - model.w[model.I[e], e, t] * eta_in[e]
             expr += + model.w[e, model.O[e][0], t] / eta_out[e]
         else:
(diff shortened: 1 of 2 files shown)


oemof/solph/linear_constraints.py
```diff
@@ -538,15 +539,17 @@ def add_storage_balance(model, block):
...
     def storage_balance_rule(block, e, t):
-        # TODO:
-        #   - include time increment
+        # TODO: include time increment
         expr = 0
         if(t == 0):
-            expr += block.cap[e, t] - cap_initial[e]
+            t_last = len(model.timesteps)-1
+            expr += block.cap[e, t]
+            expr += - block.cap[e, t_last] * (1 - cap_loss[e])
```
uvchik: The only difference can be found on line 552 of the new (green) version. Because I added the self-discharge to the first timestep. Now it will work for years without violating the balance. But than we have to discuss, if we replace `block.cap[e, t_last]` in the lines 543 and 544 of the new version with `block.cap[e, 0]` as `cap_initial[e]` sounds like the state of charge at timestep 0. But first I wanted to change as few as possible.

uvchik:
@caro-rli , @simonhilpert , @ckaldemeyer or @oemof/oemof-main 

A feedback would be nice. The most important question is:

Do we want a self-discharge between the last and the first timestep? We do not have it now but it is part of this PR. If the answer is yes, we should change the initial soc to the first timestep ([see comment above](84#discussion_r51895046)).

caro-rli:
I would say yes to having a self discharge between the first and the last time step, because otherwise it does violate the balance of one year, doesn't it?

ckaldemeyer:
+1

PR #97, opened 2016-02-23
Transformer which raises the temperature

uvchik:
The aim of this feature is to transport energy from one bus to another while the buses may have different temperature level. So if the temperature of the input bus is lower than the temperature of the output bus external energy is used to heat up the flow.

A new bus class has to be created named `HeatBus` that has an attribute for the temperature and the return flow temperature.

Furthermore a transformer is needed that has two inputs (storage-bus, external energy) and one output. The output might be a bus for district heating with temperatures between 80 and 120°C. The storage bus might be a "low" temperature bus with a temperature below 100°C (unpressurised). The external energy might be gas, oil, electricity, etc..


uvchik:
The additional device will heat up the flow and therefore will deliver an amount of energy. This amount depends on the temperature difference between the buses and the temperature difference between the input bus and the return flow temperature.

eq2

eq3

Then the amount of all input flows multiplied with their efficiency will be the output flow. The efficiency of the each will be calculated so that the sum of the efficiency of both flows might be greater than one. 

eq1



```
f=\frac{T^{flow}_{bus,o}-T^{flow}_{bus,i}}{T^{flow}_{bus,i}-T^{return}_{bus,o}}

w_{e,i_{e,1}}(t)=w_{e,i_{e,2}}(t)\cdot f,\forall e,\forall t

w_{e,i_{e,1}}(t)\cdot\eta_{e,i_{e,1}}(t)+w_{e,i_{e,2}}(t)\cdot\eta_{e,i_{e,2}}(t)=w_{e,o_{e}},\forall e,\forall t

```

uvchik:
I merged PR #107, so if I merge this branch the content of `features/slack_components` will be merged, too.

uvchik:
All tests are okay, all examples work. Plausibility test look fine. I will merge this branch.

With this branch `features/slack_components` is merged without commit 0dceeb2.

@simonhilpert : To merge #107 you will have to rebase an solve the merge conflicts and the warnings.

PR #98, opened 2016-02-24
Features/pypower

uvchik:


cswh:
Will add loadflow calculation to oemof. But first I have to adjust it to the new structure.

uvchik:
I accidentally opened this PR. There is not description or goal, so I closed it.

Issue #104, opened 2016-02-26
Adding Slack/Excess components instead of pure pyomo variables


simnh:
I created a feature branch features/slack_components where I implemented the two new components. 

Then slack will be inline with the the general structure because it will be represented as as component in the graph. It already works and I would like to merge soon. 
We only need to decide where to place the new classes. I think on the level below bus/component. 

For users nothing changes except the excess/shortage values will pop up in the result data frame aus bus output/inputs. 

Please comment @oemof/oemof-main

uvchik:
To discuss changes on a feature-branch a pull request is easier than an issue.

uvchik:
I don't know where we can continue the discussion after I accidentally merged the PR #107.

I'm quite unhappy with the actual structure. Couldn't we make `excess` and `shortage` to normal components. In my opinion they are sinks and sources and in my opinion it is clearer if they are not added automatically by the bus.

Actually the class `source.Commodity` can be used as a shortage variable. We could do the same for excess (`sink.Something`) and remove excess and shortage and the automatic creation. 

Of course we can keep it for a while to give people the chance to adapt there apps.

uvchik:
see PR #133

PR #107, opened 2016-02-26
Features/slack components

uvchik:
Continuation of #104.

### DIFF
--- a/oemof/core/network/entities/__init__.py
+++ b/oemof/core/network/entities/__init__.py
@@ -1,5 +1,6 @@
 from .. import Entity
 
+
 class Bus(Entity):
     """
     The other type of entity in an energy system graph (besides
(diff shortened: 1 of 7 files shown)


oemof/core/network/entities/__init__.py
```diff
@@ -46,6 +46,26 @@ def __init__(self, **kwargs):
...
+                             costs=self.excess_costs)
+        return excess
+
+    def create_shortage_slack_component(self):
+        """
+        """
+        shortage = ShortageSlack(uid=self.uid+'_shortage',
+                                 outputs=[self],
+                                 costs=self.shortage_costs)
+        return shortage
```
uvchik: In the long run I would remove this part because it makes it less clear to create a component together with the bus. You could also define a sink object (demand) in this way. The bus will be more complicate and the structure less clear.

I would add a warning so that we can remove it in the long run.

simonhilpert:
I move the creation of components from the bus class to optimization model. This should not be confusing for users anymore. As we (@gnn, @ckaldemeyer) want to refactor solph and the components etc anyway in the near future this implementation should be fine for now. If your test etc work you can merge.

ckaldemeyer:
:+1:

uvchik:
The last commit causes pyomo warnings in my app, so I reverted it in my branch.

```python
WARNING: Element ('bel', 'bel_excess') already exists in set w_index_0; no action taken.
WARNING:pyomo.core:Element ('bel', 'bel_excess') already exists in set w_index_0; no action taken.
WARNING: Element bel_excess already exists in set <class 'oemof.core.network.entities.components.ExcessSlack'>.uids; no action taken.
WARNING:pyomo.core:Element bel_excess already exists in set <class 'oemof.core.network.entities.components.ExcessSlack'>.uids; no action taken.
```

uvchik:
To get the warnings you can use the following example:

```python
# -*- coding: utf-8 -*-
import logging
import pandas as pd

from oemof.solph import predefined_objectives as predefined_objectives
from oemof.core import energy_system as es
from oemof.core.network.entities import Bus

time_index = pd.date_range('1/1/2010', periods=2, freq='H')

simulation = es.Simulation(
    solver='gurobi', timesteps=range(len(time_index)),
    debug=True,
    objective_options={
        'function': predefined_objectives.minimize_cost})

energysystem = es.EnergySystem(simulation=simulation, time_idx=time_index)

bel = Bus(uid="bel",
          type="el",
          excess=True)

logging.info('Start optimisation....')
energysystem.optimize()
```

uvchik:
Couldn't we make the `excess` and the `shortage` to normal components. In my opinion they are sinks and sources and I think it is clearer if they are not added automatically by the bus.

uvchik:
@simonhilpert : To merge this PR you will have to rebase an solve the merge conflicts and the warnings.

Issue #114, opened 2016-03-04
Refactor PostHeating to use it for heat pumps and instant water heater


uvchik:
With PR #97 I added an instant water heater (for whatever reason I called it postheating).

This transformer is defined by the following equations.

![eq2](https://cloud.githubusercontent.com/assets/4760584/13263983/5ff573d6-da6c-11e5-9205-7ed478ad2546.png)

![eq3](https://cloud.githubusercontent.com/assets/4760584/13264322/e1a5aae4-da6d-11e5-8acf-88c46131d44f.png)

Then the amount of all input flows multiplied with their efficiency will be the output flow. The efficiency of each will be calculated so that the sum of the efficiency of both flows might be greater than one (maximum: 2). 

![eq1](https://cloud.githubusercontent.com/assets/4760584/13264001/7ce05d80-da6c-11e5-95f4-7288a208fd1c.png)

Actually the f-factor is calculated in the function [add_postheat_relation](../blob/dev/oemof/solph/linear_constraints.py#L310). But everything we need to know to calculate the factor is already know on the application level. So if we calculated the factor before and make it to an attribute of the `PostHeating` class (we should definitly rename it) it will be more generic. Than we can use the following equation to define a heat pump (both 'eta' would be one in this case):
![eq4](https://cloud.githubusercontent.com/assets/4760584/13530507/fd76c4ea-e221-11e5-821a-59b486fab079.png)

I like it because we can calculate a heat pump, an instantaneous flow heater and other two-input-components with one class.

![fig1](https://cloud.githubusercontent.com/assets/4760584/13531480/1299b878-e227-11e5-86c0-28b9c8d20ccb.png)

@simonhilpert, @birgits  : What do you think?

```
f=\frac{T^{flow}_{bus,o}-T^{flow}_{bus,i}}{T^{flow}_{bus,i}-T^{return}_{bus,o}}

w_{e,i_{e,1}}(t)=w_{e,i_{e,2}}(t)\cdot f,\forall e,\forall t

w_{e,i_{e,1}}(t)\cdot\eta_{e,i_{e,1}}(t)+w_{e,i_{e,2}}(t)\cdot\eta_{e,i_{e,2}}(t)=w_{e,o_{e}},\forall e,\forall t

f=\frac{1}{COP-1}
```

uvchik:
I modified the picture a little bit. `A`, `B` and `C` are the flows outside the transformer.

[...]

PR #116, opened 2016-03-08
Refactor the PostHeating class and its solph functions

uvchik:
The PostHeating class was created to describe a instantaneous flow heater. The new idea is describe a more generic class that describes a general transformer that has two inputs and one output.
This transformer can represent a heat pump, an instant flow heater or any other component with two inputs and one output.

See issue #114 for more information

### DIFF
--- a/doc/whatsnew/v0-0-5.rst
+++ b/doc/whatsnew/v0-0-5.rst
@@ -4,7 +4,7 @@ v0.0.5 (DATE)
 New features
 ############
 
-* 
+* added flexible transformer with two inputs and one output (`issue #116 <https://github.com/oemof/oemof_base/issues/116>`_)
 
 Documentation
 #############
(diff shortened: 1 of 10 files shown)


uvchik:
Commit 897e4b works for me. I do not if it is pythonic to use ```try...except``` but it works for lists, numpy.arrays, pandas.Series, tuples and other array-like types.

@simonhilpert: What do you think?

uvchik:
If it is okay for you I would adapt the docstrings and merge. All tests are okay, all examples work as expected.

I think commit 897e4be touches all transformers, so if you think we should discuss this in a larger group I would merge this branch without commit 897e4be and start an extra issue.

birgits:
I think it is fine this way as it probably barely adds time to creating the problem but adds great flexibility!

PR #118, opened 2016-03-09
Features/api improvement

gplssm:
This PR consists of some API changes that simplifies usage of demandlib.

### Input
List of dictionaries changed to single level dictionary

### Output
demandlib now returns sectoral demands as dataframe columns


gplssm:
@caro-rli could you please modify inputs of additional inputs passed to the demandlib if no annual sectoral demand is specified?

PR #119, opened 2016-03-16
'Investment' blocks in add_opex_fix

SanderNielen:
I think 'False' should be changed to 'True' to obtain the set of investment blocks. Otherwise, the code doesn't match the [documentation](http://pythonhosted.org/oemof_base/api/oemof.solph.html#oemof.solph.objective_expressions.add_opex_fix).

### DIFF
--- a/oemof/solph/objective_expressions.py
+++ b/oemof/solph/objective_expressions.py
@@ -106,7 +106,7 @@ def add_opex_fix(model, block, ref=None):
         if block.uids is None:
             block.uids = [obj.uid for obj in block.objs]
         uids_inv = set([obj.uid for obj in block.objs
-                       if block.optimization_options.get('investment', False)])
+                       if block.optimization_options.get('investment', True)])
         uids = set(block.uids) - uids_inv
 
         opex_fix = {obj.uid: obj.opex_fix for obj in block.objs}


uvchik:
By default the investment mode is set to 'False'. This means that there is only a dispatch optimisation. In the investment mode the solver can build new capacities and will be charged with the annuity of the capital costs (capex). So if you want to use this mode you have to define the capex and set the invest mode to True (e.g.`transformer.Storage.optimization_options.update({'investment': True})`). 

I think we will keep the dispatch mode the default mode. Do you have reasons to change that?

get-method: http://www.tutorialspoint.com/python/dictionary_get.htm

SanderNielen:
To clarify, I do not propose to change the default mode. My comment is about the following line of code (see the 'files changed' tab):
` uids_inv = set([obj.uid for obj in block.objs if block.optimization_options.get('investment', False)])`.
I think this is incorrect, since now the uids of objects that are not in investment mode are selected, while `uids_inv` should contain the other uids.

simnh:
No I think you are wrong: 

get('investment', False)   yields: 

1) False (if the user did not change anything,  because the key 'investment' does not exist until a user specifies it)
 this means  'if False' -> uids_inv = [ ]
2) True (if the user set investment mode to true) 
inside the list comprehension then it turns to 'if True' -> ALL uids of the block are selected for uids_inv

Note that at the moment either all or none of the objects that are instances of the same class (e.g. SimpleTransformer) are set for investment or not. 
Hopefully we will change that in the near future so that every component can either be set to investment or dispatch only mode...

PR #133, opened 2016-04-08
Redesign excess and slack to normal components independet of bus flags

uvchik:
At the moment excess and slack variables are created by bus flags.
In the future they should be normal components that can be attached to buses like any other component. A slack component can be created using a commodity source. An excess component has to be created.


uvchik:
I removed the excess and shortage attributes and the corresponding behaviour from solph. To get shortage and excess objects you can use the source or sink class. A source class, that can be used for shortage already exists (`source.Commodity`).

To create an excess object one can use the `sink.Simple` class. I added an attribute to make it possible to change the bound type. Furthermore the value (`val`) will be set to zero if not set.

If you just need the excess-behaviour for debugging you can change the bound type for an existing sink object to 'min' ('fix' is not set, because it is the default value):
```diff
- demand = sink.Simple(uid="demand", inputs=[bel], val=data['demand_el'])
+ demand = sink.Simple(uid="demand", inputs=[bel], val=data['demand_el'], bound_type='min')
```
This is ugly and plotting the demand is useless but it is a fast way to make an infeasible model less restricted. If you want to plot the excess you should add an excess object:
```python
demand = sink.Simple(uid="demand", inputs=[bel], val=data['demand_el'])
excess = sink.Simple(uid="excess", inputs=[bel], bound_type='min')
```
This will add a sink object that takes everything. You can limit the sink object to an upper bound:
```python
excess = sink.Simple(uid="excess", inputs=[bel], val=sink_limit, bound_type='max')
```
Obviously ```val=zeros, bound_type='min'``` and ```val=infinity, bound_type='max'``` will be the same. The first variant is better because you do not have to define the zeros array, because it is the default value (see above).

uvchik:
@oemof/oemof-solph : What do you think?

uvchik:
I adapted the tests and all examples to the new structure and added a warning for people still using the excess=True/False or shortage=True/False parameter.

I had to merge the dev branch before to get the fixes of the tests.

ckaldemeyer:
> excess = sink.Simple(uid="excess", inputs=[bel], val=sink_limit, bound_type='max')

Correct me if I am wrong:

So far, excess and slack where internally added as pyomo variables whereas `sink.Simple` was added as a pyomo parameter and `source.Commodity` as a variable, right?

This approach somehow does not seem transparent to me and we could also consider making it more explicit.

Wouldn't it be easier and more compelling to just add `inf`as upper bound and create single classes for `slack`and `excess`?

uvchik:
I thought that we do not want to create new classes that do the same that existing classes do.

I just added the possibility to set opex values for the sink class. In future it would be nice to make it possible to set a yearly input limit.

@oemof/oemof-solph : At the moment it is possible to set a yearly output limit for buses. How is this meant to work?

caro-rli:
> excess = sink.Simple(uid="excess", inputs=[bel], val=sink_limit, bound_type='max')

What is the data type of val meant to be? Integer is not working.

caro-rli:
Should have tried a bit before asking. Seems like a numpy array works.

ckaldemeyer:
> Should have tried a bit before asking. Seems like a numpy array works.

Iterables (also lists or dataframe columns) should work

ckaldemeyer:
> I thought that we do not want to create new classes that do the same that existing classes do.

You're right, this option is also not the nicest. I guess it depends on if one thinks from the optimisation or OOP side which option seems better/worse ;-)

> I removed the excess and shortage attributes and the corresponding behaviour from solph. To get shortage and excess objects you can use the source or sink class. A source class, that can be used for shortage already exists (source.Commodity).
To create an excess object one can use the sink.Simple class. I added an attribute to make it possible to change the bound type. Furthermore the value (val) will be set to zero if not set.

What do you think about this proposal?

excess = simple transformer with excess costs that feeds into an (unbalanced) "excess bus"
shortage = simple transformer with shortage costs that  feeds from an (unbalanced) "shortage bus"

This would be of course one more step but is in line with the general concept and prevents redundancies. Additionally, it's explicit

uvchik:
For me it is cleaner to have sinks, sources, transformer and buses. For me a bus is a balanced system and therefore I do not like unbalanced buses. I rather connect a sink or a source to the bus. 

What about the demand. Do you want to model it the same way? Than you will have a lot of "strange" attributes in the bus class. Attributes that belong to sinks and sources.

And finally you do not transform anything that might be confusing, too.

ckaldemeyer:
> @oemof/oemof-solph : At the moment it is possible to set a yearly output limit for buses. How is this meant to work? 

I guess this might be a relict and should either be located in the commodity (which also has a sum out limit) or in the bus.

>  For me it is cleaner to have sinks, sources, transformer and buses. For me a bus is a balanced system and therefore I do not like unbalanced buses. I rather connect a sink or a source to the bus.
What about the demand. Do you want to model it the same way? Than you will have a lot of "strange" attributes in the bus class. Attributes that belong to sinks and sources.
And finally you do not transform anything that might be confusing, too.

I am just not a fan of special cases that have to be explained. But you're right that it might be confusing and I don't have a better idea right now. So I am fine with it.

Concerning the code: Shouldn't the new classes be placed in the "components" subfolder?

So far these are in the `__init__.py` in the entities folder despite being one level lower in the class hierachy.

    class ExcessSlack(Component):
    class ShortageSlack(Component):

ckaldemeyer:
>> @oemof/oemof-solph : At the moment it is possible to set a yearly output limit for buses. How is this meant to work? 

> I guess this might be a relict and should either be located in the commodity (which also has a sum out limit) or in the bus.

I changed my opinion. There might be cases where both is interesting!

uvchik:
This PR will be part of solph v0.1 and will not(!) be merged into actual dev branch.

PR #148, opened 2016-04-27
Features/change plot order

uvchik:
I added a method to change the order of the columns within a DataFrame subset. So people will be able to change the order and use the pandas plotting method afterwards. This is useful for stacked plots.

```python
>>> myplot = DataFramePlot()
>>> myplot.subset.columns.values
['a', 'b', 'c', 'd']
>>> myplot.rearrange_subset(['b', 'd', 'a', 'c'])
>>> myplot.subset.columns.values
['b', 'd', 'a', 'c']
>>> myplot.plot(stacked=True)
```


uvchik:
@oemof/oemof I think this will not touch anybody in a negative way.

@birgits : So I guess you can just merge it before you release oemof on Friday

ckaldemeyer:
Nice feature! From my side :+1:

PR #154, opened 2016-05-10
Features/travis ci

birgits:


gplssm:
Tests pass with if Python 3.3 is omitted. Shall we exclude Python 3.3 from test cases?

gplssm:
After @gnn fixes #155 re-add Python 3.3 to .travis.yml

uvchik:
@gplssm Will you merge it?

gplssm:
Yes, I'll do it, when @gnn gives his OK.

uvchik:
To reduce the server load, we could delete the push-test. I think the PR-test is enough.

gplssm:
I disabled "Build pushes" only "Build PR" is active. If you uncomfortable with this, we can reconfigure it...

uvchik:
> I disabled "Build pushes" only "Build PR" is active. If you uncomfortable with this, we can reconfigure it...

@gplssm : Where did you do it? Did you push it? I do not see a commit.

gplssm:
I disabled in the config on the travis website which is not visible in the commit history.

uvchik:
But there are still two checks (continuous-integration/travis-ci/push and continuous-integration/travis-ci/pr) or do I misunderstand something?

gplssm:
We still see 2 tests here because my changes in the settings are older then the last commit in this PR. Have a look at PR #159, where only one test run is shown.

PR #159, opened 2016-05-11
Features/industrial load profile

birgits:
So far the industrial load profile was part of the class bdew_elec_slp and the parameters for the creation of the step load profile were hard coded. Since the step load profile method is not a BDEW method and it was not possible to solely create an industrial load profile without calling the function that also creates residential and commercial profiles I took the industrial load profile out of that class and created a new class IndustrialLoadProfile. At the same time I added some documentation and removed the hard coded parameters. @oemof/demandlib What do you think?


caro-rli:
Sounds great from your explanation. I do not overlook the code in detail, so I would suggest to merge it if everything works fine. Just two questions:
* Did you run the demandlib test examples and do you get the same results before and after your changes?
* Does it make sense to integrate an example with industrial load profile in the test example? As far as I know it's not included so far.

gplssm:
From my perspective it definitely makes sense to have the industrial profile represented in the example! Does someone got the time to get it done?

Btw, the example than could be implemented as good integration/system test. What do think about that?

uvchik:
@gplssm did some changes on PR #154. As the branch of PR #154 is part of this PR it will be partly merged in an implicit way. The best way would be to merge #154, then pull the actual dev and then merge this PR.

uvchik:
The decider methods looks a little bit strange to me but this is something we could discuss, while creating the new demandlib structure.

gplssm:
Besides #163 it looks fine. Based on the current state we could merge it into dev. Later, we should reconsider design of demandlib in general, as we all know ;)

Some more improvement suggestions:
1. Handle missing `annual_demand` if 'simple_industrial_profile' is given as `method`. Missing `year` is to be covered to by a default value (which could be hard) or an exception.
2. Related to (1). In general, we should consider to more strictly separation of calculation of timeseries based on given annual demands and calculation of annual demands based on statistical data.

uvchik:
> Based on the current state we could merge it into dev.

We definitely have to merge PR #154 before we can merge this branch or we have to revert commit 29870fbb5 before a merge.

PR #160, opened 2016-05-12
Create a structure to use examples as system tests

uvchik:
It is planed to create some representative examples to test the system and the results.


uvchik:
I think the next step is to sort the results and write blocks for the examples that should be tested.

@oemof/oemof-namespace What do you think of keeping relevant test examples and moving everthing else to the oemof-example repository?

@oemof/oemof-namespace What examples to we need?

PR #178, opened 2016-06-16
First refactoring/solph/dev

uvchik:
What are the next steps? Use this PR for wishes and to discuss global aspects or create issues for concrete matters.


uvchik:
Tests work, examples work. Documentation needs support. I will merge it to the release branch.

Jazzpilz:
It looks like the hub.connect() function is broken, in network/__init__.py:
AttributeError: 'WeakKeyDictionary' object has no attribute 'append'

Jazzpilz:
Sorry, I mean oemof.core.energysystem.connect()

uvchik:
You are right. thank you. 

The transport class and the transformer class were the same, so we decided to remove the transport class. To use the invest mode for the transport (in both directions) there might be a transport class in the future. 

Anyhow, the connect method is not needed any more. Use two transformer objects instead.

Jazzpilz:
Ok thank you for the explanation!

PR #179, opened 2016-06-17
Flow groupings has no effect.

uvchik:
At the moment it is good that it has no effect because the InvestmentFlow should also be part of the Flow group. But in the long run it might be useful to understand what happens.

### DIFF
--- a/oemof/solph/blocks.py
+++ b/oemof/solph/blocks.py
@@ -377,13 +377,15 @@ class Flow(SimpleBlock):
 
         m = self.parent_block()
 
-        # ########################## SETS #####################################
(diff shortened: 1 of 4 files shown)


uvchik:
With commit 763c856 and 29308e8 everything seems to work as intended.

Reverting commit 763c856 should break the test but the revert does not have an effect on the tests.
This behaviour could cause problems in the future.

uvchik:
The filter seems to have an effect on the constraints but not on the objective function.

uvchik:
This PR already includes some fixes that should be merged. The error is very theoretical by now. So I merged it.

PR #180, opened 2016-06-24
Add dictionaries to soplh blocks due to performance reasons

simnh:
Calling functions in pyomo rule seems to be significantly slower and more memory
usage compared to prior creation of dicts, that then are used in the rules!!!!


uvchik:
Commit 3534c4bc2e283cbf00af44 does the same for the storage block.

Did you check solph for this issue and fix it everywhere?

simnh:
I merge some things from the nodes-csv-branch. I think before we merge into the release branch,
we should clean up the examples. Sorry for the temporary mess...

uvchik:
> Sorry for the temporary mess...

:smile:  What about my question from above?  Did you check if there are more classes where we have to create a dictionary instead of calling a function? For me the storage example is now faster than before.

> I think before we merge into the release branch, we should clean up the examples.

There are only two examples left:

1. One core example. I added a message there and will remove it until v0.1 if nobody fix it. I think we should leave it with this message for the beta version.
2. The renpass-gis example. (@simnh, @ckaldemeyer). For me it is okay if you add a warning for the beta version and clean it up until v0.1 release.

uvchik:
I talked to @ckaldemeyer to create an example_test for nodes-from-csv but he seems to be very busy.

ckaldemeyer:
> I talked to @ckaldemeyer to create an example_test for nodes-from-csv but he seems to be very busy.

Yep. I probably won't be able to to it in the next 2 weeks as there's quite a lot to do here. And I want to move the extended nodes_from_csv_example (renpassG!S) to the ZNES repository anyway.

Does it make sense to provide a minimal example to show how the csv-reader works?

simnh:
i think so...

uvchik:
I will be on holiday the next two weeks. 

What about marking the example as experimental (nodes_from_csv_experimental) for the beta release and when I'm back, @ckaldemeyer and I will create an reduced example based on the actual version. The full version will be moved to the ZNES repository until v0.1. The reduced example should be part of the example_test to make sure everything will work in the future. Just a proposal. What do you think?

@simnh 
> i think so...

What about my question from above? Did you check if there are more classes where we have to create a dictionary instead of calling a function? For me the storage example is now faster than before.

uvchik:
Commit 3534c4b does the same for the storage block. This is actually the title of this PR and if there is nothing to do we could merge this PR.

ckaldemeyer:
> I will be on holiday the next two weeks.

Enjoy you holiday!

> What about marking the example as experimental (nodes_from_csv_experimental) for the beta release and when I'm back, @ckaldemeyer and I will create an reduced example based on the actual version. The full version will be moved to the ZNES repository until v0.1. The reduced example should be part of the example_test to make sure everything will work in the future. Just a proposal. What do you think?

Sound good to me :+1:

PR #184, opened 2016-07-11
First refactoring/remove core

uvchik:
I removed the core package from oemof as it will not work with the new structure anyway. 

I moved the energy_system to the root path of oemof and the entity class from `core.network` to `network`.

I adapted the imports in solph and in the tests.

@oemof/oemof-core , @oemof/oemof-solph : Any objections?  Please check!


uvchik:
Talked to @gnn and @simnh. They are fine with the changes. All checks passed. Example_tests passed.

Issue #190, opened 2016-07-21
Input and output in pandas DataFrame are inverted in storage_invest example


uvchik:
In the example the output of the natural gas power plant is summed up. The input is requested in line 144 but the output is returned. Which is good, because the output is wanted but it is confusing that one get the input by typing output and the output by typing input. Something seems to be inverted.

The efficiency is 0.58, which is exactly the difference between the input and the output but it should be the other way round. Would be great to fix it soon, because it has a big impact on the results if people are not aware of the this bug.
```
pp_gas_output_sum: 1.943970e+08
pp_gas_input_sum:  1.127503e+08
```

simnh:
If you are talking about this: 

``` pp_gas = myresults.slice_by(obj_label='pp_gas', type='input',
                                date_from='2012-01-01 00:00:00',
                                date_to='2012-12-31 23:00:00')```

If i am correct, this will return the input into the bus (which is omitted, i do not know why) from by the object pp_gas so this will return electricity from gas turbine. 

Note that the `slice_by` method is from the bus perspective. Actually if there where another bus with the input of the object `pp_gas` in the call above, this would be included in the sum as well....

**Comment by simnh** [comment]
So I guess everything should work as it is supposed to work?

**Comment by ckaldemeyer** [comment]
> So I guess everything should work as it is supposed to work?

In my nodes_from_csv Branch everything seems to work as expected, too and I did some other fixes I don't remember now some weeks ago. I guess @simnh merged them into the release branch.

@uvchik : Could you check it again?

**Comment by uvchik** [comment]
> So I guess everything should work as it is supposed to work?

@simnh Kind of. It might not be a bug, but it definitely does not work as expected. The label is `pp_gas` and this implies, that `input` and `output`are referred to `pp_gas`. For me it is way from being intuitive if `obj_label='pp_gas', type='input'` means output from `pp_gas`. I would expect the following.

```python
pp_gas_input = myresults.slice_by(obj_label='pp_gas', type='input',
                                  date_from='2012-01-01 00:00:00',
                                  date_to='2012-12-31 23:00:00')
pp_gas_output = myresults.slice_by(obj_label='pp_gas', type='output',
                                   date_from='2012-01-01 00:00:00',
                                   date_to='2012-12-31 23:00:00')
```

[...]

PR #191, opened 2016-08-10
Fix issue 190

uvchik:
The name of the keyword is not important to me so change it if you want.

I renamed the keyword of the results DataFrame from `input` to `tobus` and from `output` to `frombus` to make it a bit more intuitive. See #190 for more details.

**Before:**

```python
input_into_pp_gas = myresults.slice_by(obj_label='pp_gas', type='output')
output_from_pp_gas = myresults.slice_by(obj_label='pp_gas', type='input')
```

**Now:**

```python
input_into_pp_gas = myresults.slice_by(obj_label='pp_gas', type='frombus')
output_from_pp_gas = myresults.slice_by(obj_label='pp_gas', type='tobus')
```

### DIFF
--- a/examples/example_tests.py
+++ b/examples/example_tests.py
@@ -123,22 +123,22 @@ except Exception as e:
 
 test_results = {
     'objective': 2947725.249402091,
-    ('b_el', 'tobus', 'pp_chp', 'val'): 11161.357450000065,
-    ('b_el', 'tobus', 'pp_coal', 'val'): 33723.047672110595,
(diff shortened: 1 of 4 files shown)


ckaldemeyer:
Thanks for your changes. One small proposal: what to you think about changing it to "to_bus" and "from_bus". I would prefer that since "tobus"/"frombus" might be confusing.

I'll change it if you're o.k. with it

uvchik:
> I'll change it if you're o.k. with it

:+1:

uvchik:
Just did the "find and replace".

ckaldemeyer:
Thanks

Issue #202, opened 2016-10-05
Failed to build the documentation at RTD


uvchik:
According to the [error message](https://readthedocs.org/projects/oemof/builds/4474754/) it seems to be related to [this faq](http://read-the-docs.readthedocs.io/en/latest/faq.html#i-get-import-errors-on-libraries-that-depend-on-c-modules).

The last line of the error message:
```python
ImportError: Building pandas requires cython
```
The error appeared with pandas v.0.19 but I wonder why it didn't before. Using C is nothing new in v0.19.

uvchik:
I added a `rtd_requirements.txt` and made RTD read this file instead of the requirements from the `setup.py`. I changed that in the "Advanced Settings".

The new `rtd_requirements.txt` excludes pandas 0.19 because in the interaction between pandas 0.19 and RTD something does not work.

:heavy_exclamation_mark: pandas 0.19 works fine with oemof, it was just a problem with building the documentation. Thats why I changed the requirements for RTD but not for normal users.

Maybe we should report this to RTD or hope that it will be solved in future pandas versions.

By now it works.

PR #213, opened 2016-10-18
Change relative paths to absolute paths in demandlib

birgits:
When using the demandlib in a virtual environment absolute paths to the data files are needed. So far the paths were relative paths, so I changed that. 
@oemof/oemof Are you okay with my changes? Should we make a new release with these changes or wait for further changes?

I also wasn't sure if using the master branch as the base branch for my PR is correct since we won't merge it directly into the master branch. Should I have created a new release branch v0.0.9 or what is the correct way?


gplssm:
Since you plan to further build your app against v0.0.8 and subsequent versions oemof that are kind of deprecated, I would recommend to start a new branch from the tag v0.0.8. You could name the branch `deprecated/dev` or even `deprecated/release_v0.0.9`. What do other @oemof/oemof people think of the procedure and the naming of the branches?

A further question: is it possible to make an release from other branches then master?

uvchik:
@gplssm I like the idea with `deprecated/dev`, `deprecated/master`, `deprecated/something` and so on. I would release always the `deprecated/master`. Soon the native `master` and `dev` will be used for v0.1.

If you upload to pip the actual copy of your code will be uploaded. Pip does not even know whether you use git. The release tag on github will not be a problem either because you can dedicate it to any branch.

@birgits  I would refer to the existing external demandlib instead of fixing the old internal demandlib, because if you do so you have to fix every bug twice, that may occur in the future. I think there even has been some fixes since we divided the packages.

uvchik:
Sorry, that I messed in your PR, but I wanted to fix the travis_ci configuration so that it shows "passed" again when all test are okay. See #202 for more information about this bug.

birgits:
@uvchik I guess you are right about rather using the external demandlib in order to not fix every bug twice. On the other hand we built our app against an oemof version and now want to run scenarios without having to adapt the actual code too much. I'm not quite sure about the best workflow here... But what would be your suggestion for this PR? Discard it because the demandlib has its own repository?

uvchik:
I would remove the demandlib from oemof (we will do that anyway), change the imports to `from demandlib import ...` and than change the few lines in your application. The API changed a little bit since you started with your app but having a look at the demandlib examples it will not take long to fix it. I guess in the long run you will have less problems going this way.

Alternatively you can copy the demandlib to your app and just change the imports. This will also work even though I would not recommand it.

I would not rely on the integrated demandlib of oemof v0.0.8 because we already decided to remove it, to avoid a  double structure. It just hasn't be done so far.

birgits:
Alright, I'll remove the demandlib from oemof and adapt our app then!

uvchik:
Thank you. Hope it's not to much work. Contact me if you have questions.

What about the branches @gplssm suggested? Do you agree?

`deprecated/dev`,
`deprecated/master`

Will you create them or shall I?

birgits:
I created the deprecated/dev and deprecated/master branches and will delete this branch (features/absolute_paths_demandlib).

Issue #214, opened 2016-10-19
solph.GROUPINGS as default parameter


uvchik:
The solph library does not work properly if `solph.GROUPINGS` is not passed to the `solph.EnergySystem`. Shouldn't we set the solph.GROUPINGS by default for the `solph.EnergySystem`?

simnh:
I think we could do this ...

simnh:
It not as easy as i assumend. For now I added a Error message. If @gnn wants to fix it, he may do it...

uvchik:
It was not a big deal but I had to move the EnergySystem to the groupings module to avoid a circular import. Therefore I renamed `groupings.py` to `system.py` to make it less confusing, that it is now more than just a grouping module. Please add comments at PR #215.

gnn:
Ok, I finally was able to wrap my head around what's happening here. The reason for the fact that solph's `GROUPINGS` are not used by default when instantiating solph's `EnergySystem` had to do with the following problem: what if the user wants to supply his own groupings? Previously he could change `EnergySystem(groupings=GROUPINGS)` to `EnergySystem(groupings=GROUPINGS + CUSTOM_GROUPINGS)` or explicitly switch off solph's `G`s if he knew what he was doing.
Now, if a user starts using just `es = EnergySystem()` and decides to change that to `es = EnergySystem(groupings=CUSTOM_GROUPINGS)`, the behavior of `es` changes drastically, as `GROUPINGS` gets switched off implicitly, which might be very surprising. So IMHO most of the time a user has to go for `EnergySystem(groupings=GROUPINGS + CUSTOM_GROUPINGS)`, but he doesn't really know that because we are "hiding" `GROUPINGS` from him in the default argument.
And if we always prepend `GROUPINGS` to the value of the `groupings` keyword argument, we are taking the option of switching off `GROUPINGS` away from the user.
That means the question is which option we prefer.
In any case I will add a (hackish) solution for the variant presented here which doesn't move `EnergySystem`, tomorrow. I'd like to keep it in network, cause I actually want it to be in network in core to. I really think `EnergySystem` belongs into the network module.

simnh:
thanks for clarifying!

uvchik:
Yes, thanks.

For me the groupings function is advanced stuff. I think users can use solph properly without understanding the inside of solph unless they understand what constraints are build. The do not have to understand how. For these users I would try to lower the entry barrier and avoid confusing stuff.

[...]

PR #215, opened 2016-10-20
Make solph.GROUPINGS a default parameter

uvchik:
According to #214.


simnh:
I dont think taht is actually i really good solution. EnergySystem rather belongs to network...But am not an expert. I think @gnn  should suggest a solution, or at least comment on this.

uvchik:
I would say that the network is part of the EnergySystem as groupings is part of the EnergySystem. So I couldn't say where to put it. Maybe outside both of them, 'plumbing` would be an alternative place without circular imports. Than I could rename it again to groupings.

> I think @gnn should suggest a solution, or at least comment on this.
I agree on that. That's why I created a PR and did not change it right away.

uvchik:
For me it is okay like this. If @simnh agrees you can merge it (solve the merge conflicts :smirk:)

simnh:
go ahead...

uvchik:
@gnn: I fixed the merge conflicts. Please merge if you agree.

In the whatsnew file you wrote about changes to the groupings API but there was no groupings API at the previous version (v0.0.7). Please have a look at the whatsnew file and please merge until tomorrow.

gnn:
The groupings API has been there since version 0.0.5, when it was mentioned in the features of whatsnew. ;) But since we are doing kind of a major release, I can remove the parts taking care of backwards compatibility now.
Also, why did you delete v-0.1.rst without copying the changes I made over to v-0.1.0.rst? Or maybe git did delete that when merging recursively. I gotta keep an eye on those cases. Don't want to loose any work when merging.
And why is `solph.GROUPINGS` still used in the [storage invest example](https://github.com/oemof/oemof/pull/215/commits/b20def518f3885a7f13eae0a6a0af736a78a46fc#diff-2ba9caf8eb44df8151fa8f893b86a82eR56)? I thought the whole point of this PR was to not have to specify it when creating the energy system.
I'm going to fix the two points mentioned. Otherwise, this looks good to me.

uvchik:
> Also, why did you delete v-0.1.rst without copying the changes I made over to v-0.1.0.rst? Or maybe git did delete that when merging recursively. I gotta keep an eye on those cases. Don't want to loose any work when merging.

Sorry @gnn this my fault not a git issue. God confused with the two different entries regarding the grouping issues. Thanks to git your work is not lost (s. below). Do you want me to copy this into the whatsnew file?
```
* Solph's :class:`EnergySystem <oemof.solph.network.EnergySystem>` now
  automatically uses solph's :const:`GROUPINGS
  <oemof.solph.groupings.GROUPINGS>` in addition to any user supplied ones.
  See the API documentation for more information.
```

> And why is solph.GROUPINGS still used in the storage invest example? I thought the whole point of this PR was to not have to specify it when creating the energy system.

Also my fault. You are right that was the point. I changed it in the documentation but forgot to remove it from the examples. With b21d142 I removed it from all examples.

My tests passed and Travis CI seems to be happy, too. Thanks a lot for you work and feel free to merge.

uvchik:
P.S.: I saw you already copied it to the whatsnew file.

uvchik:
@gnn Thank you for fixing it :sunflower:

PR #217, opened 2016-10-24
Clearfiy API oemof core: EnergySystem attribute time_idx -> time_index

uvchik:
@simnh wrote:
> We should fix this at is kind of confusing because inside the examples and the OperationalModel class we use 'time_index' whereas inside EnergySystem class we use 'time_idx'...


uvchik:
Okay, I misunderstood. I'm fine with that.

simnh:
I think it should work now, test are all passed

simnh:
Should we merge?

PR #222, opened 2016-10-25
Allow non linear investcosts

simnh:


uvchik:
Do want to merge this until November 1?

simnh:
Yes why not? This is actually already finished...I like the idea of having some features that are included (if they are not that big and do not affect the normal usage). But I think we do not need to have that extensively documented....

simnh:
we can talk about the naming though, but we already saw that it would be nice to have a module were we can keep stuff that is used by more than one block...

uvchik:
It is fine for me.

> But I think we do not need to have that extensively documented....

Yes, but I think we should document it shortly and write at least one test.

simnh:
ok you are write i guess.

simnh:
Ok, besides test etc. does any one have problems with the internal structure, naming etc. of this feature?

PR #223, opened 2016-10-26
Features/add csv reader example

ckaldemeyer:
I have added a minimal example that should explain the funtionality of the CSV-reader.

Feedback is welcome!


simnh:
I think it could be smaller but for now it's ok

ckaldemeyer:
> I think it could be smaller but for now it's ok

Do you mean the code or the energy system (regions, pps, ...)?

ckaldemeyer:
I have now reduced the code and put the model/plotting into functions. The model function returns the objective value in a dictionary!

uvchik:
Everything is fine now.

PR #224, opened 2016-10-26
Proofreading documentation

ckaldemeyer:
I have just started to proofread the documentation.

Still in progress..feedback welcome!


ckaldemeyer:
I have changed the order of "Why are we developing oemof?" and put it behind "The idea of an open framework" since it is more fluent in my opinion

ckaldemeyer:
One suggestion für the section behind the paragraph "Among other applications, the apps 'renpassG!S' and 'reegis' are currently developed based on the framework." under "The idea of an open framework":

Should we add a section "Application Examples" with a link list and some screenshots here? This might be even more appealing than putting extra work in the blog. And it's central, right in the documentation..

What do you think?

ckaldemeyer:
I would substitute the framework concept-image with this (as SVG): http://pasteboard.co/jyMu5jEhk.png

It is based on Uwes sketches and I turned into monochrome since this doesn't conflict with any possible template color and transports the same information.

Accordingly, I am just changing the text to "libraries", etc. as Uwe proposed.

ckaldemeyer:
I have added this sentence to refer to the new image: "Generally, applications can be developed highly individually by the use of one or more libraries depending on the scope and purpose."

In my opinion it gets clear enough now. If there are no objections, I would substitute the image in the next step.

uvchik:
> Should we add a section "Application Examples" with a link list and some screenshots here? This might be even more appealing than putting extra work in the blog. And it's central, right in the documentation..

I think we can add an "Application Examples" section but I would keep it short. One sentence and a link to the repository of the application.

On top we could add a link to the Wordpress page. I think the documentation is to much text for a "public relation" purpose. For me the Wordpress page should be shiny :smile: a lot of pictures!

ckaldemeyer:
> I think we can add an "Application Examples" section but I would keep it short. One sentence and a link to the repository of the application.

> On top we could add a link to the Wordpress page. I think the documentation is to much text for a "public relation" purpose. For me the Wordpress page should be shiny :smile: a lot of pictures!

I have now substituted the image and added links for HESYSOPT and renpassG!S (which is currently hidden behind dev and does not hold data due to open licensing questions).

If you have more linkable examples (reegis?), just add them!

ckaldemeyer:
My idea for the wordpress-page (gallery) would be to add some images such as in this (internal) presentation of first renpassG!S results (of course translated into english): 

https://www.dropbox.com/s/5zjngvpbte5ysod/MoVe_renpass_gis_ueberblick_erste_ergebnisse_20160909_CK.pdf?dl=0

In a style like this or similar: https://github.com/tum-ens/urbs

uvchik:
To make it easier for others to read the documentation I will merge this PR

PR #228, opened 2016-10-30
Fix "get version number in logging"

gplssm:
- When installed via pip no VERSION file available in
  virtualenvs/oemof__release_v0.1/lib/python3.4/site-packages/oemof/
- Fix is maybe slow due to import of whole oemof package, but it's easy
  (:


uvchik:
I think it is okay like this. I added a parameter to make it possible to skip the version log. If it takes to long or any problems will occur the user is free to skip it.

@gplssm What do you think?

uvchik:
As we will release tomorrow, feel free to merge...

Issue #203, opened 2016-10-07
ResultsDataFrame fails on restored energysystem if storages are involved


gplssm:
Using [a results script](https://github.com/gplssm/europepstrans/blob/8973d059984bf2ab93174f5e40501959f61e5d1b/examples/3regions_invest/plot_3regions_results.py) on results obtained from [this example app](https://github.com/gplssm/europepstrans/blob/8973d059984bf2ab93174f5e40501959f61e5d1b/examples/3regions_invest/3regions_invest_example.py) I discovered odd behavior of `ResultsDataFrame` which is probably a bug.

Whenever `energy_system` obtained from solver is stored via `energy_system.dump()` and `energy_system` contains storage objects `__init__` of  `ResultsDataFrame()` fails to build a DataFrame of results.

```
19:35:55-INFO-Restoring attributes will overwrite existing attributes.
Traceback (most recent call last):
  File "/home/guido/rli_local/oemof/examples/solph/storage_optimization/storage_invest.py", line 235, in <module>
    pp.pprint(get_result_dict(esys))
  File "/home/guido/rli_local/oemof/examples/solph/storage_optimization/storage_invest.py", line 142, in get_result_dict
    myresults = outputlib.DataFramePlot(energy_system=energysystem)
  File "/home/guido/rli_local/oemof/oemof/outputlib/__init__.py", line 208, in __init__
19:35:55-INFO-Check the results
    super(DataFramePlot, self).__init__(**kwargs)
  File "/home/guido/rli_local/oemof/oemof/outputlib/__init__.py", line 90, in __init__
    row['bus_label'] = list(k.outputs.keys())[0].label
IndexError: list index out of range

Process finished with exit code 1
```

This can be reproduced using [storage invest example](https://github.com/oemof/oemof/blob/7b18580402758745eef5d42c7a46917ad96f17b5/examples/solph/storage_optimization/storage_invest.py) when changed to dump `energy_system` between solver and results plotting [as illustrated here](https://gist.github.com/gplssm/0be4c86e67dbb4c159deeac15b89502b).

uvchik:
Might be connected with issue #186.

simnh:
Is this solved?

gplssm:
No, maybe we should move to next release

uvchik:
@gnn is about to fix it.

gplssm:
@gnn please close this one when merging #231

PR #231, opened 2016-10-31
Merge fix for 186 (and hopefully 203 too)

gnn:
I tested this with the storage invest example and it seems to fix the issue there.
Didn't test it with @gplssm's example as I couldn't get this to work, but it should fix this too. Could somebody please confirm that #203 gets fixed by this too?
Also: I ran the tests and none broke on my side.


uvchik:
I added dump and restore to the storage-example-test because we do not have real test so far and I don't think that somebody will write one today. After that I solved the merge conflict. Everything should be okay now.

@gplssm Did it solve your issue?

gnn:
If @gplssm isn't able to check it on such a short notice, I say we still include it in the release. It should fix his issue. If not, we have a good reason to investigate this more thoroughly during the next development cycle.

gplssm:
I'm not able to verify that #203 is fixed by this due to other issues regarding execution I can't solve by now.
But I tried dump&restore with _storage invest_ example and it worked. So, I'm quite sure #203 is fixed as well.

Thanks @gnn !

Issue #235, opened 2016-11-02
Missing File: logging_default.ini


fwitte:
I've just installed the recent version of oemof on ubuntu 16.04 via terminal (pip3 install oemof), the following error appeared while running the simple_least_costs example:

> Traceback (most recent call last):
  File "beispiel.py", line 177, in <module>
    run_simple_least_costs_example()
  File "beispiel.py", line 169, in run_simple_least_costs_example
    logger.define_logging()
  File "/home/francesco/.local/lib/python3.5/site-packages/oemof/tools/logger.py", line 62, in define_logging
    shutil.copyfile(default_file, log_filename)
  File "/usr/lib/python3.5/shutil.py", line 114, in copyfile
    with open(src, 'rb') as fsrc:
FileNotFoundError: [Errno 2] No such file or directory: '/home/francesco/.local/lib/python3.5/site-packages/oemof/tools/default_files/logging_default.ini'

Any idea, how to fix that?

uvchik:
A quick fix is to copy the content of the following link to `'/home/francesco/.oemof/logging.ini`.

https://raw.githubusercontent.com/oemof/oemof/dev/oemof/tools/default_files/logging_default.ini

I will see, if I find the reason for this error.

ckaldemeyer:
He says it has been fixed with you workarround!

uvchik:
The fix will be in PR #236.

PR #236, opened 2016-11-02
fix execution of examples after installation

uvchik:
If you install a clean oemof (v0.1) the execution of the example fails due to two reasons

* Matplotlib is missing, even though we decided to add matplotlib to the requirements we did not do it.
* logging.ini is not created because the default_logging.ini is not copied to the site-packages (s. #235)


uvchik:
I think we fixed it.

PR #238, opened 2016-11-07
Naming consistency of solph examples and change all solvers to 'cbc'

simnh:


uvchik:
@gplssm is it possible to pass an argument to the examples script to allow changing the solver. The default would then be the `cbc`:

```bash
oemof_examples simple_least_costs -s glpk
```

simnh:
I renamed it but i wont rename the branch if thats ok

gplssm:
Yes, it is possible to include an optional argument to specify the solver. I'll do it.

Other wished regarding names of examples and so on?

uvchik:
> I renamed it but i wont rename the branch if thats ok

Certainly.

simnh:
I did it, please check it out and finalize if there is anything still missing.

uvchik:
I will push my additions soon.

uvchik:
What I did:

 * added a `solver` parameter or `**kwargs` to make it possible to change the solver in the examples script (08d25dd)
 * cleaned up the examples according to our developer rules (we should be a good example :smirk:) (184d17a)
 * change the names in the examples script to the new names ( 38fbce6)
 * added a user feedback to the `add_constraints` example because it ended without any output (2ce6889)

I also added the possibility to run the storage_investment example without `matplotlib` (8725bd7). We do not have to do it, but I think in the long run we should remove the matplotlib from oemof's requirements. It is just an example how we can do it, without changing the code a lot.

uvchik:
What about the csv-examples?

 * csv_reader_dispatch :question: 
 * csv_reader_investment :question:

ckaldemeyer:
Yes. But I'll probably come to that until Thursday

ckaldemeyer:
Shall we leave the 'optimization' e.g. in 'storage_capacity_optimization'?

I would vote for yes!

ckaldemeyer:
Ah, already done... Forget my comment...

ckaldemeyer:
I have adapted the csv reader examples and renamed them to 'dispatch' and 'investment'.

Additionally, I have added @uvchik s hint on the modelling techniques (use of sources for pricing) to the dispatch examples and changed the paths for the tests.

Can anyone check if I have changed the paths correctly and did not forget anything?

ckaldemeyer:
And should the csv invest example be integrated in the tests as well?

uvchik:
Okay I fixed it, but the result check for the dispatch example fails since commit 515e2510.
```python
14:12:36-ERROR-R2_storage_phs: 93019.19636 not between 88822.57254397201 and 89000.395512028
14:12:36-ERROR-objective: 2204738892.3982296 not between 2323929476.7974014 and 2328581988.2624617
14:12:36-ERROR-R2_R1_powerline: 2418937.89885 not between 2275711.011 and 2280266.989
```

I just wonder why the efficiency of the some power plants was changed (eg. nuclear power plant 0.325 -> 325) with commit 515e2510.

@ckaldemeyer However, please check which version is right and adapt the results.

I added the investment example to the example tests. As there are no results it is only a run check but no result check. As we see above it is worth adding a result check.

simnh:
this remembers me of maybe introducing checks for flow attributes inside the csv-reade (min=0, max=1)...

ckaldemeyer:
> I just wonder why the efficiency of the some power plants was changed (eg. nuclear power plant 0.325 -> 325) with commit 515e251.

My mistake! I re-installed my system and my libreoffice calc locale was set to german. So just saving the csv file misinterpreted the english decimal sign (.) as a thousand sign..

Thanks for your effort!

ckaldemeyer:
The dispatch example should work now.

uvchik:
@ckaldemeyer It is good to have result tests :smirk: - works.

Could we rename the examples in the examples.py from "investment" to csv-reader-investment? The same for dispatch?

ckaldemeyer:
> @ckaldemeyer It is good to have result tests :smirk: - works.

Definitively!

> Could we rename the examples in the examples.py from "investment" to csv-reader-investment? The same for dispatch?

I am fine with it!

uvchik:
The output of your csv-examples is "Check the results".

I would prefer something like `"The results can be found in {0}".format(result_path)`

BTW: It is strange to add line breaks within logging messages.

```
15:42:03-INFO-Something to say....
15:42:04-INFO-I wanted to add a line break
So I did
15:42:07-INFO-Result are stored....

**Comment by ckaldemeyer** [comment]
I don't mind at all. I'll just fix it..

**Comment by uvchik** [comment]
> I don't mind at all. I'll just fix it..

It is just for users executing the script. They do not see the code and "Check the results" will say nothing to them. 

I am fine now. I would just wait for @gplssm to adapt the example script and then merge it.

**Comment by ckaldemeyer** [comment]
> It is just for users executing the script. They do not see the code and "Check the results" will say nothing to them.

No worries, you are right ;-) I just didn't notice..

I fixed it now in the last commit. Together with an error that was induced by a redundant folder 'scenarios' in the file path.

**Comment by ckaldemeyer** [comment]
Oh, it didn't fix it but just fixed the error message. I'll look at it again..

**Comment by uvchik** [comment]
Okay, I think everything works fine now.

@ckaldemeyer If you change the shape of the results you could remove the following logging message;
```
11:32:20-INFO-Read the documentation (outputlib) to learn how to process the results.
```

simnh:
Ok, I will merge for now but we might add some text to the whats new? (maybe)

PR #239, opened 2016-11-07
Features/Inputlib with csv_tools

simnh:
I like the idea of an solph toolbox. One first feature is a timesampling functionality for the csv seqence files. Please give feedback on structure and naming etc.


ckaldemeyer:
I quite like the idea of having such a feature and might also use it at some stage. But I am not sure about the best place for it.

As the whole thing is closely related to the csv-reader and the csv-reader could get a facelift at some stage, too (subdivision in functions, multi-indexing directly from columns, more comments to make it easier to understand,...), we could also think about bringing different chunks together.

During the oemof-paper I also noticed the -yet not realized- pre- and postprocessing in the oemof-core. From my point of view both, the csv-reader and the resampling functionality could be some sort of pre-processing, either in the core or -to be realistic- solph-specific.

What do you think?

simnh:
I had similar thoughts on the location regarding the connection to the csv-reader and its underlying csv-sequence file structure. I also like the idea of an preprocessing module? and we could move away the csv reader from plumbing. This location does not make sense really anyway...as core development is kind of stuck, we can do it in solph and maybe later identify common structures with core to move some functionalities to core.

ckaldemeyer:
> I had similar thoughts on the location regarding the connection to the csv-reader and its underlying csv-sequence file structure. I also like the idea of an preprocessing module? and we could move away the csv reader from plumbing. This location does not make sense really anyway...as core development is kind of stuck, we can do it in solph and maybe later identify common structures with core to move some functionalities to core.

Sounds reasonable to me. As long as the functionality is encapsulated (as you did it now), it should be easy to migrate and adapt since it's mainly only moving files and chaning imports. Maybe we can restructure this already for the next release..

simnh:
so does it maybe make sense to have a preprocessing / postprocessing package ? Our do why create a inputlib package where we move csv-reader and new functionalties as a module 'tools'/'toolbox' ?

uvchik:
I would start with a `csv-reader` class in a `csv-tools`, `input-tools` or just `tools` module until we do have anything else. But I do not have a strong preference.

ckaldemeyer:
> so does it maybe make sense to have a preprocessing / postprocessing package ? Our do we create a inputlib package where we move csv-reader and new functionalties as a module 'tools'/'toolbox' ?

My first idea was to create an inputlib package in accordance to the outputlib package and put all the stuff there. We already have two modules now and further applications could be type day clustering, etc.

> I would start with a csv-reader class in a csv-tools, input-tools or just tools module until we do have anything else. But I do not have a strong preference.

I don't mind as well. We could also think this independently from the csv-reader and solph and just create a packages "tools-for-energy-system-time-series-manipulation" where different functionalities can be found. So basically just a group of functions for different purposes where you put some iterable/sequence in, and something pops out depending on the passed paramset.

This would also enable other users apart from solph to use it..

simnh:
I like the input lib idead, but I think it would be handy to have these functionalities inside solph, maybe later oemof, but not in general general, because it builds upon the internal structure of (oemof) data. And then its get useful. 

I propose: 

inputlib package inside solph containing a csv-tools module where i will place the csv reader and the resampling functionalities. 

?

uvchik:
@simnh I'm fine with that and as I understand it will not break the API which is important until 0.2.

simnh:
yes you are, because NodesFromCsv is in the solph  init anyway...

ckaldemeyer:
> I like the input lib idea, but I think it would be handy to have these functionalities inside solph, maybe later oemof, but not in general general, because it builds upon the internal structure of (oemof) data. And then its get useful.

Yes, but it also has to be maintained and is basically applicable to general time series resampling. I quite like the idea from the linux development "One program, one task." and just think that one could also go in that direction if we think of oemof as a framework for many things and not only energy system optimization. This would be more flexible but less convenient.

So we can also decide to include it in solph and go for convenience.

ckaldemeyer:
Maybe we just follow your proposal after all our "genericly generic redesigns" for potential new users that in the end just came out as solph users... ;-)

simnh:
> optimization. This would be more flexible but less convenient.
But we can also decide to include it in solph and go for convenience..

I think your are right, but maybe we can start like this and get more generic later on

simnh:
I altered the resampling function and added some attributes. Now, I think it should be possible to use it 
for every csv-file that has a datetimeindex format in the first column and some columns with data...header might need to be specified. 

So I think it can be merged. We can even think of adding the resample functionality to oemof core at some point...

ckaldemeyer:
> I altered the resampling function and added some attributes. Now, I think it should be possible to use it 
for every csv-file that has a datetimeindex format in the first column and some columns with data...header might need to be specified.

> So I think it can be merged. We can even think of adding the resample functionality to oemof core at some point...

:+1: I think this is a good way to go!

ckaldemeyer:
We just have to remember to adjust all examples for the imports

simnh:
> We just have to remember to adjust all examples for the imports

This is no problem, as the examples use oemof.solph.NodesFromCSV and this is changed inside the solph __init__.py. So, all examples already work fine. Or is there anything else?

uvchik:
> We just have to remember to adjust all examples for the imports

No, it is very important, that we do not change the examples. Following our decision from the last developer meeting it is not allowed to change the API until `v0.2`. That means that all examples have to work with v0.1.0 as well as with v0.1.x.

ckaldemeyer:
Ah, o.k...

ckaldemeyer:
I forgot about the init.py, this was a nice idea

simnh:
We will add documentation in the new pull request #242

uvchik:
I do not know if it is a good idea to separate documentation, whatsnew, test from the feature. In my opinion a feature PR should contain at least one test, documentation and whatsnew according to our [developer rules](http://oemof.readthedocs.io/en/stable/developing_oemof.html#collaboration-with-pull-requests).

PR #247, opened 2016-11-10
Revise Example

uvchik:
If new users will use oemof, the example are one important entry to the oemof cosmos.

In addition to PR #238 the revision of the example should be completed.

- [x] Stored csv-file of csv-dispatch example should be more user-friendly @ckaldemeyer 
- [x] example script should allow to choose the solver with optional parameter to check the installation with optional parameter (default: cbc) @gplssm 
- [x] Check output of examples if executed from script @uvchik
- [x] Add testing "example" @uvchik

**Check comments inside examples**
(e.g. by someone who did not created the example) @oemof/oemof-solph
- [ ] csv-reader - invest
- [x] csv-reader - dispatch @uvchik
- [ ] flexible modelling
- [x] simple dispatch @uvchik
- [x] storage invest @uvchik

**Check whether the description of your example is okay [-> here](http://oemof.readthedocs.io/en/stable/oemof_solph.html#solph-examples)** 
 - [x] csv-reader @ckaldemeyer 
 - [x] flexible modelling @simnh 
 - [x] simple dispatch @simnh 
 - [ ] storage invest


simnh:
What about this one ?

uvchik:
I changed the description to make things clearer.

uvchik:
If you think your name shouldn't be there please write a comment and remove your name.

simnh:
What about this one? Can the person assigned to task do it and we can merge this one soon?

uvchik:
@ckaldemeyer 'Would you add your new unstack function to the csv examples to make the results stored in a csv-file readable.

uvchik:
I added a testing example.

ckaldemeyer:
> @ckaldemeyer 'Would you add your new unstack function to the csv examples to make the results stored in a csv-file readable.

I could do it. In both of them? I think one example would be sufficient.

uvchik:
> I could do it. In both of them? I think one example would be sufficient.

Yes.

Please the check boxes of the first comment.

ckaldemeyer:
> Please the check boxes of the first comment.

Check

ckaldemeyer:
> Stored csv-file of csv-dispatch example should be more user-friendly @ckaldemeyer

I don't know who wrote this and what is meant by it exactly. The CSV file contains a description of its structure and should be self-explaining, the solver is CBC and the examples are linked correctly in the documentation. From this point I think it should be fine.

Or am I missing something?

uvchik:
> Stored csv-file of csv-dispatch example should be more user-friendly @ckaldemeyer

That was about the output file. The output file is one column and as we talked (a long time ago) we agreed that it would be more user-friendly if we would store it in a different way. Maybe a bus-balance-slice before the `to_csv` will do the trick.

ckaldemeyer:
> That was about the output file. The output file is one column and as we talked (a long time ago) we agreed that it would be more user-friendly if we would store it in a different way. Maybe a bus-balance-slice before the to_csv will do the trick.

Ah, you're right.. Thanks! I'll do it this afternoon.

ckaldemeyer:
> That was about the output file. The output file is one column and as we talked (a long time ago) we agreed that it would be more user-friendly if we would store it in a different way. Maybe a bus-balance-slice before the to_csv will do the trick.

I have now added a part that saves one csv-file with outputs/inputs/other per bus.

We could also add this as a method to the ResultsDataFrame like "bus_balance_to_csv" with a list of busses as argument and per default all busses to be written to the file system. I guess this function might be used quite often..

What do you think?

ckaldemeyer:
> I have now added a part that saves one csv-file with outputs/inputs/other per bus.
> 
> We could also add this as a method to the ResultsDataFrame like "bus_balance_to_csv" with a list of busses as argument and per default all busses to be written to the file system. I guess this function might be used quite often..
> 
> What do you think?

I think this would be useful and just implemented this functionality in the RDF. Please check if something's missing

uvchik:
I think the examples are much cleaner now and this lowers the entry barrier for new users.

Thank you @simnh and @ckaldemeyer for your help.

uvchik:
> I think this would be useful and just implemented this functionality in the RDF. Please check if something's missing

I really get used to the MulitiIndex DataFrame but within office-applications the new form is much better. Cool feature :tada:

PR #248, opened 2016-11-11
Add a transformer with one input and two outputs and a variable fraction between output flows

uvchik:
In contrast to the simple transformer this transformer will have a variable fraction between its output flows. As there are (so far) no use case for transformers that have a variable fraction and more than two outputs this transformer class will be restricted to one input and two outputs.

The name of the class could be `VariableFractionTransformer` (to be discussed)

- [x] add VariableFractionTransformer class
- [x] add block for VariableFractionTransformer
- [x] add grouping for VariableFractionTransformer
- [x] update whatsnew
- [x] write test
- [x] write example
- [x] write documentation (API, solph, example)
- [x] all tests still work (including example test)


simnh:
We should check if the internal calculation of parameters also works with BinaryFlow objects...

uvchik:
We do not need the internal calculation. We can move it to the constraint if it does not work. Will you check it.

I am almost done if everything works.

simnh:
Are you sure that this wokrs as expected, e.g. if the power to heat ratio is 1 the loss index will by -0.5 for 
a condensing eff of 0.5, or am I wrong?

uvchik:
No I'm not sure, but on the first glance it looks not to bad. I wrote a lot today and of course there could be bugs.

A typical [VariableFractionTransformer](http://oemof.readthedocs.io/en/features-variable_fraction_transformer/oemof_solph.html#variablefractiontransformer) is defined as follows:

```python
solph.VariableFractionTransformer(
    label='variable_chp_gas',
    inputs={bgas: solph.Flow(nominal_value=10e10)},
    outputs={bel: solph.Flow(), bth: solph.Flow()},
    conversion_factors={bel: 0.3, bth: 0.5},
    efficiency_condensing={bel: 0.5}
    )
```

The  `main_flow_loss_index` should not be negative unless the efficiency_condensing (of the main flow) is smaller than the conversion_factor of the main flow. But this makes no sense.
```python
(efficiency_condensing(bel, t) - conversion_factors(bel, t)) / conversion_factors(bth, t)
```

Did you run the example. Can you give me the concrete parameters that causes the problem?

simnh:
Oh missed the brackets...sorry. Will you provide an example or extend one? Then it would be easy to test this functionality? e.g. the simple_dispath example might be extenden by this transformer...this would also add it to the tests...

uvchik:
There is an example if you switch to the right branch. And it is already added to the example tests.

Checkout the `features/variable_fraction_transformer`branch and run the `variable_chp` example. Does it work for you? If you have time you can have a look at the documentation.

http://oemof.readthedocs.io/en/features-variable_fraction_transformer/oemof_solph.html#variablefractiontransformer

http://oemof.readthedocs.io/en/features-variable_fraction_transformer/api/oemof.solph.html#oemof.solph.network.VariableFractionTransformer

uvchik:
- [ ] add VariableFractionTransformer class
- [ ] add block for VariableFractionTransformer
- [ ] add grouping for VariableFractionTransformer
- [ ] update whatsnew
- [ ] write test
- [ ] write example
- [ ] write documentation (API, solph, example)
- [ ] all tests still work (including example test)

Anything missing?

uvchik:
Anything missing?

@simnh What about your BinaryFlow problem?

If everything is fine, you can merge.

uvchik:
We may want to rename the `efficiency_condensing` parameter.

 * max_conversion_factor
 * conversion_factor_one_output
 * conversion_factor_single_output

simnh:
I can try to look at the problem regarding binary things later..hopefully..if the example is up, what I think it is.

uvchik:
I renamed the `efficiency_condensing` to `conversion_factor_single_flow`.

uvchik:
I think I am done so far.

I did not use any special attributes (related to #253) but only the `self.conversion_factor_single_flow`.

uvchik:
@simnh If you are satisfied you can merge.

simnh:
I am not sure, but I think it should work...Maybe the docstring should contain some infos on limitations, e.g. that the nominal_value of the flow should be set via the inflow and not on the outflows, and thus the binary attributes should also be set on this flow working.

ckaldemeyer:
> Checkout the features/variable_fraction_transformerbranch and run the variable_chp example. Does it work for you? If you have time you can have a look at the documentation.
>
> http://oemof.readthedocs.io/en/features-variable_fraction_transformer/oemof_solph.html#variablefractiontransformer
>
> http://oemof.readthedocs.io/en/features-variable_fraction_transformer/api/oemof.solph.html#oemof.solph.network.VariableFractionTransformer

First of all, thanks for implementing this functionality!

From my point of view the documentation could be improved a bit as it is quite text-heavy and even for me (as I know the functionality) hard to understand.

I would offer to look after some texttual formulations. Additionally, I think it could be helpful to provide some sort of visualisation such as (http://pasteboard.co/9huBf6bKt.png) but in an abstract form, not specificly for power and heat. I think this would make things a bit clearer..

What do you think?

uvchik:
@ckaldemeyer You are right, this would definitly improve the documentation. For the TwoInputsOneOutput transformer (old solph) I added a picture like the one below. But this picture needs to be revised as the naming is too far from the real parameters. I think this should be a new PR where we create a template that we can use it with all components.

oemof_transformer

@simnh  It is always good to make a documentation as clear as possible but it is not forbidden to define the nominal_value for the output flows. If people do not understand what they are doing they should keep with the example.

ckaldemeyer:
> @ckaldemeyer You are right, this would definitly improve the documentation. For the TwoInputsOneOutput transformer (old solph) I added a picture like the one below. But this picture needs to be revised as the naming is too far from the real parameters.

I think 2D or 3D visualisations in coordinate systems such as in [this](http://www.sciencedirect.com/science/article/pii/S0360544211004348) or [this](http://www.sciencedirect.com/science/article/pii/S0377221702004605) arcticle are quite standard for LP/MILP and I would suggest that we use a similar approach.

> I think this should be a new PR where we create a template that we can use it with all components.

I agree. It makes sense to use a uniform way to decribe equations!

ckaldemeyer:
I have pushed my revision of the rst-paragraph/docstring. If you don't like it or if I got something wrong, just change/remove it..

ckaldemeyer:
> I think 2D or 3D visualisations in coordinate systems such as in this or this arcticle are quite standard for LP/MILP and I would suggest that we use a similar approach.

I can also make a suggestion. But after christmas.. ;-)

uvchik:
I would like to merge it until v0.1.2. Any serious objections beside optional improvements?

@oemof/oemof-solph The name is `VariableFractionTransformer`. Any suggestion for improvement?

PR #249, opened 2016-11-13
Features/solph speed up

simnh:
This pull request aims to use functionality that has been implemented by the pypsa developers to speed up constraint building etc. We had that for the old version already, I think we should try to use it now as well. For the bus balance for example it is faster, and I think significantly for large instances.  Took me some time to implement it for lineartransformer as well. There it does not speed thins up (it's even slower but maybe it's still not equivalent to the old expression). So from my point of view the procedure would look like this:
- taking major constraints (busbalance, lineartransformer etc) and compare results for a large test instance (maybe renpassis europe model) 
- where we identify enhancement we will use the improved code. 

Right now you can check it out in the blocks.py module (look at the LinearTransformer class there) 

The test fail at the moment, because the lp-files look differently (but this is just due to different signs inside equations.

### DIFF
--- a/oemof/solph/blocks.py
+++ b/oemof/solph/blocks.py
@@ -8,7 +8,7 @@ from pyomo.core import (Var, Set, Constraint, BuildAction, Expression,
 from pyomo.core.base.block import SimpleBlock
 from .plumbing import LConstraint, LExpression, linear_constraint
 import timeit
-fast_build = True
+
 
 class Storage(SimpleBlock):
     """ Storages (no investment)
@@ -818,7 +818,7 @@ class LinearTransformer(SimpleBlock):
         I = {n: n._input() for n in group}
         O = {n: [o for o in n.outputs.keys()] for n in group}
 
-        fast_build = False
+        fast_build = True
         start = timeit.default_timer()
         if not fast_build:
             self.relation = Constraint(group, noruleinit=True)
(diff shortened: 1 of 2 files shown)


ckaldemeyer:
Does this build on the standard API of pyomo or is it some sort of hack? I am just asking because of maintenance reasons and the  usability for new users..

simnh:
this is a hack..so we should probably keep both, the original code and the hack...but it's definitely worth it.

simnh:
I just thought I move the code here, so that anyone can simply start altering and testing faster code.

ckaldemeyer:
> this is a hack..so we should probably keep both, the original code and the hack...but it's definitely worth it.

Sounds good! Especially for memory consuming models or if models are embedded (e.g. in an EA). Thanks

simnh:
Ah for memory consumption @gnn  will probably add another hack (at least i hope so), also from pypsa....i'll be great!

uvchik:
> i'll be great! 

You definitely will :smirk: 

By now you just replaced the code (I guess for testing purpose). Do you have an idea how to organise both ways next to each other. Shouldn't we add something like `fast_blocks`.... I think plumbing is not a good place for a special feature.

simnh:
I talked to Tom B. and Gabriel H from the pyomo developers. The developers actually recommen doing it this way (like in PyPSA) to speed things up...So maybe we should do this ? The implementation is basically ready. Only the lp-files inside tests have to be updated...

ckaldemeyer:
> I talked to Tom B. and Gabriel H from the pyomo developers. The developers actually recommen doing it this way (like in PyPSA) to speed things up...So maybe we should do this ? The implementation is basically ready. Only the lp-files inside tests have to be updated...

Based on this, for me, this is more like a "Must Do" because it affects every model and only requires some time.

But can it break in certain cases when the API changes, etc..?! If not, I would implement it.

simnh:
Ok, I guess API won't be a problem so far...But of course I can not guarantee it...(however we have requirements and use 'fancy' pyomo stuff everywhere...thus I think this is not an argument for me...

uvchik:
The LP-files are not the problem, because I guess it is just the order.

If it is faster, we should give it a try. (Do we know how much faster?)

We have to keep an eye on the pyomo updates. In urgent case we can just change the version number in the dependencies. We (developer) should try to keep pyomo up to date, that we become aware of problems.

simnh:
There are some new discussions going on. I will open an issue where we can discuss these things. these implementations will be obsolete.

PR #250, opened 2016-11-14
oemof-pyPSA integration

simnh:
This branch will create a powerflow module or package to enable pyPSA integration via oemof.powerflow components.


simnh:
Due to limeted project time we agreed to use pyPSA right away for now. Therefore this one is on hold. If somebody wants to take over, feel free. 

I might be easy though to add linear optimal power flow to solph with minimal effort.  As mathematically this is straight forward...

PR #253, opened 2016-11-17
Change structure of network classes

uvchik:
I think it would be faster if the methods are changed to attributes. As I understand it a methods like `_input` will search for the input flow every time it is called. If it is an attribute it will be done once. These attributes are needed when the constraints are build in the blocks module. If we use them inside rules they are called 8760 times (or more). 

Furthermore I do not understand why do use protected methods as we do want to use them in other modules and classes that are not child classes.

### DIFF
--- a/doc/whatsnew/v0-1-2.rst
+++ b/doc/whatsnew/v0-1-2.rst
@@ -35,6 +35,7 @@ Bug fixes
 Other changes
 #############
 
+* speed up constraint-building process by removing unnecessary method call
 
 
 
--- a/oemof/solph/blocks.py
+++ b/oemof/solph/blocks.py
@@ -59,6 +59,9 @@ class Storage(SimpleBlock):
         if group is None:
             return None
 
+        I = {n: [i for i in n.inputs][0] for n in group}
+        O = {n: [o for o in n.outputs][0] for n in group}
+
         self.STORAGES = Set(initialize=[n for n in group])
 
         def _storage_capacity_bound_rule(block, n, t):
@@ -87,9 +90,9 @@ class Storage(SimpleBlock):
             expr += block.capacity[n, t]
             expr += - block.capacity[n, m.previous_timesteps[t]] * (
                 1 - n.capacity_loss[t])
-            expr += (- m.flow[n.input, n, t] *
+            expr += (- m.flow[I[n], n, t] *
                      n.inflow_conversion_factor[t]) * m.timeincrement[t]
-            expr += (m.flow[n, n.output, t] /
+            expr += (m.flow[n, O[n], t] /
                      n.outflow_conversion_factor[t]) * m.timeincrement[t]
             return expr == 0
         self.balance = Constraint(self.STORAGES, m.TIMESTEPS,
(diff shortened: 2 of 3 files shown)


uvchik:
If we change the methods to attributes they are created while the class is initialised.  I can change it if you do not like this behavior.

ckaldemeyer:
> I think it would be faster if the methods are changed to attributes. As I understand it a methods like _input will search for the input flow every time it is called. If it is an attribute it will be done once. These attributes are needed when the constraints are build in the blocks module. If we use them inside rules they are called 8760 times (or more).

> Furthermore I do not understand why do use protected methods as we do want to use them in other modules and classes that are not child classes.

I am not really into it at the moment and maybe cannot estimate the full consequences. But if the inputs/outputs only hold the data and do not (and are not meant to) do something (in future, which I do not see), I absolutely agree with you. And maybe this really makes a difference in terms if model building speed!

uvchik:
They are duplicate anyway because we already do have the `inputs` and `outputs` attribute. These attributes may contain more than one output. In classes where we have restricted flows (only one input flow allowed...) we created the additional input, output, ... method. We already realised that this is not very fast so called the method once to assign a variable. This does not make sense because if we do so we do not need this method at all.

In my opinion we should keep the attribute anyway to make clear that this is restricted. In that case we should check this while initialising the object. If we do not want to do this and keep it in the hands of the users we should remove this attribute/method totally and use the variable assignments in the blocks as we already do.

I think it is a good way to check the inputs directly to make errors better understandable because we already know that a storage with two input flows will not work as expected.

simnh:
If outputs are added after object instantiation, this won't work. So , for example the csv - reader would not work either...so from my perspective we should keep the methods or find a way on how these attributes are set once outputs/inputs are set in a instantiatet object

uvchik:
As I understand the method it is just a shortcut. We do not need it at all, right?

e.g. LinearTransformer:

```python
   def _input(self):
        """ Returns the first (and only) input of the storage object
        """
        return [i for i in self.inputs][0]
```
As you remember we used this shortcut in the constrains but found out that it is too slow. Now we use something like this ` I = {n: n._input() for n in group}` and then use the `I` in the constraints so that the method is not called again and again. This is a shortcut for the shortcut. If we change the lines (similar to the outputs) in the `blocks.LinearTransformer`, we do not need the method anymore.
```diff
- I = {n: n._input() for n in group}
- O = {n: [o for o in n.outputs.keys()] for n in group}
+ I = {n: [i for i in n.inputs.keys()] for n in group}
+ O = {n: [o for o in n.outputs.keys()] for n in group}
```

I don't know how it is used in the csv-reader but no I think it is better to remove it at all. It is more confusing than helpful.

simnh:
But how do I get the input/output from a component, when they have not been added inside the constructor? We need a method that can be called, and of course the can set the attribute of the object instead of returning a value.

uvchik:
> But how do I get the input/output from a component, when they have not been added inside the constructor?

`[i for i in self.inputs]`

simnh:
if i do this : 

```
LinearTransformer(label='LF', intputs={}, outputs={})
```

for what ever reason and add the flows later, this option will not work i assume

uvchik:
> ...above mentioned option will not work i assume

Why not? If you add the flows later, you can get the flows later. The `_input()` method also gets its values from the `self.inputs` attribute. You just have to add the line from your method directly to your code. It is a little bit longer but not much and much cleaner.

uvchik:
How do we go on with this issue. I would be nice to get a feedback of @gnn but he seems to be busy.

@simnh : Are you convinced or do you still have objections?

simnh:
Go ahead as you like (inside solph)

uvchik:
Okay, all tests (example/nosetests) do work.

Issue #254, opened 2016-11-18
New pyomo version breaks oemof tests


uvchik:
If you upgrade to the newest pyomo version (> 5.x) the oemof-tests break. This is not caused by an error but an API change. The equations in the lp-files are still equivalent (see example). 

```diff
 c_e_Bus_balance(electricityBus_1)_:
-+1 flow(electricityBus_excess_1)
-= 720000
+-1 flow(electricityBus_excess_1)
+= -720000
```

1. We could change the rtd_requirements.txt to pyomo >= 5.0 and change the lp-file. This will be fast, does not harm anybody and will fix the Travis-CI problems. In that case you have to make sure that you use an environment with pyomo > 5.0 if you run the tests locally. 

2. We could also set the setup.py requirements to pyomo >= 5.0. This solves everything but is not true. People using pyomo in any other case will have to upgrade pyomo to use oemof even though oemof works fine with pyomo 4.x. In the long run we should do that if we do the tests with pyomo >= 5.0. People might want to upgrade to 5.x anyway.

@oemof/oemof-solph Do you agree to do (1.) now and (2.) later?

ckaldemeyer:
I would vote for 2.) as updating pyomo is not a big deal.

uvchik:
I did (1.) already to fix the Travis issue. Now (2.) is not urgent. I will wait for further comments.

gnn:
I think we should eventually find a way to fix this, by having a more robust test setup that doesn't break with those changes. I'll postpone this to the next release, hoping that I'll find the time to figure something out.

gnn:
Should be solved. See PR #357.

PR #255, opened 2016-11-21
Fix tests for pyomo >= 0.5

uvchik:
Tests broke for pyomo >= 0.5. This PR will try to fix it.


uvchik:
Travis CI is happy again. If you want to run the test locally you have to make sure that pyomo >= 5.0 is installed. We could also change the general requirements but it is not urgent. I will leave issue #254 open to discuss this point but merge this.

Every PR should merge the actual dev to fix the problems with Travis-CI.

PR #268, opened 2017-01-09
Features/addtional flow attributes in NodesFromCSV

simnh:
At one point we might also just use ad dictionary with keys for specific attributes: 

```
addtional_attributes = {'flow': [], 'seq': [], 'bus' =[]} 
```

But this would change the API, so for now just adding an attribute

### DIFF
--- a/doc/whatsnew/v0-1-2.rst
+++ b/doc/whatsnew/v0-1-2.rst
@@ -11,7 +11,7 @@ API changes
 New features
 ############
 
-
+* Allow to set addtional flow attributes inside NodesFromCSV in solph inputlib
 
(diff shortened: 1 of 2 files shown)


ckaldemeyer:
> At one point we might also just use ad dictionary with keys for specific attributes:
> 
> addtional_attributes = {'flow': [], 'seq': [], 'bus' =[]} 
> But this would change the API, so for now just adding an attribute

Thanks for implementing and the quick review process ;)

simnh:
Sorry, I thought you answered on this one instead of the other pull -request because I did not check here again :D ....

PR #270, opened 2017-02-10
Update parameters or sequences in a csv file

uvchik:
I added to functions to update parameters of existing csv-files.

The idea is that you have a list of region names (or names of power lines) and a pattern.

If the csv-file (my_csv_seq.csv) contains the columns "DE_pv_feedin, NL_pv_feedin, DK_pv_feedin" and you want to update them with the data from a pv_feedin Dataframe (e.g. from feedinlib) you can do the following:

```python
name = 'my_csv'
regions = ['DE', 'NL', 'DK']  # object_names
pattern = '{0}_pv_feedin'
pv_feedin = pd.DataFrame({'DE': [1, 5, 3, 2...], 'NL': [3, 2, 6,...], ...})  # data

update_sequence(name, pattern, data, regions)
```

### DIFF
--- a/oemof/solph/__init__.py
+++ b/oemof/solph/__init__.py
@@ -8,3 +8,4 @@ from oemof.solph.models import OperationalModel
 from oemof.solph.groupings import GROUPINGS
 from oemof.solph.options import (Investment, BinaryFlow, DiscreteFlow)
 from oemof.solph.inputlib.csv_tools import NodesFromCSV
+from oemof.solph.inputlib import csv_tools
--- a/oemof/solph/inputlib/csv_tools.py
+++ b/oemof/solph/inputlib/csv_tools.py
@@ -375,7 +375,7 @@ def resample_sequence(seq_base_file=None, output_path=None, samples=None,
 
 
 def update_parameter(name, pattern, query_col, target_col, data, object_names,
-                     scenario_path='scenarios'):
+                     scenario_path):
     """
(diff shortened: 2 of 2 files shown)


ckaldemeyer:
I'll look at it tomorrow!

uvchik:
> 1.) "from shutil import copyfile" does this module also work on windows/osx?

It is supposed to work for all os (https://docs.python.org/3/library/shutil.html).

> 2.) Should we provide an example on how to use this?

I think the whole csv-tool could have a more extensive documentation. I will write an example if you basically agree.

The functions `resample_sequence` and `update_sequence` do similar stuff in a different way. We should discuss which way is better (more stable, faster,...) and use the same way for both functions. I realised this after implementing my function. Otherwise I would have done this before.

ckaldemeyer:
> I think the whole csv-tool could have a more extensive documentation. I will write an example if you basically agree.

Maybe we can just blow up one of the [csv-reader examples](https://github.com/oemof/oemof/tree/dev/examples/solph/csv_reader) to demonstrate the functionality.

Or we just create a new one that focusses on special features such as this function!

ckaldemeyer:
> The functions resample_sequence and update_sequence do similar stuff in a different way. We should discuss which way is better (more stable, faster,...) and use the same way for both functions. I realised this after implementing my function. Otherwise I would have done this before.

:+1:

uvchik:
I decide to make general approach see #294.

PR #273, opened 2017-02-22
Allow conversion factors as sequence

ckaldemeyer:


ckaldemeyer:
This should allow to add conversion factors as a sequence.

The CSV reader could have a complete remastering as you suggested but it should work for now..

simnh:
Looks good, but do we have to include 'conversion_factors' in seq attributes? If this is not needed it is kind of confusing that for some attributes it needs to be in 'seq_attributes' and for others not, though for all it is possible...do you understand what I mean? 

However, rather something for the remastering :)

ckaldemeyer:
> Looks good, but do we have to include 'conversion_factors' in seq attributes?
@uvchik seems to need this possibility. What would be the alternative?

> If this is not needed it is kind of confusing that for some attributes it needs to be in 'seq_attributes' and for others not, though for all it is possible...do you understand what I mean?

I am not sure. For regular node/flow attributes that are sequences it should work as expected when they are in the csv-file. The conversion factors are a special case treated seperately because they are an attribute of the node in form of a dict.

But maybe I just don't get your point ;)

> However, rather something for the remastering :)

:+1:

ckaldemeyer:
If it works we can add it to WHATSNEW

simnh:
I think, @uvchik can test it because he is working on an app with this anyway...

uvchik:
I will test it until Friday. Thank you.

oemof/solph/inputlib/csv_tools.py
```diff
@@ -11,7 +11,7 @@
 
 def NodesFromCSV(file_nodes_flows, file_nodes_flows_sequences,
                  delimiter=',', additional_classes={},
-                 additional_seq_attributes=[], addtional_flow_attributes=[]):
+                 additional_seq_attributes=[], additional_flow_attributes=[]):
```
uvchik: It does not really belong to you changes, so you may ignore this comment but your default attributes are mutable.

It might not be a problem in our case but I would like to be in line with common python rules.

http://docs.python-guide.org/en/latest/writing/gotchas/

ckaldemeyer: > It does not really belong to you changes, so you may ignore this comment but your default attributes are mutable.
> 
> It might not be a problem in our case but I would like to be in line with common python rules.
> 
> http://docs.python-guide.org/en/latest/writing/gotchas/

This has been added from @simnh if I remember it right and I think he just needed the functionality quickly and didn't think about it too much. Neverthess, you are right!

Should we fix this?

ckaldemeyer:
> It works but will I have the same problem if I add an 'seq' to an upper or lower bound (max/min)?

Actually this should work since only the conv-factors were a special case because of the dict. Have you tested?

oemof/solph/inputlib/csv_tools.py
uvchik: I think we should. Just change the default argument to None (`additional_classes=None`) and add an if-clause.

```python
if additional_classes is None:
    additional_classes = dict() #  or list()
```

uvchik:
Not so far, but I will.

In the long run we should divide this `godfather function` into sub-functions but by now I would merge it soon (after fixing the default arguments).

Then I would ask users who uses the csv-reader to test it before we release it in March.

BTW: It's not to nice that the name is formatted like a class name with capital letters. It is against our own rules.

oemof/solph/inputlib/csv_tools.py
uvchik: Just did it. You can revert it if you do not like it.

uvchik: According to: http://docs.python-guide.org/en/latest/writing/gotchas/

ckaldemeyer:
You were faster, I just wanted to push... ;-)

> Not so far, but I will.
> 
> In the long run we should divide this godfather function into sub-functions but by now I would merge it soon (after fixing the default arguments).

I think the function just grew because many people are using it meanwhile. I can help you if you want.




> BTW: It's not to nice that the name is formatted like a class name with capital letters. It is against our own rules.

Thats for the same reason. But there were also people reviewing the PR ;-)

Let's just create a better and leaner function adhering to the coding guidelines!

ckaldemeyer:
> BTW: It's not to nice that the name is formatted like a class name with capital letters. It is against our own rules.

I think this was because in the beginning it was planned as a class..

uvchik:
Could you ask users you know to test the new version. I just want to be sure that old scenarios will still work. I definitely want to get this into the next release.

ckaldemeyer:
@s3pp tested one renpassG!S scenario and it works without a problem. same for my current mini-example

ckaldemeyer:
I guess it should be fine

uvchik:
Okay, for me it's done.

uvchik:
I know, that it is not the core of this PR but the fixed_costs seem to be ignored as I can write whatever I like in this field. It is the third field from the right in the screenshot. No error is raised. If I write such strings in one of the other fields an error is raised. 

bildschirmfoto_csv_reader

simnh:
Die you set the nominal_values, otherwise nothing will happen?

uvchik:
Thanks @simnh, now I found the right expression to get the value of the attribute:

```diff
- om.flow['DE01_pp', 'DE01_bus_el'].fixed_costs
+ om.flows[es.groups['DE01_pp'], es.groups['DE01_bus_el']].fixed_costs
```

As I work on a different computer today I do not have the problem assigned yesterday.

ckaldemeyer:
> As I work on a different computer today I do not have the problem assigned yesterday.

So it throws an error now?

uvchik:
> So it throws an error now?

Yes, everything works as supposed. No objections any more. If @simnh agrees, you can merge.

simnh:
If it works, please merge :)

uvchik:
Thanks to @krolga for reporting, @ckaldemeyer for fixing and @simnh for help :tada:

ckaldemeyer:
> Thanks to @krolga for reporting, @ckaldemeyer for fixing and @simnh for help

I'll merge now. Creating a good mood here using the emojis is a good idea @uvchik  :sunny: :joy:

ckaldemeyer:
I have added some lines to whatsnew

uvchik:
I :heart:  emojis :smile: 

And it was a really good example how open source improves the quality of the code.

uvchik:
I moved your commit to dev.

PR #277, opened 2017-02-23
Add method to slice a balance around a bus

ckaldemeyer:
I have added a method to slice inputs/outputs/other around busses as is seems to be a common subset to be extracted from the RDF.

TODOs:

* Update WHATSNEW
* Check docstrings

Could you check and give some feedback?

### DIFF
--- a/oemof/outputlib/__init__.py
+++ b/oemof/outputlib/__init__.py
@@ -186,10 +186,7 @@ class ResultsDataFrame(pd.DataFrame):
         outputs = self.slice_unstacked(bus_label=kwargs['bus_label'],
                                        type='from_bus',
                                        formatted=True)
-        other = self.slice_unstacked(bus_label=kwargs['bus_label'],
-                                     type='other',
-                                     formatted=True)
-        subset = pd.concat([inputs, outputs, other], axis=1)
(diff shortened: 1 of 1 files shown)


ckaldemeyer:
I have missed that sometimes "other" does not exist.

The best would be to only use existing levels of "type":

    results.index.levels[1]
    Out[81]: Index(['from_bus', 'to_bus'], dtype='object', name='type')

I'll have a look how this could be solved..

ckaldemeyer:
It should work with the last commit. Would it be o.k. for you?

Then I would add it to whatsnew

uvchik:
What is `kwargs` for? As I understand it we have one parameter and it is not optional!

```diff
- def slice_bus_balance(self, **kwargs):
+ def slice_bus_balance(self, bus_label):
```

ckaldemeyer:
> What is kwargs for? As I understand it we have one parameter and it is not optional!
> 
> - def slice_bus_balance(self, **kwargs):
> + def slice_bus_balance(self, bus_label):

You're right. I'll kick it out

ckaldemeyer:
Is this fine for you now? I could merge it then

uvchik:
I didn't test it but I trust you that you did. Go ahead.

PR #279, opened 2017-02-27
reconcile rtd_requirements and setup.py

uvchik:
These files are different due to hot fix. But this is confusing and we should revert it.


uvchik:
I did some tests and fixed my own mistakes but now it seems to work without the requirements file.

This eliminates a source of errors because all requirements are in one place.

ckaldemeyer:
> I did some tests and fixed my own mistakes but now it seems to work without the requirements file.
> 
> This eliminates a source of errors because all requirements are in one place.

Thanks for the fix

Issue #280, opened 2017-02-28
Solph-Constraint Sandbox?


simnh:
Ola! 

We already talked about that, but we do not have a place at the moment for this. What I was thinking is a module where we collect functions that allow to add additional constraints that people can use inside their apps. For example one function could look like this: 
```
def flow_share(om, f1, f2, share):
    """
    """
    _def flow_rule(m, t):
        return m.flow[f1, t] == m.flow[f2, t] * share
    om.flow_share_constraint = po.Constraint(om.TIMESTEPS, rule=_flow_rule)

```

I frequently have such functions, so I think a oemof place to share it would be nice...

What do you think?

ckaldemeyer:
You mean similar to this example? 

https://github.com/oemof/oemof/tree/dev/examples/solph/flexible_modelling

Sounds like a good idea and for me it would be a no-regret as the code is hidden and does not affect any functionalities. On the other hand it has to be maintained. But if the functions are commonly used its definetely an improvement!

ckaldemeyer:
Concerning you provided example I am not sure if it should be externally or basic functionality as I stumbled upon this lacking functionality last week

simnh:
yes kind of in this example. I mean, it can also be used as an example collection of functions, for users who need adapt their models. 

> Concerning you provided example I am not sure if it should be externally or basic functionality as I stumbled upon this lacking functionality last week

Well I thought, internal in a sense of a solph module but external in a sense of the creation of contraints from parameter settings...In this way it's inside oemof and we can see whether some of the functionality inside the 'sandbox' has to be integrated in the 'solph-core' .

ckaldemeyer:
> Well I thought, internal in a sense of a solph module but external in a sense of the creation of contraints from parameter settings...In this way it's e.g. inside oemof.solph.sandbox and we can see whether some of the functionality inside the 'sandbox' has to be integrated in the 'solph-core' .

Sound very good!!!

simnh:
I talked to @gnn and the proposed why is as follows: 

* create class like 'Container' where we add additional constraint functionality
* Add this class as a parent to OperationalModel, i.e. subclassing from two classes
* In this way we can use all methods of the container classes as methods inside the operationalmodel

This procedure is called 'mix-in' and should be the best way to do it. Only thing we need to check which class of the two we are subclassing from comes first ( most likely the new 'Container' class). 
Additionally we need to make sure that the Container class will not hold any class variables, but this should be no problem. 

[...]

PR #284, opened 2017-03-01
Add solph mixin classes

simnh:
This is linked to #280 . With this structure we are even able to have multiple Mixin-classes

### DIFF
--- a/oemof/solph/mixins.py
+++ b/oemof/solph/mixins.py
@@ -1,0 +1,24 @@
+# -*- coding: utf-8 -*-
+"""
+"""
+import pyomo.environ as po
+
+class ConstraintMixin():
+    """ The ConstraintMixin class contains mehtods that are used in the
+    :class:`OperationalModel <oemof.solph.models.OperationalModeel>`. This is
+    achivieved as the OperatioanlModel inherits from the ConstraintMixin class.
+
+    """
+
+
+    def flow_share(self, f1, f2, share):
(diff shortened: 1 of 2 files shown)


uvchik:
I gernerally like the proposal. I think we should keep in mind what @ckaldemeyer said [here](https://github.com/oemof/oemof/issues/280#issuecomment-283274646) but I think in this case we just add new methods. It is a very simple construction and therefore it will be easy to improve or change it later.

1. How is the call? Like this:

```python
om.flow_share(flow_object1, flow_object2, 0.5)
```

2. I would add a typecheck (check if f1 is of type flow) . It makes debugging easier.

3. Is it possible to pass the labels (string) instead of the objects? How do I get the flows? Like this:

```python
flow_object1 = om.flows[energysystem.groups['label_source'], energysystem.groups['label_target']]
# The short version would be nicer in may opinion:
flow_object1 = ('label_source', 'label_target')
```

simnh:
I think this is obsolete for now, see after version 0.2

uvchik:
A pitty, I really liked the idea.

simnh:
yeah we still do not have a solution for Constraints....thats right...

ckaldemeyer:
> yeah we still do not have a solution for Constraints....thats right...

What about just writing a module with functions that receive the `es` or `om` as params and additional params to set constraints!? This would also be straightforward..

simnh:
yes I think we could do that as a start

PR #288, opened 2017-03-03
Improve csv-reader

uvchik:
I would like to improve the  the csv-reader. To do so, we have to improve readability.

In the first step, I created sub-functions.

@s3pp : May you rename the functions and add 1-5 lines to the docstring about what they do?

If you would create more sub-functions it would be even better.


uvchik:
Closed in favour of #294

PR #294, opened 2017-03-20
Transfer csv reader to scenario class

uvchik:
The main idea is to place the pandas DataFrame in the centre of the class and not the csv-file.

1. I would like to divide the NodesFromCSV function into sub functions to improve the readability of the csv-reader . Then it might be easier to improve parts of the functions. So I created some subfunctions. @ckaldemeyer, @simnh or @s3pp : May you rename the functions and add 1-5 lines to the docstring about what they do? If you create more sub-functions, it will be even better.

2. For me it would be nice to make it possible to pass a DataFrame directly to the NodesFromCSV function. At the moment I have to store the DataFrame to a csv-file and read it again. In the SolphScenario class the DataFrame has a MultiIndex but in the NodesFromCSV function you transfer the DataFrame into a different shape. @ckaldemeyer @s3pp @simnh Ii it important to be like this or do you think you could handle a DataFrame as it is read by the SolphScenario class. Otherwise I have to transform the DataFrames before calling the NodesFromCSV function.

3. We also could revise the NodesFromCSV function itself but this has not the highest priority.

### DIFF
--- a/oemof/solph/__init__.py
+++ b/oemof/solph/__init__.py
@@ -10,3 +10,4 @@ from oemof.solph.models import OperationalModel
 from oemof.solph.groupings import GROUPINGS
 from oemof.solph.options import (Investment, BinaryFlow, DiscreteFlow)
 from oemof.solph.inputlib.csv_tools import NodesFromCSV
+from oemof.solph.inputlib.csv_tools import Scenario
--- a/oemof/solph/inputlib/csv_tools.py
+++ b/oemof/solph/inputlib/csv_tools.py
@@ -19,9 +19,10 @@ PARAMETER = (
 INDEX = ('class', 'label', 'source', 'target')
 
 
-class SolphScenario:
+class SolphScenario(EnergySystem):
 
     def __init__(self, **kwargs):
+        super().__init__(**kwargs)
         self.p = kwargs.get('parameters')
         self.s = kwargs.get('sequences')
         self.path = kwargs.get('path', path.dirname(path.realpath(__file__)))
--- a/oemof/solph/inputlib/csv_tools.py
+++ b/oemof/solph/inputlib/csv_tools.py
@@ -2,6 +2,7 @@
 
 import pandas as pd
 import os
+import warnings
 import os.path as path
 import logging
 from oemof import network
@@ -423,6 +424,9 @@ def NodesFromCSV(file_nodes_flows, file_nodes_flows_sequences, delimiter=',',
     delimiter : str
         Delimiter of CSV file
     """
+    msg = ("\nNodesFromCSV will be removed in oemof v0.2.0. " +
+           "\nUse the 'create_nodes' method of the 'Scenario' class instead.")
+    warnings.warn(msg, FutureWarning)
     nodes_flows = pd.read_csv(file_nodes_flows, sep=delimiter)
     nodes_flows_seq = pd.read_csv(file_nodes_flows_sequences,
                                   sep=delimiter, header=None)
@@ -448,6 +452,9 @@ def merge_csv_files(path=None, output_path=None, write=True):
     -------
     Tuple of dataframes (nodes_flows, nodes_flows_seq)
     """
+    msg = ("\nmerge_csv_files will be removed in oemof v0.2.0. " +
+           "\nUse pandas.concat and the 'Scenario' class instead.")
+    warnings.warn(msg, FutureWarning)
     if output_path is None:
         output_path = path
 
@@ -509,6 +516,10 @@ def resample_sequence(seq_base_file=None, output_path=None,
     header : list
         List of integers to specifiy the header lines
     """
+    msg = ("\nmerge_csv_files will be removed in oemof v0.2.0. " +
+           "\nUse pandas.DataFrame.reindex and the 'Scenario' class instead.")
+    warnings.warn(msg, FutureWarning)
+
     if samples is None:
         raise ValueError('Missing sample attribute. Please specifiy!')
     if output_path is None:
--- a/oemof/solph/inputlib/csv_tools.py
+++ b/oemof/solph/inputlib/csv_tools.py
@@ -473,9 +473,9 @@ def NodesFromCSV(file_nodes_flows, file_nodes_flows_sequences, delimiter=',',
     delimiter : str
         Delimiter of CSV file
     """
-    msg = ("\nNodesFromCSV will be removed in oemof v0.2.0. " +
-           "\nUse the 'create_nodes' method of the 'Scenario' class instead.")
-    warnings.warn(msg, FutureWarning)
+    # msg = ("\nNodesFromCSV will be removed in oemof v0.2.0. " +
+    #        "\nUse the 'create_nodes' method of the 'Scenario' class instead.")
+    # warnings.warn(msg, FutureWarning)
     nodes_flows = pd.read_csv(file_nodes_flows, sep=delimiter)
     nodes_flows_seq = pd.read_csv(file_nodes_flows_sequences,
                                   sep=delimiter, header=None)
--- a/oemof/solph/inputlib/csv_tools.py
+++ b/oemof/solph/inputlib/csv_tools.py
@@ -56,12 +56,14 @@ class Scenario:
     >>> my_es.add(nodes)  # doctest: +SKIP
     """
 
-    def __init__(self, **kwargs):
-        super().__init__(**kwargs)
-        self.p = kwargs.get('parameters')
-        self.s = kwargs.get('sequences')
-        self.path = kwargs.get('path', path.dirname(path.realpath(__file__)))
-        self.name = kwargs.get('name')
+    def __init__(self, scenario_path=None, name=None, p=None, s=None):
+        logging.warning(
+            "Scenario class is still experimental. API may change frequently.")
+        if scenario_path is None:
+            self.scenario_path = path.dirname(path.realpath(__file__))
+        self.name = name
+        self.p = p
+        self.s = s
 
     def create_parameter_table(self, additional_parameter=None):
         """Create an empty parameter table."""
--- a/oemof/solph/inputlib/csv_tools.py
+++ b/oemof/solph/inputlib/csv_tools.py
@@ -39,15 +39,10 @@ class SolphScenario(EnergySystem):
         self.p = pd.DataFrame(columns=PARAMETER + tuple(additional_parameter),
                               index=my_index)
 
-    def create_sequence_table(self, datetime_index=None, year=None,
-                              interval=None):
+    def create_sequence_table(self, datetime_index=None):
         """Create an empty sequence table."""
-        if interval is None:
-            interval = '60min'
         if datetime_index is None:
-            date_from = '{0}-01-01 00:00:00'.format(year)
-            date_to = '{0}-12-31 23:00:00'.format(year)
-            datetime_index = pd.date_range(date_from, date_to, freq=interval)
+            datetime_index = self.timeindex
 
         my_index = pd.MultiIndex(
             levels=[[1], [2], [3], [4], [5]], labels=[[0], [0], [0], [0], [0]],
@@ -61,9 +56,7 @@ class SolphScenario(EnergySystem):
         """Create empty scenario tables (sequence and parameter)."""
         self.create_parameter_table(
             additional_parameter=kwargs.get('additional_parameter'))
-        self.create_sequence_table(datetime_index=kwargs.get('datetime_index'),
-                                   year=kwargs.get('year'),
-                                   interval=kwargs.get('interval'))
+        self.create_sequence_table(datetime_index=kwargs.get('datetime_index'))
 
     def read_parameter_table(self, filename=None):
         """Read existing parameter table from file."""
@@ -78,14 +71,10 @@ class SolphScenario(EnergySystem):
         self.s = pd.read_csv(filename, header=[0, 1, 2, 3, 4], parse_dates=True,
                              index_col=0)
 
-    def read_tables(self, name=None, scenario_path=None):
+    def read_tables(self, parameterfile=None, sequencefile=None):
         """Read existing scenario tables (parameter and sequence)"""
-        if name is not None:
-            self.name = name
-        if scenario_path is not None:
-            self.path = scenario_path
-        self.read_parameter_table()
-        self.read_sequence_table()
+        self.read_parameter_table(parameterfile)
+        self.read_sequence_table(sequencefile)
 
     def write_parameter_table(self, filename=None):
         """Write parameter table to file."""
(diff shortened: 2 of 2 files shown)


uvchik:
Now it should be possible to test the new class.

The call seems to be more complicated but in the long run there are advantages in my opinion.

OLD CALL:

```python  
scenario_path = 'my/path/scenario'
name = 'test_scenario'

nodes = NodesFromCSV(file_nodes_flows=os.path.join(scenario_path, name + '.csv'),
                     file_nodes_flows_sequences=os.path.join(scenario_path, name + '_seq.csv'))
```
NEW CALL
```python  
scenario_path = 'my/path/scenario'
name = 'test_scenario'

my_scenario = SolphScenario(path=scenario_path, name=name)
my_scenario.read_tables()
nodes = my_scenario.create_nodes()
```

With the new class it is much easier to use pandas-methods. At the moment `p` is the parameter table and `s` is the sequence table. But we can talk about names if we agree in the basic structure.
```python  
scenario_path = 'my/path/scenario'
name = 'test_scenario'

my = SolphScenario(path=scenario_path, name=name)
my.read_tables()
my.p = pandas.concat([my.p, df_with_additional_rows])
my.p.reindex(new_index)  #  pandas.DataFrame.reindex method
my.p['new_attribute'] = ""
my.p.loc[index_of_an_object, 'new_attribute'] = 'seq'  # set parameter to 'seq'
my.s[index_of_an_object + ('new_attribute',) = my_time_series  # add sequence
# ....
my.name = 'test_scenario_edited'
my.write_tables()
nodes = my.create_nodes()
```

I added a new column to the DataFrame (`sort_index`). This is necessary to keep the csv-file readable after you have added objects etc..

uvchik:
I thought it might be a good idea to inherit from the EnergySystem class. In that case I just have to add a `timeindex` to make it possible to pass the scenario-object directly to the OperationalModel instance.

```python  
scenario_path = 'my/path/scenario'
name = 'test_scenario'
datetime_index = pd.date_range('2015-01-01 00:00:00'.format(year),
                               '2015-12-31 23:00:00'.format(year), freq='60min')
my_scenario = SolphScenario(path=scenario_path, name=name, timeindex=datetime_index)
my_scenario.read_tables()
my_scenario.create_nodes()
om = OperationalModel(my_scenario)
om.receive_duals()
om.solve(solver='cbc', solve_kwargs={'tee': True})
...
```
Without inheritance it is more confusing, because two instances (my_es and my_scenario) are created and only one (my_es) is passed to the om and the other is somehow integrated under the hood (magic!).
```python  
scenario_path = 'my/path/scenario'
name = 'test_scenario'
datetime_index = pd.date_range('2015-01-01 00:00:00'.format(year),
                               '2015-12-31 23:00:00'.format(year), freq='60min')
my_scenario = SolphScenario(path=scenario_path, name=name)
my_es = EnergySystem(timeindex=datetime_index)
my_scenario.read_tables()
my_scenario.create_nodes()
om = OperationalModel(my_es)
om.receive_duals()
om.solve(solver='cbc', solve_kwargs={'tee': True})
...
```

ckaldemeyer:
> I would like to divide the NodesFromCSV function into sub functions to improve the readability of the csv-reader . Then it might be easier to improve parts of the functions. So I created some subfunctions. @ckaldemeyer, @simnh or @s3pp : May you rename the functions and add 1-5 lines to the docstring about what they do? If you create more sub-functions, it will be even better.

I'll start with this today and have a look at the other points

ckaldemeyer:
> I thought it might be a good idea to inherit from the EnergySystem class. In that case I just have to add a timeindex to make it possible to pass the scenario-object directly to the OperationalModel instance.

I see your point with the datetime index! Anyway, to me subclassing `SolpScenario` from `EnergySystem` does not feel right since a subclass regularly stands in an `"is a-"relation` to its parent class e.g. the classes `Cat` and `Dog` are subclasses of `Mammal`.

Generally, I like your idea of having some sort of (DataFrame) structure that can be filled somehow and serves as a base for the object instantiation.

An architectural idea (using the current, multi-indexed or xarray or whatever): For me `SolphScenario` or the responding DataFrame(s) structure could be rather a subset/attribute of an energy system. The structure is fixed and it can be filled via defined methods, etc. or one could simply say it has an API. Now, this API can be used by other pieces of code e.g. functions like `nodes_from_csv` from modules like `csv_tools` that could also deliver the structure directly e.g. by adding a flag.

Nevertheless, extending the csv-reader solution to a full scenario class of course leads to code that has to be maintained and understood and might not be relevant for some users. Additionally, we have one more way on how to use oemof and it might confuse some folks if we have X examples having X ways to create an energy system with its nodes and flows. And if let's say @simnh and @s3pp start to build scenarios using `.geojson` files it should be possible easily without having to go normally unneccessary ways.

For me, this discussion feels quite similar as the one we had with the `ResultsDataFrame` and the `DataFramePlot`. I prefer quite general and minimalistic approaches and would stop creating new things where other projects/documentations start e.g. when it comes to plot a pandas DataFrame. But of course this is only one point of view and yours is valid, too.

The questions is how to deal with this: integrate more code when in doubt if it can just remain on the file system or agree on a common thread?

The first is probably easier in the first place but might also lead to a blown up and specific code base which has to be maintained. The second is more difficult and needs more discussion and participation which is sometimes not possible when people are not available for some reason.

@s3pp , @simnh : What do you think about this?

ckaldemeyer:
I have thought about this after our talk on Friday (my brain seems to work again) and meanwhile like the idea and also understand the subclassing procedure in some way.

For me it could be rather called `Scenario` instead of `SolphScenario` as its the same as `Transformer` and `LinearTransformer` as some things are already answered by the place of the code or functionality of the package.

> For me it would be nice to make it possible to pass a DataFrame directly to the NodesFromCSV function. At the moment I have to store the DataFrame to a csv-file and read it again. In the SolphScenario class the DataFrame has a MultiIndex but in the NodesFromCSV function you transfer the DataFrame into a different shape. @ckaldemeyer @s3pp @simnh Ii it important to be like this or do you think you could handle a DataFrame as it is read by the SolphScenario class. Otherwise I have to transform the DataFrames before calling the NodesFromCSV function.

So normally, there should be `nodes_from_df` that creates the nodes from the df and could be used within `nodes_from_csv` after reading and converting the csv-file?

ckaldemeyer:
By the way I do not get notification mails today due to problems with out mail server..

uvchik:
The name `Scenario` is better as the typical call will be `solph.Scenario`. I do not care much for the name if somebody likes anything else better.

At the moment the `nodes_from_csv` function can do both (from csv, from DF), so it could be something like `nodes_from_table`. In the long run we could remove the reading part of the `nodes_from_` function, but this might be a discussion between the contributors who wrote `nodes_from_csv`. I do not care as long as it is possible to pass the DF directly in any way.

ckaldemeyer:
> At the moment the nodes_from_csv function can do both (from csv, from DF), so it could be something like nodes_from_table.

Yep, also because of the single responsibility principle which I would like to adhere to.

> In the long run we could remove the reading part of the nodes_from_ function, but this might be a discussion between the contributors who wrote nodes_from_csv.

I guess it looks like a discussion between us.

uvchik:
Okay the most consequent way would be the following.

* Copying `NodesFromCSV` and renaming it to `nodes_from_df`.
* Removing the "reading-part" from `nodes_from_df`.
* Change the call in `create_nodes` to the new function `nodes_from_df`.
* Moving the DF reshaping of the `create_nodes` method to the `nodes_from_df` function.
* Keeping `NodesFromCSV`,  `merge_csv_files` and `resample_sequence`with a `DeprecatedWarning`: "...be removed in version 0.2.0...". If other contributors decide to keep the functions they can just remove the warning until v0.2.0. 
* Adapt csv-example to the new form.

For me this would be the cleanest way and it will take me about two hours, I guess.

The other way would be to keep both ways in parallel, but this might be more confusing than a hard cut.

ckaldemeyer:
> Okay the most consequent way would be the following.
> 
> Copying NodesFromCSV and renaming it to nodes_from_df.
> Removing the "reading-part" from nodes_from_df.
> Change the call in create_nodes to the new function nodes_from_df.
> Moving the DF reshaping of the create_nodes method to the nodes_from_df function.
> Keeping NodesFromCSV, merge_csv_files and resample_sequencewith a DeprecatedWarning: "...be removed in version 0.2.0...". If other contributors decide to keep the functions they can just remove the warning until v0.2.0.
> Adapt csv-example to the new form.
> For me this would be the cleanest way and it will take me about two hours, I guess.
> 
> The other way would be to keep both ways in parallel, but this might be more confusing than a hard cut.

I could go with this. The only thing that's important for me would be that `nodes_from_df` could be used standalone without the `Scenario` class to read csv-files which would be the case

uvchik:
NodesFromCSV, resample_sequence and merge_csv  will work until oemof v0.2.0, if you want to keep it you shall object until v0.2.0 is released.

The new function `nodes_from_df` will not be able to read csv-files. The input has to be two DataFrames (parameter, sequence). You can use the Scenario class methods to read csv-files (see examples [above](https://github.com/oemof/oemof/pull/294#issuecomment-288034438).

In the next step I would like to adapt the csv-examples. I will keep the old examples for the example test to make sure both ways work until the old way is removed in v0.2.0.

I do not fully understand the difference between the `additional_seq_attributes` and the `additional_flow_attributes` parameter.

ckaldemeyer:
> NodesFromCSV, resample_sequence and merge_csv will work until oemof v0.2.0, if you want to keep it you shall object until v0.2.0 is released.
> 
> The new function nodes_from_df will not be able to read csv-files. The input has to be two DataFrames (parameter, sequence). You can use the Scenario class methods to read csv-files (see examples above).
> 
> In the next step I would like to adapt the csv-examples. I will keep the old examples for the example test to make sure both ways work until the old way is removed in v0.2.0.
> 
> I do not fully understand the difference between the additional_seq_attributes and the additional_flow_attributes parameter. Can anybody help?

Thanks! I am on a conference today and tomorrow and will look at it on Monday.

ckaldemeyer:
> I do not fully understand the difference between the additional_seq_attributes and the additional_flow_attributes parameter. Can anybody help?

Seq attributes are to be converted into flow sequences:

    # attributes that have to be converted into a solph sequence
    seq_attributes = ['actual_value', 'min', 'max', 'positive_gradient',
                      'negative_gradient', 'variable_costs',
                      'capacity_loss', 'inflow_conversion_factor',
                      'outflow_conversion_factor', 'capacity_max',
                      'capacity_min'] + additional_seq_attributes

ckaldemeyer:
> In the next step I would like to adapt the csv-examples. I will keep the old examples for the example test to make sure both ways work until the old way is removed in v0.2.0.

Should we adapt one example to work with the scenario class?

uvchik:
> Should we adapt one example to work with the scenario class?

By now the old variant is marked as deprecated, so we have to adapt the examples. But I would keep the old examples only for the example tests, to make sure the function works until we remove it.

ckaldemeyer:
> Should we adapt one example to work with the scenario class?
> By now the old variant is marked as deprecated, so we have to adapt the examples. But I would keep the old examples only for the example tests, to make sure the function works until we remove it.

:+1:

uvchik:
@oemof/oemof-developer-group : I hope we reached a consensus in this point. I do not want to spent time just to revert my changes afterwards:

 * NodesFromCSV is deprecated and will be removed in v0.2.x
 * It will be replaced by the new `Scenario` class and the `nodes_from_df` function.
 * the API will be kept until v0.2.x
 * by now v0.2.x is not scheduled

BTW: You will have to adapt renpassG!S until v0.2.x.

simnh:
I finally had a look at it and I think its a strong improvement. In particular adding other attributes etc is easier I assume. The only thing I do not like so much is the subclassing from EnergySystem. I have some things to improce as well, as there might be additional attributes in the future. Can we merge this one soon into dev?

simnh:
BTW is there an example on how to use the Scenario class? can we adapt the existing one / provide a new one . @uvchik  I assume you have files ? Can you just provide a simple example in the examples directory?

uvchik:
1. I agreed with @ckaldemeyer to adapt the existing examples but these are just examples for "how to use an existing csv file". My de21 application shows how to build a scenario from scratch. I can add an example on how to do that soon.

2. From my point of view the `Scenario` class is an extension of the `EnergySystem` class and therefore an inheritance makes sense. In [this comment](https://github.com/oemof/oemof/pull/294#issuecomment-288073554) you can see what will happen if we do not inherit. The `Scenario` object includes all nodes after you call the `create_nodes` method. For me this is coherent.

uvchik:
If you agree with this approach, I will adapt the existing example and merge soon. I will add the new example until the User Meeting but I think it is no problem to start a new PR for that and merge this one before.

simnh:
> From my point of view the Scenario class is an extension of the EnergySystem class and therefore an inheritance makes sense. In this comment you can see what will happen if we do not inherit. The Scenario object includes all nodes after you call the create_nodes method. For me this is coherent.

I do really not want you to merge this with the inheritance...If you remove this @ckaldemeyer and me are fine I guess. Please lets not discuss any longer. Lets just take this a compromise...

uvchik:
For me it does not make sense without the inheritance. It is confusing and not coherent.

We can discuss it on the WebCo next Tuesday if you like. If we do not find a compromise I can offer to move it to my application and close this umerged.

simnh:
> Could you tell me why you do not like sub-classing in this case.

1) The Scenario Class does not need/use any method from the scenario class
2) The Energysystem should hold the scenario (multiple ones at best). Then you could run (future)
```
es = EnergySystem()
es.scenarios = [s for s in scenarios]
es.run_scenarios() 
```
It's neither coherent to subclass from EnergySystem nor do not subclassing, we just do not have any structure for this now. Therefore, I think the best way is to replace the csv-reader for now and then look at a nice concept later on...I just prefer on step after the other. Otherwise we end up with coupled functionalities that we can not decouple in the future easily. 

Can you at least add the example...? If you want a subclassed scenario class you can do that on app level as well, but we should add the functionality of the new class soon. This discussion (again) blocks progress without any need, just because you do not want to take the 'safe' solution...

uvchik:
After I talked to @simnh we found the following compromise. We will merge the class as an "hidden" class. We will use the next time to test it and talk about the overall solph.input concept. After that we will embed the Scenario class in this concept and remove the "hidden" status. 

Therefore I did the following.

* marked the class as "experimental"
* temporarily removed the FutureWarning from NodesFromCSV as the Senario class it is not an official replacement so far.
* did not change the csv-reader examples
* did not add new examples
* did not change the attribute names (`p` and `s`) as we may change them in the next revision of the class
* added a docstring with a small example


@simnh or @ckaldemeyer: Could you test whether the following examples works for you in replacement for the `NodesFromCSV` function.

```python
import os
import oemof.solph as solph
my_path = os.path.join('my_folder', 'my_scenarios')
my_name = 'cool_scenario_name'
my_timeindex = pd.date_range('2012-01-01', '2012-12-31', freq='60min')
my_scenario = solph.Scenario(scenario_path=my_path, name=my_name)
my_es = solph.EnergySystem(timeindex=my_timeindex)
# The following works if the files are named `my_name`-based:
# in this case: "cool_scenario_name.csv" and "cool_scenario_name_seq.csv"
# Otherwise you have to pass the full path to the read_tables method. 
# read_tables(self, parameterfile='/path/to/p_file.csv', sequencefile='path/to/s_file.csv')
my_scenario.read_tables() 
my_scenario.create_nodes()
om = OperationalModel(my_es)
```

I rather would define the `EnergySystem` add the end but this does not seem to work. Or do I think in a wrong way.

```python
# ... (see above but without 'my_es')
nodes = my_scenario.create_nodes()
my_es = solph.EnergySystem(timeindex=my_timeindex)
my_es.add(nodes)
om = OperationalModel(my_es)
```

If I get an "approved" from both of you I will merge :smile:

uvchik:
To merge or not to merge, that's the question? Otherwise I will close this branch and transfer it to my application.

@gnn had an idea how to integrate this class better, but this will be a new PR.

uvchik:
It is pending since April, so I will move it to my application.

PR #295, opened 2017-03-22
Remove index levels from column index

ckaldemeyer:
This is only a one-liner that removes the levels from the column index when using the newly introduced bus balance slice and should be mergeable directly.


ckaldemeyer:
We could also integrate it into the current release..

uvchik:
If it is tested you can merge it into dev and I will integrate it into the release.

ckaldemeyer:
> If it is tested you can merge it into dev and I will integrate it into the release.

Should be fine as it only affects the labels..

ckaldemeyer:
Thanks!

PR #296, opened 2017-03-22
read ResultsDataFrame from file

uvchik:
With this PR it will be possible to instantiate an outputlib.ResultsDataFrame not only from an energy-system but also from a csv-result-file.


uvchik:
All test are okay. I fixed the docstrings.

PR #297, opened 2017-03-27
Features/improve result dataframe

ckaldemeyer:


ckaldemeyer:
Why is the outputlib actually located on the core level?

I know it was planned like this but at the moment it is absolutely solph specific from my point of view as the result dataframe completely relies on the result dict.

@uvchik : Should we move it down into `/solph` ? When changing the inits it does not change the API..

ckaldemeyer:
I have also moved the result-dict and -dataframe (+dataframeplot) into single modules within the outputlib-package since this is cleaner in my opinion. Since I have adjusted the inits, it does not breaks anything and the tests are still running well

ckaldemeyer:
I can then implement the new RDF structure in a similar fashion!

simnh:
because it it was designed for all 'level' (core , solph etc) ...at least should be, especially in future, when the results dict is removed...

ckaldemeyer:
> because it it was designed for all 'level' (core , solph etc) ...at least should be, especially in future, when the results dict is removed...

That's what I wrote above. The question is what "future" means and what will happen there. At the moment the place is only for architectural reasons and nothing actually happens at the core.

For me, the inputlib and output could be located on the same level, or not?

What about this:

* Leaving the `outputlib` package the the core level with an init that imports modules 
* Having a `solph.outputlib` package
* Planning an `inputlib` package at the core layer in a similar fashion

Or what do you think?

ckaldemeyer:
My quick ideas for a generic/solph specific outputlib/result structure:

* Core-layer: thinking in terms of data structures e.g. sequences, scalars, ... that can be related to each node type
* Solph-layer (inheriting properties of core layer to be compatible): Thinking in terms of information that can be related to single node types e.g. decision variables, parameters, ...
  * Do we want to save decision vars only or also add params?
  * Should other params such as efficiencies, etc. be integrated as well?
* Is there anything that could be planned that will break this structure?

Asking/answering questions about potential connected information to entities might help in finding the structure:

* A `Flow` can have: scalars (investment results, ...), sequences (actual values, ...)
* A `Component` can have: scalars (investment results, ...), sequences (socs, ...)
* A `Bus` can have: scalars (a georeferenced location), sequences (prices, temperatures, ...)

I just think that we should move forward with the issue of structuring the results since it is an issue we have been discussing about for ages now and we will not move forwards if we do not agree on something :smile:

ckaldemeyer:
At some stage I sometimes think the original OOP idea of writing results back into the nodes was not the worst one as it is very clear and one could also use structures directly, e.g. via pandas structures:

   MyFreakyTransformer.some_special_attribute_which_could_be_a_var_or_param.plot(kind='line', title='bla')

But let me know what you think... ;-)

ckaldemeyer:
@s3pp You are also invited to discuss ;)

uvchik:
I problem is, that the flows are created implicitly and it is not so easy to reach them.

The second thing is that I really like the DataFrame. It is easy to store and easy to handle. You can slice and plot easily.

What about the following structure: `source, target, attribute, datetime`

**Examples:**

Storage: `label, label, 'capacity', :`
Storage: `label, label, 'investment', :`
Flow: `source, target,:,:`

We could also reduce the number of index columns to three, but it might be less clear:

`level1, level2, level3`
**Example:**
Storage: `label, 'capacity', :`
Storage: `label, 'investment', :`
Flow: `source, target,:`

uvchik:
In the second case `target` is the attribute for the flows. That might be confusing.

uvchik:
But actually the first variant is also not totally clean. But I think we can get used to both variants. And DataFrame are fast and convenient.

ckaldemeyer:
> I problem is, that the flows are created implicitly and it is not so easy to reach them.
> 
> The second thing is that I really like the DataFrame. It is easy to store and easy to handle. You can slice and plot easily.

I agree on this. It's nice for analysing, etc.

> But actually the first variant is also not totally clean. But I think we can get used to both variants. And DataFrames are fast and convenient.

I am still struggeling with the difference between 'some_attribute' and 'investment' since it seems to be a `can have exactly one` relation, so investment seems to be an an attribute-related option.

uvchik:
> I am still struggeling with the difference between 'some_attribute' and 'investment' since it seems to be a can have exactly one relation, so investment seems to be an an attribute-related option.

I do not understand what you mean.

ckaldemeyer:
> I am still struggeling with the difference between 'some_attribute' and 'investment' since it seems to be a can have exactly one relation, so investment seems to be an an attribute-related option.
> I do not understand what you mean.

Let's say the `nominal_value` of a flow or `capacity` of a storage (both attributes) can be optionally invested in but not mandatory. So it's a kind of attribute-related option.

ckaldemeyer:
Yesterday I had another idea to combine the core/solph network classes and postprocessing.

What about views that are also subclassed according to the network classes?:

    MyCoreComponent(object) # core level
    MySolphComponent(MyCoreComponent) # solph level

    MyCoreView(object) # core level / ES as argument / could also be a function..
    MySolphView(MyCoreView) # solph level / e.g. ES/OM as argument / could also be a function..

This each layer can develop its own classes and views independently..

simnh:
My preference would look like this right now: 

Timeseries (all arrays): 
```
source, target, attribute, datetime
```
attribute could then be flow but also co2_emissions, costs, status variables from MILP etc. , we could then agree on specific names. but just start with the flow ; ) 

scalars: 
```
source, target, attribute, datetime
```
for all component associated thing just take source and target the same object....

That's my proposal, so we would end up with two dataframes. I do not know if the handling is good by using the source,target thing, but it seems the most generic way a the moment, otherwise we would end up with more dataframes ...

If you want to do it differently, its ok for me as well, there are many good ways to do it that we discussed above I think @ckaldemeyer will do it right! 

Though I think resolving the problem of timeseries and scalars by using two dataframes would be good...

uvchik:
I mostly agree with @simnh. It is similar to what we have and we know that it works fine.

I just would prefer one DataFrame. In that case scalars would just have one entry in the `datetime` column (e.g. the year).

s3pp:
I also would prefer to have (a) DataFrame(s) in the end, for accessibility reasons only. With regard to the different proposals. I slightly prefer the one and only DataFrame approach, because having only one DataFrame is really convenient for results handling, but I'm not so sure about kind of highjacking the datetime index for differentiating between scalars and timeseries data. What about an additional boolean field in the index for doing just that? As a consequence and on the downside obviously datetime would always hold some NaT entries and all slicing would have to include the extra field.

ckaldemeyer:
I have just talked to @gnn and he already sees some problems with the current idea. So we put this on hold until the openmod conference.

ckaldemeyer:
> I have just talked to @gnn and he already sees some problems with the current idea. So we put this on hold until the openmod conference.

But should we already merge back the current PR as it ports the results-dict into a single package and does not change any functionality?

I think it is already a slight improvement of the structure and the new RDF (with some fancy structure) can be realized in the same manner :smile:

uvchik:
Could you first integrate PR #296? 

It is pending since three weeks and I didn't merge it just because there was no feedback. It will collide with your changes.

ckaldemeyer:
> Could you first integrate PR #296?
> 
> It is pending since three weeks and I didn't merge it just because there was no feedback. It will collide with your changes.

I have approved the PR

uvchik:
@ckaldemeyer Thank you, but now you have the conflict - sorry. :smirk:

ckaldemeyer:
> @ckaldemeyer Thank you, but now you have the conflict - sorry.

I'll deal with it :)

uvchik:
...but overall I like you restructure. You may fix the conflict and merge. I'm fine.

ckaldemeyer:
> ...but overall I like you restructure. You may fix the conflict and merge. I'm fine.

See https://github.com/oemof/oemof/pull/303

PR #301, opened 2017-04-04
Features/add graph tools

ckaldemeyer:
This adds the possibility to create and plot [networkx](http://networkx.readthedocs.io/en/networkx-1.11/) graphs in [dot graphviz](http://www.graphviz.org/) layout from a given energy system and operational model.

The returned graph can be exported, changed, ...

I wasn't sure if it's located right on the core level since it needs an opt model to work properly. But this is the case for the result-df, too and so I put it in the outputlib at the moment.

### DIFF
--- a/oemof/outputlib/graph_tools.py
+++ b/oemof/outputlib/graph_tools.py
@@ -1,9 +1,16 @@
 # -*- coding: utf-8 -*-
 """Modules for creating and manipulating energy system graphs."""
 
-import pylab as plt
-import networkx as nx
-from networkx.drawing.nx_agraph import graphviz_layout
+import logging
+try:
+    import pylab as plt
+except ImportError:
+    logging.warning('Pylab could not be imported. Plotting will not work.')
+try:
+    import networkx as nx
+    from networkx.drawing.nx_agraph import graphviz_layout
+except ImportError:
+    logging.warning('Networkx could not be imported. Plotting will not work.')
 
 
 def graph(energy_system, optimization_model, edge_labels=True,
@@ -47,7 +54,6 @@ def graph(energy_system, optimization_model, edge_labels=True,
     Needs graphviz and networkx (>= v.1.11) to work properly.
     Tested on Ubuntu 16.04 x64.
     """
-
     # construct graph from nodes and flows
     G = nx.DiGraph()
     for n in energy_system.nodes:
--- a/oemof/outputlib/graph_tools.py
+++ b/oemof/outputlib/graph_tools.py
@@ -91,7 +91,7 @@ def graph(energy_system, optimization_model, edge_labels=True,
     pos = graphviz_layout(G)
     nx.draw(G, pos=pos, **options)
 
-    # add edge labels for all edges or a defined subset
+    # add edge labels for all edges
     if edge_labels is True:
         labels = nx.get_edge_attributes(G, 'weight')
         nx.draw_networkx_edge_labels(G, pos=pos, edge_labels=labels)
--- a/oemof/outputlib/graph_tools.py
+++ b/oemof/outputlib/graph_tools.py
@@ -36,6 +36,7 @@ def graph(energy_system, optimization_model, edge_labels=True,
     optimization_model : `oemof.solph.models.OperationalModel`
 
     edge_labels: Boolean
+        Use nominal values of flow as edge label
 
     remove_nodes: list of string
         Nodes to be removed e.g. ['node1', node2')]
--- a/oemof/outputlib/graph_tools.py
+++ b/oemof/outputlib/graph_tools.py
@@ -1,0 +1,97 @@
+# -*- coding: utf-8 -*-
+"""Modules for creating and manipulating energy system graphs."""
+
+import pylab as plt
+import networkx as nx
+from networkx.drawing.nx_agraph import graphviz_layout
+
+
+def graph(energy_system, optimization_model, edge_labels=True,
+          remove_nodes=None, remove_nodes_with_substrings=None,
+          remove_edges=None, node_color='#AFAFAF', edge_color='#CFCFCF',
+          plot=True, node_size=2000):
+    """
+    Draw a directed graph for the passed energy system.
+
+    Parameters
+    ----------
+    es : `oemof.solph.network.EnergySystem`
+
+    om : `oemof.solph.models.OperationalModel`
+
+    edge_labels: Boolean
+
+    remove_nodes: list of string
+        Nodes to be removed e.g. ['node1', node2')]
+
+    remove_nodes_with_substrings: list of substrings
+        Nodes that contain substrings to be removed e.g. ['elec_', 'heat_')]
+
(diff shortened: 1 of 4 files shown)


ckaldemeyer:
Per standard it draws the nominal values and has greyish colors: http://pasteboard.co/dClTjdII.png

ckaldemeyer:
I have added a doctest now but have problems with multiple lines in the example code:

    Examples
    --------
    >>> datetimeindex = pd.date_range('1/1/2017', periods=3, freq='H')
    >>> es = EnergySystem(timeindex=datetimeindex)
    >>> b_gas = Bus(label='b_gas', balanced=False)
    >>> b_el = Bus(label='b_el')
    >>> demand = Sink(label='demand_el', \n
        inputs={b_el: Flow(nominal_value=85,
        actual_value=[0.5, 0.25, 0.75], \n
        fixed=True)})
    >>> pp_gas = LinearTransformer(label='pp_gas', \n
        inputs={b_gas: Flow()}, \n
        outputs={b_el: Flow(nominal_value=41, variable_costs=40)})
    >>> om = OperationalModel(es=es)
    >>> my_graph = graph(energy_system=es, optimization_model=om, plot=False)
    >>> print(my_graph.nodes())
    ['demand_el', 'b_el', 'pp_gas', 'b_gas']

Error:

	File "graph_tools.py", line 78, in __main__.graph
	Failed example:
	    print(my_graph.nodes())
	Expected:
	    ['demand_el', 'b_el', 'pp_gas', 'b_gas']
	Got:
	    ['b_el', 'b_gas']

Does anyone know how to fix the multi-line commands e.g. `demand = Sink(label='demand_el',` ?

I couldn't figure it out quickly and now I am off surfing ;)

uvchik:
```python
>>> import package as pack
>>> if pack.x == 13:
...     print("something")
... else:
...     print("something else")
```

As far as I know the imports belong to the test not to the module

uvchik:
What do you think of 1a6202e319a31b7b5ec987? In this case the dispatch example will not exit if networkx is not installed.

ckaldemeyer:
> What do you think of 1a6202e? In this case the dispatch example will not exit if networkx is not installed.

:+1:

ckaldemeyer:
> >>> import package as pack
> >>> if pack.x == 13:
> ...     print("something")
> ... else:
> ...     print("something else")
> As far as I know the imports belong to the test not to the module

I have now fixed the doctest, changed the solph imports to the docstring and added an example on how to export to .graphml.

ckaldemeyer:
One more idea as the module is currently located at the core level but needs a solph.OperationalModel (actually it should also work with all models):

* Should we make it work with the `es` (core nodes only and no flows) and make the `om` from solph optional?
* Or should we just move it one level deeper into the solph hierachy?

uvchik:
1. I do not have the time to change it, but it would be nice if the Flows would be part of the EnergySystem

2. I would prefer to make the OM optional.

3. It is difficult to write a test for an optional function. We should ask @gnn how to write a appropriate test that will have one result if networkx is present and another result if not.

ckaldemeyer:
> I do not have the time to change it, but it would be nice if the Flows would be part of the EnergySystem

I agree as it is the weight of the edge but also do not want to spend time on it.

> I would prefer to make the OM optional.

I'll have a look at that.

> It is difficult to write a test for an optional function. We should ask @gnn how to write a appropriate test that will have one result if networkx is present and another result if not.

:+1: Good idea.

ckaldemeyer:
> I do not have the time to change it, but it would be nice if the Flows would be part of the EnergySystem
> 
> I would prefer to make the OM optional.

It is now possible to use it only with an `oemof.EnergySystem` but then the edges do not have arrows but it just shows the topology.

ckaldemeyer:
TODOs:

- [ ] Check how to test optional modules with @gnn  and adapt (doc)test
- [ ] Adapt example usage
- [ ] Update WHATSNEW

uvchik:
You can skip the doctest and use it just as an example. Then you can write a normal test.

In the pvlib they skip tests in case the package is not installed.

conftest.py:

```python
try:
    import scipy
    has_scipy = True
except ImportError:
    has_scipy = False

requires_scipy = pytest.mark.skipif(not has_scipy, reason='requires scipy')
```

some_test_file.py:
```python
from conftest import requires_scipy

@requires_scipy
def some_test():
    something happens
```

ckaldemeyer:
> You can skip the doctest and use it just as an example. Then you can write a normal test.
> 
> In the pvlib they skip tests in case the package is not installed.
> 
> conftest.py:
> 
> try:
>     import scipy
>     has_scipy = True
> except ImportError:
>     has_scipy = False
> 
> requires_scipy = pytest.mark.skipif(not has_scipy, reason='requires scipy')
> some_test_file.py:
> 
> from conftest import requires_scipy
> 
> @requires_scipy
> def some_test():
>     something happens

So your proposal would make travis happy but I have to create an external test and remove the self contained one?

I quite like the self-contained one, especially for optional modules. Isn't there an easy way to make travis happy by excluding this doctest on the server. Would this also be a compromise?

ckaldemeyer:
Anyway, thanks for your suggestion and help!

uvchik:
You have to create a test but the test could be equal to the doctest. You do NOT have to remove the doctest. You can just add an `+ SKIP` to skip it.



``` 
try:
    import networkx as nx
    from networkx.drawing.nx_agraph import graphviz_layout
except ImportError:
    nx = None
    graphviz_layout = None
    logging.warning('Networkx could not be imported. Plotting will not work.')

....something

if nx is not None: 
    my_graph = graph(energy_system=es, optimization_model=om, plot=False)
    [my_graph.has_node(n) for n in ['b_gas', 'b_el', 'pp_gas', 'demand_el']]  # doctest: +SKIP
[True, True, True, True]
```

Normally `+SKIP` is enough but `# doctest: +SKIP` hides the command in the sphinx documentation.

oemof/outputlib/graph_tools.py
```diff
@@ -0,0 +1,169 @@
...
+"""Modules for creating and manipulating energy system graphs."""
+
+import logging
+import warnings
+
+try:
+    from matplotlib import pyplot as plt
+except ImportError:
+    logging.warning('Matplotlib could not be imported.',
+                    ' Plotting will not work.')
```
uvchik: You should set `plt = None` in the except case. This is a common practice because it is more difficult to catch an undefined variable than asking if it is None. And it is much easier to find bugs.

As matplotlib is a requirement of oemof, the whole try/except statement might not be necessary.

gnn:
So, I've added code to the end of the file to conditionally change the docstring of the problematic functions. If the necessary imports aren't there, the test is now skipped.
If you two think that this is a good way to go about it, merge the PR.

uvchik:
I can not overview the implications but as it is not a main oemof function I am totally fine with trying it as it is. If a problem occurs we can change it again.

uvchik:
Thank you for this feature :tada:

PR #303, opened 2017-04-12
Features/put result dict and df into package

ckaldemeyer:
This only holds the first changes of the big dataframe refactoring (https://github.com/oemof/oemof/pull/297) which put the result-dict and -df into single modules in the outputlib package.

The merge conflicts with https://github.com/oemof/oemof/pull/296 should be fixed and tests are running well.

@uvchik : Where did you keep the notes about how to test properly after the problems with the last release? I don't remember it anymore but know you kept it somewhere..

PR #307, opened 2017-04-22
First draft for writing results back to components

simnh:


ckaldemeyer:
If we decide on this, we could also think of putting this into the outputlib as I did with the RDF/result dict in https://github.com/oemof/oemof/pull/303

simnh:
Closed in favour of #303

PR #308, opened 2017-04-26
Features/multiperiod

simnh:
I do need a new functionality and I would like to integrate that in oemof solph. The basic idea here is the following; 

Change the timeindex for t=[0, 1, 2, etc...] to (a,t) = [(0, 1), (0, 2) .., (1, 5)...]. We will keep timesteps, so if there are two years to be calculated timesteps would be 0...8760*2 other index sets will be: 

* TIMEINDEX: (a,t) 
* PERIODS:  
* PERIOD_TIMESTEPS: the sliced timesteps for each period e.g. for a=0 -> 0...8769, for a=1 -> 8760-> 8760*2 -1

As there will still be the TIMESTEP index the results etc will not be affected. So everything works as it has before. 

There is one new functionality: summed_may/min per period (Of course the does not need to be a year, we can keep it more flexible e.g month...)

Tests work, bu lp-files are broken of course...

This pull request should only change the index to allow further development.
We can start of with this minimal solution and add extra functionality (s. below) based on this, e.g. get specific period bounds on storages, look how we use the investment for this (investment per period or the complete timesteps)

TODO: 

* [ ] Update constraints documentations 

For the future: 
* [ ] Add discount per period to the objective expressions? (decide where to put that, I assume best would be in the energysystem 
* [ ]  Allow to do investment per period (consider lifetime in Investment object, max invest rate, init lifetime etc) 
* [ ] Add periodincrement

### DIFF
--- a/examples/solph/flexible_modelling/add_constraints.py
+++ b/examples/solph/flexible_modelling/add_constraints.py
@@ -74,22 +74,22 @@ def run_add_constraints_example(solver='cbc', nologg=False):
     # add the sub-model to the oemof OperationalModel instance
     om.add_component('MyBlock', myblock)
 
-    def _inflow_share_rule(m, s, e, t):
+    def _inflow_share_rule(m, s, e, a, t):
         """pyomo rule definition: Here we can use all objects from the block or
         the om object, in this case we don't need anything from the block except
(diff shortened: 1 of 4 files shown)


simnh:
closed in favour of #303

PR #309, opened 2017-04-26
Graph tools: Add possibility to add colors to nodes

ckaldemeyer:
This allows for different colored nodes: http://pasteboard.co/8SuYcdy9x.png

Should be mergeable directly..


ckaldemeyer:
I was thinking about adding the functionality to write variable and fixed costs or generally different flow attributes on the edges by passing some iterable with the attributes and additional units (MW, EUR/MWh, EUR/MJ, ...) to be drawn.

Thus, if one wants one can draw a full descriptive graph with resource costs, installed capacities and so on.

@uvchik : What do you think?

simnh:
did it, comes with version 0.2 but on #329

ckaldemeyer:
Nice, thanks! I knew I had started it somewhere :-)

PR #317, opened 2017-05-26
Update installation_and_setup.rst

drjod:
dont know if (solph) is correct. Only tried with cbc


simnh:
yes thats solph...are there more changes to come? Otherwise we can merge...

PR #319, opened 2017-05-26
Update developing_oemof.rst

drjod:
das hatte nicht funktioniert

### DIFF
--- a/doc/developing_oemof.rst
+++ b/doc/developing_oemof.rst
@@ -20,7 +20,7 @@ To avoid problems make sure you have fully uninstalled previous versions of oemo
 
 .. code:: bash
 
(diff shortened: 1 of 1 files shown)


uvchik:
Both ways should work. You need an ssh-key to use it: https://help.github.com/articles/which-remote-url-should-i-use/

We could just write "Clone the repository" and add a link to the following help page: https://help.github.com/articles/cloning-a-repository/

uvchik:
I can fix this when we revise the documentation at the meeting right before the release.

ckaldemeyer:
> I can fix this when we revise the documentation at the meeting right before the release.

:+1:

uvchik:
Fixed with commit 90a30afaef0369.

PR #321, opened 2017-06-14
Add youtube tutorial to documentation

birgits:


birgits:
Did we want to add the comment "SEND US YOUR EXAMPLES" to getting_started/examples? I forgot...

uvchik:
We may want to create a folder structure in oemof.examples and let the people publish their examples by PRs. But we should discuss this separately.
```
--oemof_examples
----v0.1.x
------notebooks
------python_scripts
----v0.2.x
```

birgits:
We could still let people know that their examples are welcome and should be introduced by a PR in oemof.examples without having a final structure? I guess otherwise we won't get any examples.

uvchik:
> We could still let people know that their examples are welcome and should be introduced by a PR in oemof.examples without having a final structure? I guess otherwise we won't get any examples.

Feel free. Should we also offer to send it by e-mail for non-github-users?

ckaldemeyer:
thanks, birgit!

birgits:
About sending the example by e-mail - did we ever discuss having one oemof e-mail address (e.g. contact@oemof.org) now that we have our own domain?

I could add my notebook from the user meeting to the new examples repo if it's good enough for a first example? Or does anyone else have a good example we could add?

ckaldemeyer:
> About sending the example by e-mail - did we ever discuss having one oemof e-mail address (e.g. contact@oemof.org) now that we have our own domain?

I think we only have the contact page on oemof.org. But that should be fine from my point of view.

> I could add my notebook from the user meeting to the new examples repo if it's good enough for a first example? Or does anyone else have a good example we could add?

I don't. But I would appreciate if you upload yours!

PR #325, opened 2017-07-25
Remove datetimeindex from OperationalModel

simnh:
I looked at the examples and at the documentation and we only use the datetimeindex inside the EnergySystem so I like to remove it from OM


uvchik:
Looks consistent. Thanks.

Issue #292, opened 2017-03-14
Throw error if problem is infeasible


simnh:
At the moment the solve method ot the OperationalModel throws an error when the problem is infeasible, but this error comes from the result() method which is called. 

This can be extend to the full information from the solver BTW,..

uvchik:
:+1: 

Good idea. We could catch the error and raise an Infeasibility error/warning. This could be more specific than the actual error.

simnh:
I will do this on the dev today...

PR #314, opened 2017-05-10
Added download links for the documentation

stianchris:
As the file for the documentation is a bit hidden, I included the links and a little description on how to find other versions.

PR #329, opened 2017-07-26
Restructuring of outputlib and custom components

ckaldemeyer:
Some leftovers/reminders/questions/todos: from #303:

* [x] Should we rename the modules to "results_to_x" instead of "results_x" ? (Like I did for components?)
* [x] Write to components needs to be checked for correct 
* [x] We should a check in the OperationModel, if the problem is infeasible, right now there only comes an error when writing results, that not exist. ( see: #292 ) 
* [x] Write checks inside the modules, if results actually exist (I called on a not-solve instance-> throw error.
* [ ] Adapt docstring in results to dict -> there is still something in from before refactoring (DispatchSource etc.)

* [x] Adapting or removing old result-features (plots, slices,...)

* [x] Adapt `om.results()` method to new result object.

* [x] Adapt all examples:
    * [x] storage_invest ( @uvchik )
    * [x] csv-reader ( @ckaldemeyer )

And some that are related to the general restructuring

* [x] Think about creating the result dataframe directly and skipping the result-dict. @gnn already said that he would have no objections as the result-dict was something that came out of the box spontaneously

* [x] How to import custom components? Add them to `solph/__init__.py` or always do something like `from oemof.solph.custom import MyComp`?

* [x] Should we remove the standard Storage and only keep the GenericStorage?

* [ ] Revise all custom docstrings and make them uniform


ckaldemeyer:
https://github.com/oemof/oemof/issues/315

ckaldemeyer:
https://github.com/oemof/oemof/issues/252

ckaldemeyer:
https://github.com/oemof/oemof/issues/251

ckaldemeyer:
Is https://github.com/oemof/oemof/pull/326 related to this e.g. to expansion model results require another structure than investment ones @simnh ?

From the code and my guess I would think not..

simnh:
I am not sure...but i do not think so, same problems as everyhwere, timeseries for flows, scalar for flows and or components...as I used the timeindex the index does not need to change because then there are just different years inside the index...so, summarizing if we find a good way for all results we currently have, expansion should not add any additional problems...

ckaldemeyer:
> I am not sure...but i do not think so, same problems as everyhwere, timeseries for flows, scalar for flows and or components...as I used the timeindex the index does not need to change because then there are just different years inside the index...so, summarizing if we find a good way for all results we currently have, expansion should not add any additional problems...

:+1:

ckaldemeyer:
> I am not sure...but i do not think so, same problems as everyhwere, timeseries for flows, scalar for flows and or components...as I used the timeindex the index does not need to change because then there are just different years inside the index..

If we have different investments in different periods, shouldn't it also lead to a sequence (of investments)?

Then, we again have the option to use the start/end date(s) as an index to store the scalar value(s) or just use the whole period(s) as an index with a constant value. I think in terms of require storage space it is not a big deal as pandas internally holds references (not call by value)..

ckaldemeyer:
More historical discussion: https://github.com/oemof/oemof/pull/297

ckaldemeyer:
@simnh: I have added a quick and dirty prototype for the new dict/df structure with an exemplary usage in the storage invest example. Duals, etc. do not work but can be added later on if the structure holds for all cases.

The structure already has several advantages over the standard multi-indexed df that lie in the capability for potential individual components, etc.

Not surprising if the result data structure becomes more similar to the object data structure. But it might be a good compromise. Let's discuss it on Monday :smile:

ckaldemeyer:
Investment results are already working. The code is not really nice but can be easily cleaned..

ckaldemeyer:
I'll continue here next week but think could work well :-)

ckaldemeyer:
It took some time but now I am finished with my first proposal!

Here's what the (possible) new API looks like:

    # create a dictionary with the results
    results = results_to_dict(energysystem, om)
    
    # access the results: id=(node,) for nodes, id=(node,node) for flows
    # scalars are saved within a pandas series
    # sequences are saved within a pandas dataframe
    results[id]['scalars']
    results[id]['sequences']

It integrates all results including investment results for nodes and flows should be more flexible as things do not have to be integrated by hand. Block variables are collected from Pyomo and integrated within the structure which means that data for arbitrary nodes is included automatically as long as the classes/blocks inherit from `oemof.Node`.

The are two examples that illustrate how to use the API which also include some basic plots:

* Storage Investment Example
* CSV-Reader Investment Example

These should be reset later because I have not adapted the tests and checks fail..

Generally, discussion and contribution is very welcome. I am not sure about the right name and also think that some pieces e.g. the double-created dict might be improved. But it's already working and I will first go on holiday until 19th of September ;-)

@simnh , @uvchik , @gnn : Let me know what you think and go ahead here if you have ideas or want to finish for the solph-paper. I am happy about any progress!

Cheers

ckaldemeyer:
I have just seen that I haven't integrated the ideas/TODOs from above. But they are of general nature, anyway and could be added to any structure..

ckaldemeyer:
Again in progress :-)

Actually I am thinking about merging https://github.com/oemof/oemof/pull/332 into this branch as I could already check the functionality for custom components..

simnh:
> Again in progress :-)

> Actually
I am thinking about merging #332 into this branch as I could already check the functionality for custom components..

yeah why not? go for it....we can then continue in this branch will this improvment as well if possible...that maybe cleans up some of the branch mess :)

ckaldemeyer:
I have now added an exemplary slicing function which extracts all data that is related to a Node (Component/Bus) and an example can be found in the [csv investment example](https://github.com/oemof/oemof/blob/features/restructure-outputlib/examples/solph/csv_reader/investment/investment.py). 

General usage:

    # create a dictionary with the results
    results = results_to_dict(es, om)

    # standard api: results for a flow
    my_id = (es.groups['REGION1_pp_oil'], es.groups['REGION1_bus_el'])
    results[my_id]['scalars']
    results[my_id]['sequences']

    # standard api: results for a component
    my_id = (es.groups['REGION1_storage_phs'],)
    results[my_id]['scalars']
    results[my_id]['sequences']

    # slicing functions: get all node results (bus)
    region1 = node_results(results, es.groups['REGION1_bus_el'])
    region1['sequences']
    region1['scalars']

    # slicing functions: get all node results (component)
    phs = node_results(results, es.groups['REGION1_storage_phs'])
    phs['sequences']
    phs['scalars']

    # example plot for sequences
    phs['sequences'].columns = ['P-IN', 'P-OUT', 'CAP']
    ax = phs['sequences'].plot(kind='line', drawstyle='steps-post')
    ax.set_title('Dispatch results')
    ax.set_xlabel('Time')
    ax.set_ylabel('Power (MW) / Energy (MWh)')
    plt.show()

    # example plot for scalars
    phs['scalars'].index = ['P-IN', 'P-OUT', 'CAP']
    ax = phs['scalars'].plot(kind='bar')
    ax.set_title('Investment results')
    ax.set_xlabel('')
    ax.set_ylabel('Storage investment in MWh / MW')
    plt.show()

I still have to add duals and of course there is space for improvements. Let me know what you think, then I'll continue!

ckaldemeyer:
We could also think about a decorator to be able to slice with strings directly, etc.. But I'll wait for feedback first ;-)

ckaldemeyer:
I have talked to @uvchik today and he generally likes the structure.

Here are some things we would like to do if you (@simnh, @gnn) also agree:

* Add duals in current structure
* Remove the old multi-indexed dataframe and the old result-dict completely from the code base
* Adapt all examples to use the new result structure
* Directly jump to V.0.2 for the next release which includes this feature, the custom components, expansion models and some other polishing
* Provide separate modules for the "core results" and "views of core results (slicing functions)"

@uvchik also gave feedback and ideas for discussion:

* Should we keep the 1D tuple index for components e.g. `my_id = (es.groups['REGION1_storage_phs'],)` or just replace it by the object itself `my_id = es.groups['REGION1_storage_phs']`? @gnn, @simnh: Do you know any technical pros/cons!? @uvchik and me have different preferences and would be interested in yours.
* Should we allow direct slicing with strings in the views module so that we do not have to use `es.groups` all the time?
* Should we include solver results e.g. status messages, matrix dimensions, .... in the result object?

@simnh, @gnn: Could you give some feedback? I would like to continue ASAP ;-)

ckaldemeyer:
Duals are now integrated and I have created single modules for the data structure and views..

simnh:
> I have talked to @uvchik today and he generally likes the structure.

> Here are some things we would like to do if you (@simnh, @gnn) also agree:
    Add duals in current structure
    Remove the old multi-indexed dataframe and the old result-dict completely from the code base
    Adapt all examples to use the new result structure
    Directly jump to V.0.2 for the next release which includes this feature, the custom components, expansion models and some other polishing
    Provide separate modules for the "core results" and "views of core results (slicing functions)"

yes to all points from gnn and my side (though I don't know, when I will be able to finish an test the expansion model). Especially, because we would need to agree to have the new timeindex, as on the expansion model feature branch, which is: [year, time step, source , target]. We could skip the expansion model functionality, though I would like to prepare for it by changing the index to the one I described above and to introduce a Base model, from which the 'new' models can be derived...we could also subclass the OperationModel, but we can also leave it like it is. maybe then rename it, as it is the more general model that combine investment, UC and dispatch...(if you understand what I mean) 

> @uvchik also gave feedback and ideas for discussion:
    Should we keep the 1D tuple index for components e.g.
    my_id = (es.groups['REGION1_storage_phs'],) or just replace it by the object itself
    my_id = es.groups['REGION1_storage_phs']? @gnn, @simnh: Do you know any technical pros/cons!? @uvchik and me have different preferences and would be interested in yours.
    Should we allow direct slicing with strings in the views module so that we do not have to use es.groups all the time?
    Should we include solver results e.g. status messages, matrix dimensions, .... in the result object?

I am not too much into the grouping thing, but the second one is more user-friendly...but actually, it should already work like the second way!?

> @simnh, @gnn: Could you give some feedback? I would like to continue ASAP ;-)

ckaldemeyer:
> yes to all points from gnn and my side (though I don't know, when I will be able to finish an test the expansion model). Especially, because we would need to agree to have the new timeindex, as on the expansion model feature branch, which is: [year, time step, source , target].

I think including source and target makes it not a pure timeindex. Is this meant to be the new index for all models?

So far, we have scalars (d=1) and sequences (d=len(timeindex/index set). But we could look into integrating that..

> We could skip the expansion model functionality, though I would like to prepare for it by changing the index to the one I described above and to introduce a Base model, from which the 'new' models can be derived...we could also subclass the OperationModel, but we can also leave it like it is. maybe then rename it, as it is the more general model that combine investment, UC and dispatch...(if you understand what I mean)

I would support proper names but wouldn't put effort in it from my side.

> I am not too much into the grouping thing, but the second one is more user-friendly...but actually, it should already work like the second way!?

Not, it doesn't. Only if you use references e.g. the objects themselves. I will try to find an easy solution for this.

ckaldemeyer:
I'll merge the custom components branch today and make some tests!

ckaldemeyer:
> yes to all points from gnn and my side (though I don't know, when I will be able to finish an test the expansion model). Especially, because we would need to agree to have the new timeindex, as on the expansion model feature branch, which is: [year, time step, source , target].

Actually year and time step are included in a timeindex and could be easily extracted. This seems to be kind of redundant to me (not normalized). But I guess you need the separate fields as index sets for the exp. model..

ckaldemeyer:
> yes to all points from gnn and my side (though I don't know, when I will be able to finish an test the expansion model). Especially, because we would need to agree to have the new timeindex, as on the expansion model feature branch, which is: [year, time step, source , target].
> I think including source and target makes it not a pure timeindex. Is this meant to be the new index for all models?
> 
> So far, we have scalars (d=1) and sequences (d=len(timeindex/index set). But we could look into integrating that..

I think everything works/is adaptable easily as long as we keep the same time index for different models. But I bet the expansion model will not work out of the box if the index differs from the regular one as it leads to other index tuples within pyomo which are processed. So either we have the same time index and one general result processing function or different time indices for different models and thus different result processing functions.

Nevertheless, the overall result structure could remain the same.

simnh:
Regarding the index: yes the index for all models, timeindex would switch from [t] to [p, t] (period, timestep)...of course source, target would stay as well for flows.

simnh:
> Not, it doesn't. Only if you use references e.g. the objects themselves. I will try to find an easy solution for this.

I am pretty sure that every node is stored under its label in `es.groups['label']`

simnh:
> Actually year and time step are included in a timeindex and could be easily extracted. This seems to be kind of redundant to me (not normalized). But I guess you need the separate fields as index sets for the exp. model..

yes thats right. for clarification: the timeindex would stay '1D' pandans index, but inside the model we would end up with `[period, timestep, source, traget]` which is currently` [timestep, source, target]`

ckaldemeyer:
> Not, it doesn't. Only if you use references e.g. the objects themselves. I will try to find an easy solution for this.
> I am pretty sure that every node is stored under its label in es.groups['label']

I think this is not the question. It only depends on how the result dict is keyed. I can change it and will try to provide both ways.

> Actually year and time step are included in a timeindex and could be easily extracted. This seems to be kind of redundant to me (not normalized). But I guess you need the separate fields as index sets for the exp. model..
> yes thats right. for clarification: the timeindex would stay '1D' pandans index, but inside the model we would end up with [period, timestep, source, traget] which is currently[timestep, source, target]

Let's check it when it's finished. The quick solution would be to check the model type and offer a different processing for the expansion model. Should be easy and I think the number of models is limited anyway, so from my side it would be fine.

uvchik:
> yes thats right. for clarification: the timeindex would stay '1D' pandans index, but inside the model we would end up with [period, timestep, source, traget] which is currently[timestep, source, target]

I do not fully survey the consequences for the existing models but if the default value for periods is one, it will not effect the existing models at all. So it seems to me that it is no problem to add the periods. I think it is not necessary to show the periods in the results, it might be enough to show the whole sequence (e.g. 1.1.2015-31.12.2025). The periods can be added afterwards. The only difference might be scalar-results given for every period. We might need to change the result API a little.

Actually I would prefer a web conference to talk about the upcoming release and what will be part of it and when are we going to release it. It is not a good idea to change the API too soon after the v0.2.0 release, so we should have a brief plan for these changes.

birgits:
Just as a reminder, the next regular oemof webco is scheduled for Oct 2nd. We could use it to talk about the roadmap if it isn't too late for you.

ckaldemeyer:
> Just as a reminder, the next regular oemof webco is scheduled for Oct 2nd. We could use it to talk about the roadmap if it isn't too late for you.

I won't be there and think many won't be since the 2nd of October is holiday. Should we postpone it?

uvchik:
> I won't be there and think many won't be since the 2nd of October is holiday. Should we postpone it?

-> [14](https://github.com/oemof/organisation/issues/14)

ckaldemeyer:
> yes thats right. for clarification: the timeindex would stay '1D' pandans index, but inside the model we would end up with [period, timestep, source, traget] which is currently[timestep, source, target]
> I do not fully survey the consequences for the existing models but if the default value for periods is one, it will not effect the existing models at all. So it seems to me that it is no problem to add the periods. I think it is not necessary to show the periods in the results, it might be enough to show the whole sequence (e.g. 1.1.2015-31.12.2025). The periods can be added afterwards. The only difference might be scalar-results given for every period. We might need to change the result API a little.
> 
> Actually I would prefer a web conference to talk about the upcoming release and what will be part of it and when are we going to release it. It is not a good idea to change the API too soon after the v0.2.0 release, so we should have a brief plan for these changes.

I think we should test and decide how to integrate the expansion model results in the result API before v.0.2. because it will be a quite major release.

@simnh: Do you know when you will be able to test/integrate the expansion model?

ckaldemeyer:
I have now merged the custom component branch and added some scalar and sequence test vars for the custom investment storage. They're automatically integrated in the new result structure seems to be working as expected! The only thing I still have to deal with is the naming of flows within the views. But the general result object should be already fine.

Additionally, I have added a flag `keys_as_strings` to create the result dict with strings tuples to allow easy access without grouping. So one can decide which one to choose.

Performance-wise, I have tested the speed of the new structure against the old one using renpassG!S: the old one is faster with 24 seconds against 110 seconds creation time of the new one. But I guess tweaking should be easy and it is still acceptable.

Generally, a webco would be cool to fix the whole naming (result_dict etc. is maybe not the best naming, etc.). When we have decided on certain things, I can implement the changes and finish everything.

@simnh : We should also check if the custom component structure is fine like this. If we are both fine, I would implement the generic CHP formulation from Mollenhauer et. al as a first custom component.

ckaldemeyer:
> The only thing I still have to deal with is the naming of flows within the views. But the general result object should be already fine.

Implementing this was somewhat of a brainfuck but it works now, too..

simnh:
Aren't the first custom components the storage? We should move variable fraction transformer also, or not? Or should be just give it a depreciated warning, as we will have the new chp components? I guess mollenhauer has no pure linear formulation, which would be good to have for large economic dispatch models as well, I would vote for extraction turbine then by the way..

simnh:
I think the custom sturcture is fine for now, we will see how it works with more complex and new components. does the storage example work BTW? Where are the custom_groupings added, should that not happen in the example somewhere? (I kind of lost track here)

simnh:
And we need to make sure, that if statements of 'standard' grouping and 'custom' grouping are compatiable (i.e. components are grouped correctly and not mixed up)

ckaldemeyer:
I have now moved the string keyed variant into the views module, changed the naming, etc..

Here's an example on how to use the new result object:

    # import
    from oemof.outputlib import results, node_results

    # create a dictionary with the results
    opt_results = results(es, om)

    # standard api: results for a flow
    my_id = (es.groups['REGION1_pp_oil'], es.groups['REGION1_bus_el'])
    # or directly with objects:
    # my_id = (REGION1_pp_oil, REGION1_bus_el)
    print(opt_results[my_id]['scalars'])
    print(opt_results[my_id]['sequences'].describe())

    # standard api: results for a component
    my_id = (es.groups['REGION1_storage_phs'],)
    # or directly with objects:
    # my_id = (REGION1_storage_phs,)
    print(opt_results[my_id]['scalars'])
    print(opt_results[my_id]['sequences'].describe())

    # slicing functions: get all node results (bus)
    # works with both node objects and string labels as argument
    region1 = node_results(opt_results, es.groups['REGION1_bus_el'])
    region1 = node_results(opt_results, 'REGION1_bus_el')
    print(region1['sequences'].max())
    print(region1['scalars'])

    # slicing functions: get all node results (component)
    # works with both node objects and string labels as argument
    phs = node_results(opt_results, es.groups['REGION1_storage_phs'])
    phs = node_results(opt_results, 'REGION1_storage_phs')
    print(phs['sequences'].max())
    print(phs['scalars'])

    # example plot for sequences
    cols = [c for c in phs['sequences'].columns if 'my_sequence_var' not in c]
    phs['sequences'] = phs['sequences'][cols]
    phs['sequences'].columns = ['P-IN', 'CAP', 'P-OUT']
    ax = phs['sequences'].plot(kind='line', drawstyle='steps-post')
    ax.set_title('Dispatch results')
    ax.set_xlabel('Time')
    ax.set_ylabel('Power (MW) / Energy (MWh)')
    plt.show()

    # example plot for scalars
    idx = [i for i in phs['scalars'].index if 'my_scalar_var' not in i]
    phs['scalars'] = phs['scalars'][idx]
    phs['scalars'].index = ['P-IN', 'CAP', 'P-OUT']
    ax = phs['scalars'].plot(kind='bar')
    ax.set_title('Investment results')
    ax.set_xlabel('')
    ax.set_ylabel('Storage investment in MWh / MW')
    plt.show()

@uvchik , @simnh : Are you fine with the naming (results, node_results) and the functionality?

We could then already adapt all existing examples and completely remove the old result-dict from this branch!

ckaldemeyer:
> Aren't the first custom components the storage? We should move variable fraction transformer also, or not? Or should be just give it a depreciated warning, as we will have the new chp components?

I would move them.

> I guess mollenhauer has no pure linear formulation, which would be good to have for large economic dispatch models as well, I would vote for extraction turbine then by the way..

We should keep both, shouldn't we?

> I think the custom sturcture is fine for now, we will see how it works with more complex and new components. does the storage example work BTW? Where are the custom_groupings added, should that not happen in the example somewhere? (I kind of lost track here)

It works. See csv-invest example where I have added some scalar and sequence variables to the blocks of the custom investment storage.

The grouping is added at the bottom of the custom module: https://github.com/oemof/oemof/blob/features/restructure-outputlib/oemof/solph/custom.py

> And we need to make sure, that if statements of 'standard' grouping and 'custom' grouping are compatiable (i.e. components are grouped correctly and not mixed up)

Maybe we should talk on this. But couldn't we just go with standard grouping per default?

ckaldemeyer:
> I think the custom sturcture is fine for now, we will see how it works with more complex and new components. does the storage example work BTW? Where are the custom_groupings added, should that not happen in the example somewhere? (I kind of lost track here)
> It works. See csv-invest example where I have added some scalar and sequence variables to the blocks of the custom investment storage. Both are integrated into the new structure automatically.

Actually, I have just looked at it and I am wondering the the block `GenericInvestmentStorageBlock`is used instead of the regular investment storage. I thought that I have changed the class within `nodes_flows.csv` but it was still `nodes_flows.csv`. Do you know why?

ckaldemeyer:
Anyhow, decision variables are collected for any subclass of `Node`.

ckaldemeyer:
> I think the custom sturcture is fine for now, we will see how it works with more complex and new components. does the storage example work BTW? Where are the custom_groupings added, should that not happen in the example somewhere? (I kind of lost track here)
> It works. See csv-invest example where I have added some scalar and sequence variables to the blocks of the custom investment storage. Both are integrated into the new structure automatically.
> Actually, I have just looked at it and I am wondering why the block GenericInvestmentStorageBlockis used instead of the regular investment storage. I thought that I have changed the class within nodes_flows.csv but it was still nodes_flows.csv. Do you know why?

Now I remember! I have at first tested the new API in the storage invest example where it behaves correctly: https://github.com/oemof/oemof/pull/329/commits/343c3d1b0ab61e3aeb033088175eb45da8600b55

But I am wondering why the variables also appear for the basic storage (not created as `solph.custom.GenericStorage` in the csv reader investment example. Weird..

simnh:
Regarding the custom stuff, it sounds good! 

Ah  ok now I see: The custom grouping is automatically added to the energy system in the solph/network.py , thats cool!

simnh:
actually I am thinking to also move the Flow to the custom stuff..and strip some arguments seldomly used in the 'standard' form in network.py....I guess if we move a lot of stuff to custom, we do not need to have to always think. 'ah, but that not really generic!', we just do not have to care anymore :)...but its just an idea

ckaldemeyer:
> Regarding the custom stuff, it sounds good!
> 
> Ah ok now I see: The custom grouping is automatically added to the energy system in the solph/network.py , thats cool!

Yes. I think that was already your code ;-)

> But I am wondering why the variables also appear for the basic storage (not created as solph.custom.GenericStorage) in the csv reader investment example. Weird..

Do you have an idea here?

ckaldemeyer:
@simnh : are you fine with the naming `from oemof.outputlib import results, node_results`?

ckaldemeyer:
> But I am wondering why the variables also appear for the basic storage (not created as solph.custom.GenericStorage) in the csv reader investment example. Weird..
> Do you have an idea here?

There is only a regular storage object in this example and nevertheless variables created in `solph.custom.GenericStorage` are collected within the result object.

simnh:
> @simnh : are you fine with the naming `from oemof.outputlib import results, node_results`?

results would be the dataframe and node_results one view? I think it would be greate to have 

`import views`
`views.node['node_label']`

Something like that? Because 'view' or `results_views` makes it much clearer that its basically just one view / slice of the whole thing...and you can import all views with one statement...

but I am also fine to go with the above...

ckaldemeyer:
Ah, I think because it's created within the base storage set:

> @simnh : are you fine with the naming from oemof.outputlib import results, node_results?
> results would be the dataframe and node_results one view? I think it would be greate to have

Yes.



> results would be the dataframe and node_results one view? I think it would be greate to have
> 
> import views
> views.node['node_label']
> 
> Something like that? Because 'view' or results_views makes it much clearer that its basically just one view / slice of the whole thing...and you can import all views with one statement...

Nice suggestion! I like it, too and will change it!

ckaldemeyer:
> results would be the dataframe and node_results one view? I think it would be greate to have
> 
> import views
> views.node['node_label']
> 
> Something like that? Because 'view' or results_views makes it much clearer that its basically just one view / slice of the whole thing...and you can import all views with one statement...
> Nice suggestion! I like it, too and will change it!

Here's the new API:

    from oemof.outputlib import results, views

    # generic result object
    opt_results = results.results(es, om)

    # view for a node
    custom_storage = views.node(opt_results, 'storage')
    custom_storage['sequences'].plot(kind='line', drawstyle='steps-post')
    plt.show()

In the first place, I thought it would be cool to have `results.dict()` or  `results.dictionary()` but as these are keywords and names should not represent datatypes its o.k. like this, I guess. I am quite happy with it!

ckaldemeyer:
@simnh , @uvchik : First read the last e-mail for all information!

New API after @uvchik s feedback:

    from oemof.outputlib import results, views    

    results = results.create_results(energysystem, om)

    custom_storage = views.get_node(opt_results, 'storage')
    custom_storage['sequences'].plot(kind='line', drawstyle='steps-post')
    plt.show()

This way, the functions can also be used standalone e.g. `get_node(opt_results, 'label')`.

Is everybody happy now? We can then continue..

uvchik:
Thank you very much for your work. The naming is always a matter of taste but I am fine with it.

The API seems to be quite clear. I would adapt the investment example to test the new API. I would start after @simnh and @gnn agreed with the API.

simnh:
You can leave it like that if you want, however I like to mention the following: 

'create_results' sounds like 'compute' which suggest that the model is solved, so I would like

`create_resutlts_df` or `get_results` or anything else which is less misleading...

actually then old api `results.results(es,om)` sounds better to me...but I do not want to discuss anymore...

uvchik:
> 'create_results' sounds like 'compute' which suggest that the model is solved

Sounds comprehensible. What about `fetch_results`?

ckaldemeyer:
> You can leave it like that if you want, however I like to mention the following:
> 
> 'create_results' sounds like 'compute' which suggest that the model is solved, so I would like
> 
> create_resutlts_df or get_results or anything else which is less misleading...
> 
> actually then old api results.results(es,om) sounds better to me...but I do not want to discuss anymore...

I think it is important to discuss because it will be fixed for the next couple of versions and will be changed everywhere.

> 'create_results' sounds like 'compute' which suggest that the model is solved
> Sounds comprehensible. What about fetch_results?

I prefer `get_results` and think it's a good compromise. Decision made.. :-)

ckaldemeyer:
https://github.com/oemof/oemof/issues/252

ckaldemeyer:
      File "investment.py", line 129, in <module>
        run_investment_example()
      File "investment.py", line 75, in run_investment_example
        results = results.get_results(es, om)
    UnboundLocalError: local variable 'results' referenced before assignment

Ahhhhhh, I would propose to just leave it like this and go with `result = results.get_results(es, om)` or change it to `results = data.get_results(es, om)`. But `data` is misleading because `views` also deals with data. I hate this naming discussion...

ckaldemeyer:
I have changed the module name to `processing`. It is quite generic but can hold everything arround data processing. Additionally, we do not run in naming conflicts with other objects like `results`.

New API:

    from oemof.outputlib import processing, views    

    # all results
    results = processing.get_results(energysystem, om)

    # node results
    custom_storage = views.get_node(opt_results, 'storage')

    # plot
    custom_storage['sequences'].plot(kind='line', drawstyle='steps-post')
    plt.show()

I hope this is o.k. for you @uvchik . @simnh is fine with it.

ckaldemeyer:
@gnn did not like the "get_" so we ended in:

New API:

    from oemof.outputlib import processing, views    

    # all results
    results = processing.results(energysystem, om)

    # node results
    custom_storage = views.node(results, 'storage')

    # plot
    custom_storage['sequences'].plot(kind='line', drawstyle='steps-post')
    plt.show()

I think this is a good compromise!

uvchik:
> @gnn did not like the "get_" 

:smile:

I'm fine with it and I will start to adapt the "storage_investment" example as soon as possible.

ckaldemeyer:
> @gnn did not like the "get_"
> 
> 
> I'm fine with it and I will start to adapt the "storage_investment" example as soon as possible.

Great! Finally, we made it :D

oemof/solph/network.py
```diff
@@ -129,18 +131,18 @@ def __init__(self, **kwargs):
...
-        self.actual_value = sequence(kwargs.get('actual_value'))
-        self.positive_gradient = sequence(kwargs.get('positive_gradient'))
-        self.negative_gradient = sequence(kwargs.get('negative_gradient'))
-        self.variable_costs = sequence(kwargs.get('variable_costs'))
-        self.fixed_costs = kwargs.get('fixed_costs')
-        self.summed_max = kwargs.get('summed_max')
-        self.summed_min = kwargs.get('summed_min')
-        self.fixed = kwargs.get('fixed', False)
-        self.investment = kwargs.get('investment')
+        scalars = ['nominal_value', 'actual_value', 'fixed_costs',
```
simnh: actual value is a sequence

ckaldemeyer: I fixed it.

ckaldemeyer:
@gnn : it works as expected. Thanks!

gnn:
@ckaldemeyer: I guess you already saw it, but if you merge [features/modifyable-inputs-and-outputs](https://github.com/oemof/oemof/tree/features/modifyable-inputs-and-outputs), you are able to modify the `inputs` and `outputs` dictionaries after construction.

ckaldemeyer:
> @ckaldemeyer: I guess you already saw it, but if you merge features/modifyable-inputs-and-outputs, you are able to modify the inputs and outputs dictionaries after construction.
> Thanks. I'll check it again and include it later.
> 
> @gnn @simnh @uvchik : Could you have a look at 68f5f02 (generic_chp.py, LP-file and custom.py) meanwhile?
> 
> For some reason, the outflows are not included in the bus balance within the LP-file even if the decision variables are existent and included within the objective:
> 
> \* Source Pyomo model name=OperationalModel *\
> 
> min 
> objective:
> -24 flow(bel_demand_el_0)
> -22.23 flow(bel_demand_el_1)
> +1 flow(pp_generic_chp_bth_0)
> +1 flow(pp_generic_chp_bth_1)
> +1000 flow(source_th_bth_0)
> +1000 flow(source_th_bth_1)
> 
> s.t.
> 
> c_e_Bus_balance(bel_0)_:
> +1 flow(bel_demand_el_0)
> = 0
> 
> c_e_Bus_balance(bel_1)_:
> +1 flow(bel_demand_el_1)
> = 0
> 
> c_e_Bus_balance(bgas_0)_:
> -1 flow(bgas_pp_generic_chp_0)
> +1 flow(rgas_bgas_0)
> = 0
> 
> c_e_Bus_balance(bgas_1)_:
> -1 flow(bgas_pp_generic_chp_1)
> +1 flow(rgas_bgas_1)
> = 0
> 
> c_e_Bus_balance(bth_0)_:
> +1 flow(source_th_bth_0)
> = 56.861198740000006
> 
> c_e_Bus_balance(bth_1)_:
> +1 flow(source_th_bth_1)
> = 56.279104700000005
> 
> c_e_ONE_VAR_CONSTANT: 
> ONE_VAR_CONSTANT = 1.0
> 
> bounds
>    0 <= flow(bel_demand_el_0) <= +inf
>    0 <= flow(bel_demand_el_1) <= +inf
>    0 <= flow(bgas_pp_generic_chp_0) <= +inf
>    0 <= flow(bgas_pp_generic_chp_1) <= +inf
>    0 <= flow(pp_generic_chp_bth_0) <= +inf
>    0 <= flow(pp_generic_chp_bth_1) <= +inf
>    0 <= flow(rgas_bgas_0) <= +inf
>    0 <= flow(rgas_bgas_1) <= +inf
>    0 <= flow(source_th_bth_0) <= +inf
>    0 <= flow(source_th_bth_1) <= +inf
> end
> At first, I thought the error is due to some constraints I implemented. But I uncommented everything in the block and it still occurs. So it seems to be due to the subsequent setting of inputs/outputs!? But everything looks fine until the om is created later..
> 
> At the moment, I don't see the wood anymore due to the trees... ;-)
> 
> Any hints?

I already appears with the `LinearTransformer` if outputs and inputs are updated subsequently. Only output flows are not included within the balances but input flows are.

@gnn: see https://github.com/oemof/oemof/pull/339 for a minimal example. It might be easier to solve it there..

gnn:
Sorry for posting a potentially dumb question, but I have to ask it before I have a look at this: since every output flow of one node is also the input flow of a different node, isn't the case you state to be expected? I mean, otherwise one would consider duplicate flows which is not what we want, I guess?
Unless you mean, that flows which are set modifying `outputs` are not considered at all. In that case I guess why this is a bug.
Now here are two suggestions: could you try generating the same lp-file using flows which are set on construction? Because then we (I) can check the diff of the lp-files and see what's missing. If that's too much work, I can do it. Eventually. ;)
Secondly: since this is the second time where you figure out that something doesn't work, maybe you can find out how to transform your minimal example into a failing test case? Cause then I could try to make this failing test case pass, which makes it easier for me to figure out when I'm successful in fixing the bug.
Just some friendly suggestions. I know you're already doing a lot. ;)

ckaldemeyer:
> Sorry for posting a potentially dumb question, but I have to ask it before I have a look at this: since every output flow of one node is also the input flow of a different node, isn't the case you state to be expected? I mean, otherwise one would consider duplicate flows which is not what we want, I guess?
> Unless you mean, that flows which are set modifying outputs are not considered at all. In that case I guess why this is a bug.

This is exactly what I mean. But they are considered only in the objective but not in the balance.

> Now here are two suggestions: could you try generating the same lp-file using flows which are set on construction? Because then we (I) can check the diff of the lp-files and see what's missing. If that's too much work, I can do it. Eventually. ;)

I have chosen an easier solution by just creating two linear transformers :-) See example, for the on-construction-set outputs, all flows are included within the balances!

> Secondly: since this is the second time where you figure out that something doesn't work, maybe you can find out how to transform your minimal example into a failing test case? Cause then I could try to make this failing test case pass, which makes it easier for me to figure out when I'm successful in fixing the bug.
> Just some friendly suggestions. I know you're already doing a lot. ;)

Shouldn't we do this if we (you :) have fixed the bug? I was thinking about turning all existing examples into modularized development examples (shifting the detailed ones into a separate repo) and writing tests for each module.

Let's continue with this issue here: https://github.com/oemof/oemof/pull/339

Thanks already :)

ckaldemeyer:
@uvchik : I have now adapted the simple dispatch example in https://github.com/oemof/oemof/pull/329/commits/2c032ca058d0fda1677791092c9804a1591fd2ec and tests are running successfully except for import errors due to the custom storage which has been ported and the investment object (I do not understand why but haven't checked yet).

@uvchik : Have a look if the structure is o.k. for you. I have now ported everything into one function, removed all logging, etc. and returned a result dict generated from the new outputlib.

oemof/outputlib/processing.py
```diff
@@ -0,0 +1,135 @@
...
+
+    # create final result dictionary by splitting up the dataframes in the
+    # dataframe dict into a series for scalar data and dataframe for sequences
+    results = {}
+    for k, v in df_dict.items():
+        df_dict[k].set_index('timestep', inplace=True)
+        df_dict[k] = df_dict[k].pivot(columns='variable_name', values='value')
+        df_dict[k].index = es.timeindex
+        try:
+            condition = df_dict[k].isnull().any()
```
uvchik: in commit 4fefc4bb71f @simnh added a possibility to check the solver output.

Furthermore [chapter 18.10 of the pyomo documentation](https://software.sandia.gov/downloads/pub/pyomo/PyomoOnlineDocs.html#_accessing_solver_status) shows a way to check the status.

```python
from pyomo.opt import SolverStatus, TerminationCondition
if es.results.solver.Solver.Status == SolverStatus.ok) and (es.results.solver.termination_condition == TerminationCondition.optimal):
     logging.error("The solver terminated in an non-optimal state. You cannot process any results)
```

Afterwards you could keep your check but change the message to "Something odd happend..." because you know that it is not caused by the solver because you already checked this case before.

oemof/outputlib/processing.py
```diff
@@ -0,0 +1,135 @@
...
+
+    # order the data by oemof tuple and timestep
+    df = df.sort_values(['oemof_tuple', 'timestep'], ascending=[True, True])
+
+    return df
+
+
+def results(es, om):
+    """
+    Create a result dictionary from the result DataFrame.
```
uvchik: Why do you pass the EnergySystem instance? As I understand it the  EnergySystem instance is became an attribute of the OperationalModel.

The following code should work if you need a shortcut. 
```python
def results(om):
    es = om.es
```

Or is it planned to remove the EnergySystem from the OperationalModel?

uvchik:
I fixed some tests but there is still a problem with the constraint test of the storage.

@ckaldemeyer Do you have an idea?

uvchik:
Finally I found the bug. I don't know why but the EnergySystem in the constraint test was taken
from the network package and not from solph package. I think it makes more sense this way.

Afterwards I had to change the lp-files because the name of the class changed from `Storage` to `GenericStorageBlock` etc.. I hope I did not oversee anything but this was the only difference in the lp files after I changed the EnergySystem.

Now all test are fine. The output of the `oemof_full_check`:
```
Ran 42 tests in 0.397s

OK
All nosetests passed!
All example tests passed!
```

😄

ckaldemeyer:
> Finally I found the bug. I don't know why but the EnergySystem in the constraint test was taken
> from the network package and not from solph package. I think it makes more sense this way.
> 
> Afterwards I had to change the lp-files because the name of the class changed from Storage to GenericStorageBlock etc.. I hope I did not oversee anything but this was the only difference in the lp files after I changed the EnergySystem.
> 
> Now all test are fine. The output of the oemof_full_check:
> 
> Ran 42 tests in 0.397s
> 
> OK
> All nosetests passed!
> All example tests passed!

Nice! Thanks a lot for your effort!

ckaldemeyer:
> Why do you pass the EnergySystem instance? As I understand it the EnergySystem instance is became an attribute of the OperationalModel.
> 
> The following code should work if you need a shortcut.
> 
> def results(om):
>     es = om.es
> Or is it planned to remove the EnergySystem from the OperationalModel?

Thanks for the hint. I didn't know that the es was integrated into the om and the only reason to pass it was the datetimeindex which also annoyed me. I removed it with https://github.com/oemof/oemof/pull/329/commits/79b554f3506a571cedd6ba52129167c4b335fcb2

ckaldemeyer:
Now only the variablefractiontransformer example is left and the old result-dict and outputlib can be removed completely!

uvchik:
> Now only the variablefractiontransformer example is left and the old result-dict and outputlib can be removed completely!

Go ahead with removing everything. And if everything but this examples works fine I will fix the example tonight.

ckaldemeyer:
> Now only the variablefractiontransformer example is left and the old result-dict and outputlib can be removed completely!
> Go ahead with removing the old stuff. And if everything but this examples works fine I will fix the example tonight.

I have now removed the old result dict and outputlib.

Two errors have shown up.

1.) Error within constraint tests -> the test must be omitted or re-written for the next outputlib:

	  File "/home/cord/Programmierung/oemof/tests/regression_tests.py", line 8, in <module>
	    from oemof.outputlib import ResultsDataFrame as RDF
	ImportError: cannot import name 'ResultsDataFrame'

2.) Error within example tests -> I don't see where the objective is checked @uvchik: do you have a quick idea!?:

	15:18:02-INFO-*********************************************
	15:18:02-INFO-Variable CHP example
	15:18:02-INFO-Used solver: cbc
	15:18:02-INFO-Run check: Failed
	15:18:02-INFO-Result check: Failed
	15:18:02-ERROR-'dict' object has no attribute 'objective'
	15:18:02-WARNING-Optimization with unknown termination condition. Writing output anyway...
	15:18:02-INFO-Optimization successful...

We are getting closer..

ckaldemeyer:
> 2.) Error within example tests -> I don't see where the objective is checked @uvchik: do you have a quick idea!?:
> 
> 15:18:02-INFO-*********************************************
> 15:18:02-INFO-Variable CHP example
> 15:18:02-INFO-Used solver: cbc
> 15:18:02-INFO-Run check: Failed
> 15:18:02-INFO-Result check: Failed
> 15:18:02-ERROR-'dict' object has no attribute 'objective'
> 15:18:02-WARNING-Optimization with unknown termination condition. Writing output anyway...
> 15:18:02-INFO-Optimization successful...

I found it in `models.py` in the methods `solve()` where results are appended to `es.results`. I'll fix it!

simnh:
> Or  is it planned to remove the EnergySystem from the OperationalModel?

No, though @gnn and I where thinkig, that it might make sense to hand the model to the energy system

I guess we can discuss this at the webco, if and how we want to 'improve' the workflow and the handling as described in #314

ckaldemeyer:
> 2.) Error within example tests -> I don't see where the objective is checked @uvchik: do you have a quick idea!?:
> 
> 15:18:02-INFO-*********************************************
> 15:18:02-INFO-Variable CHP example
> 15:18:02-INFO-Used solver: cbc
> 15:18:02-INFO-Run check: Failed
> 15:18:02-INFO-Result check: Failed
> 15:18:02-ERROR-'dict' object has no attribute 'objective'
> 15:18:02-WARNING-Optimization with unknown termination condition. Writing output anyway...
> 15:18:02-INFO-Optimization successful...
> I found it in models.py in the methods solve() where results are appended to es.results. I'll fix it!

Fixed it with: https://github.com/oemof/oemof/pull/329/commits/8be7260bbd805cbde9dde83c5a7e8b184735d798

uvchik:
With commit 00e898f8c I added the meta results function according to #340. For the storage_investment example the output of pprint looks like this:

```python
{'objective': 1.2682652286866047e+19,
 'problem': {'Lower bound': 1.2682652e+19,
             'Name': 'tmped7eimr4.pyomo',
             'Number of constraints': 9413,
             'Number of nonzeros': 25542,
             'Number of objectives': 1,
             'Number of variables': 9412,
             'Sense': EnumValue(<pyutilib.enum.enum.Enum object at 0x7fd43add6be0>, 1, 'minimize'),
             'Upper bound': 1.2682652e+19},
 'solver': {'Error rc': 0,
            'Status': EnumValue(<pyutilib.enum.enum.Enum object at 0x7fd43ae3d320>, 0, 'ok'),
            'Termination condition': EnumValue(<pyutilib.enum.enum.Enum object at 0x7fd43ae3d358>, 8, 'optimal'),
            'Time': 0.44408273696899414,
            'User time': -1.0}}
```

Feel free to comment this approach or to add any meta results you are missing directly to the function.

uvchik:
@ckaldemeyer Please be more carefully with the tests. I had to fix the same tests again.

@gnn There is a `Todo` in the regression test since v0.1. Therefore the regression test is disabled. Do you think we should adapt it or remove it?

ckaldemeyer:
> With commit 00e898f I added the meta results function according to #340. For the storage_investment example the output of pprint looks like this:
> 
> {'objective': 1.2682652286866047e+19,
>  'problem': {'Lower bound': 1.2682652e+19,
>              'Name': 'tmped7eimr4.pyomo',
>              'Number of constraints': 9413,
>              'Number of nonzeros': 25542,
>              'Number of objectives': 1,
>              'Number of variables': 9412,
>              'Sense': EnumValue(<pyutilib.enum.enum.Enum object at 0x7fd43add6be0>, 1, 'minimize'),
>              'Upper bound': 1.2682652e+19},
>  'solver': {'Error rc': 0,
>             'Status': EnumValue(<pyutilib.enum.enum.Enum object at 0x7fd43ae3d320>, 0, 'ok'),
>             'Termination condition': EnumValue(<pyutilib.enum.enum.Enum object at 0x7fd43ae3d358>, 8, 'optimal'),
>             'Time': 0.44408273696899414,
>             'User time': -1.0}}
> Feel free to comment this approach or to add any meta results you are missing directly to the function.

I am fine with it. Thanks for your effort!

ckaldemeyer:
> @ckaldemeyer Please be more carefully with the tests. I had to fix the same tests again.
> 
> @gnn There is a Todo in the regression test since v0.1. Therefore the regression test is disabled. Do you think we should adapt it or remove it?

Thanks. Did I miss something?

ckaldemeyer:
I have added:

> How to import custom components? Add them to solph/__init__.py or always do something like from oemof.solph.custom import MyComp?
> 
>  Should we remove the standard Storage and only keep the GenericStorage?

We could take the TODOs from this PR and the ones from the roadmap issue as agenda for the WebCo!

I have also moved the VariableFractionTransformer to the custom module with
https://github.com/oemof/oemof/pull/329/commits/c335f1ce9bf672367ae3d2256aec037c57b83801 but now the constraint test fails despite the VFT-example is running for itself. Does someone has an idea why!?

ckaldemeyer:
> I have also moved the VariableFractionTransformer to the custom module with
> c335f1c but now the constraint test fails despite the VFT-example is running for itself. Does someone has an idea why!?

I had subclassed from `network.Transformer` instead of `solph.LinearTransformer` and fixed it in: https://github.com/oemof/oemof/pull/329/commits/113594769f3520b3484038e2f3a31add5275cf32

Now its working (running with flows > 0) in the example. But still the constraint test fails. @uvchik Do you have an idea why?

uvchik:
> Now its working (running with flows > 0) in the example. But still the constraint test fails. @uvchik Do you have an idea why?

The test fails because the component is not working right. I will have a look.

simnh:
So I removed the storage and renamed custom to components...the checks pass, except the variable fraction transformer. @ckaldemeyer and I would like to merge soon, to enable new branches from dev with different features...
could you have a look @uvchik ? Or should we maybe merge and fix the component VFT on another branch afterwards?

uvchik:
I found out that accidentally the constraints for the LinearTransformer are build because it is a child class form the LinearTransformer. Changing the groupings in groupings.py solves one Problem.

```diff
+    if type(node) == LinearTransformer:
+        return blocks.LinearTransformer
-    if isinstance(node, LinearTransformer):
-        return blocks.LinearTransformer
```

On the upside you see that the constraints for the LinearTransformer are no longer build on the downside you also see that the constraints of the VariableFranctionTransformer are still not build.

@simnh , @ckaldemeyer , @gnn Any idea, why the grouping does not work?

simnh:
> @simnh , @ckaldemeyer , @gnn Any idea, why the grouping does not work?

I think i fixed it, you still need to change the isinstance conditional to type as above. I did not do this yet, but the variable faction transformer group is now present, and constraints should be build..

simnh:
@uvchik  can you merge this one when you fixed the VFT problem? We can than start new, smaller feature branches :)

ckaldemeyer:
> uvchik deleted the features/restructure-outputlib  branch 2 minutes ago

Yeeehaaa! Thanks :smile:

ckaldemeyer:
> git checkout dev!

uvchik:
Done 🎉  Many, many thanks to @ckaldemeyer to the cool results API!

@simnh thank you for your fix 😄

ckaldemeyer:
> Done  Many, many thanks to @ckaldemeyer to the cool results API!
> 
> @simnh thank you for your fix

Also thanks to you and @gnn for all the help!

By the way I just did the first hotfix (breaking graph tools test) on the dev!

PR #331, opened 2017-08-01
Polish VariableFractionTransformer

ckaldemeyer:
Some real minor stuff I came across when I checked if this component could be used within our project..


simnh:
This will happen with restructure outputlib branch

PR #333, opened 2017-08-24
Integrate VariableFractionTransformer into CSV-Reader

ckaldemeyer:


ckaldemeyer:
It should work now although the CSV-Reader is really hacky. Yesterday I have thought about alternatives but I think the overall structure is not bad. Maybe someday we chop it into smaller units as @uvchik has already suggested..

simnh:
what about this one?

ckaldemeyer:
> what about this one?

It whould principally work but the imports have to be adapted since VFT is now in the components module.

@uvchik @simnh : Will it remain there? Then, I could easily adjust and merge it.

simnh:
From my perspective we could be more specific with the naming, and transfer the VFT to a simple Extraction Turbine as it is inside the components module, where we can be a little less mystic with the names?

ckaldemeyer:
> yes I think so, but from my perspective we could be more specific with the naming, and transfer the VFT to a simple Extraction Turbine as it is inside the components module, where we can be a little less mystic with the names?

:+1:

ckaldemeyer:
Obsolete as the csv-reader will be removed: https://github.com/oemof/oemof/issues/355

PR #338, opened 2017-10-11
changing simple_dispatch with simple_dispatch

wolfbunke:
The simple_least_costs example is not executable in the last oemof version, so I suggest to change it to  simple_dispatch or similar.


ckaldemeyer:
Thanks. Have you tested if it works?

wolfbunke:
Yes, it works :-)
works

ckaldemeyer:
> Yes, it works :-)

Thanks!

PR #339, opened 2017-10-11
Modifyable inputs and outputs not working properly

ckaldemeyer:
@gnn: See example which should illustrate the unexpected behaviour!


ckaldemeyer:
I thought it must be somewhere in the bus balance:

        I = {}
        O = {}
        for n in group:
            I[n] = [i for i in n.inputs]
            O[n] = [o for o in n.outputs]

        def _busbalance_rule(block):
            for t in m.TIMESTEPS:
                for n in group:
                    lhs = sum(m.flow[i, n, t] * m.timeincrement[t]
                              for i in I[n])
                    rhs = sum(m.flow[n, o, t] * m.timeincrement[t]
                              for o in O[n])
                    expr = (lhs == rhs)
                    # no inflows no outflows yield: 0 == 0 which is True
                    if expr is not True:
                        block.balance.add((n, t), expr)
        self.balance = Constraint(group, noruleinit=True)
        self.balance_build = BuildAction(rule=_busbalance_rule)

But to me it looks o.k...

gnn:
> Shouldn't we do this if we (you :) have fixed the bug? I was thinking about turning all existing examples into modularized development examples (shifting the detailed ones into a separate repo) and writing tests for each module.
Nope. If you find a bug, write the failing test first, check that it actually fails and then write the code that makes it pass. Otherwise, you can't really be sure that your test actually tests the behaviour you want to have tested the right way. ;)

oemof/solph/blocks.py
```diff
@@ -824,25 +824,25 @@ def _create(self, group=None):
...
-                            lhs = m.flow[I[n], n, t] * \
-                                  n.conversion_factors[o][t]
-                            rhs = m.flow[n, o, t]
-                        except:
-                            raise ValueError("Error in constraint creation",
-                                             "source: {0}, target: {1}".format(
-                                                 n.label, o.label))
-                        block.relation.add((n, o, t), (lhs == rhs))
-        self.relation_build = BuildAction(rule=_input_output_relation)
+        # I = {n: [i for i in n.inputs][0] for n in group}
```
gnn: @ckaldemeyer You explain, why you loop for `n.inputs` here but over `n.outputs.keys()` below? Also, why are you accessing the first input flow here, but all output flows below?

ckaldemeyer: > @ckaldemeyer could you explain why you loop over n.inputs here but over n.outputs.keys() below?

This is the standard `LinearTransformer` and I haven't written the code. Looping over `n.inputs` should result in the same as looping over `n.inputs.keys()`. I found it just more explicit to write the keys..

> Also, why are you accessing the first input flow here, but all output flows below?

Because  a `LinearTransformer` always has one input and can have multiple outflows.

ckaldemeyer:
I have traced the error a bit further with: https://github.com/oemof/oemof/pull/339/commits/ad4d54152d046b50a4828f1a96d8ab5afb8b3ca4

The problem is that node outputs that are added subsequently and not registered as bus inputs. This must be dealed with somewhere in the network which I'll trace now..

    ### inputs/outputs after updating outputs manually
    Node bel
    Outputs:  {'demand_el': <oemof.solph.network.Flow object at 0x7f803b70a5f8>}
    Inputs:  {'chp_gas_working': <oemof.solph.network.Flow object at 0x7f803b70a9b0>}
    Node chp_gas_working
    Outputs:  {'bth': <oemof.solph.network.Flow object at 0x7f803b70ab38>, 'bel': <oemof.solph.network.Flow object at 0x7f803b70a9b0>}
    Inputs:  {'bgas': <oemof.solph.network.Flow object at 0x7f803b70a828>}
    Node chp_gas_not_working
    Outputs:  {'bth': <oemof.solph.network.Flow object at 0x7f803b71c160>, 'bel': <oemof.solph.network.Flow object at 0x7f803b70af60>}
    Inputs:  {}

gnn:
> Sorry, but I do not want to write an extra test now. Why can't you just run the example and look at the .lp-file?! Everything gets clear this way.

Yeah, no biggie. Didn't want you to write a test now. Just wanted to point this out for the next time s.th. like this happens.

> Because a LinearTransformer always has one input and can have multiple outflows.

That's not what confused me. What confuses me is, that this way the values in `I` are plain objects (probably `Node`s) while the values in `O` are lists. But since you didn't write this, I guess I have to ask whoever wrote it.

> The problem is that node outputs that are added subsequently and not registered as bus inputs. This must be dealed with somewhere in the network which I'll trace now..

Nice. Now I have a pretty clear idea what's going wrong. Fix inbound.

ckaldemeyer:
> The problem is that node outputs that are added subsequently and not registered as bus inputs. This must be dealed with somewhere in the network which I'll trace now..
> Nice. Now I have a pretty clear idea what's going wrong. Fix inbound.

Nice. I'll have a tea now ;-)

gnn:
Sorry for taking so long. Please check whether it works for you now.

ckaldemeyer:
> Sorry for taking so long. Please check whether it works for you now.

Thanks. No, is still does not work.

Adding outflows manually via:

    pp_not_working.outputs.update({bel: solph.Flow(nominal_value=100,
                                                   variable_costs=2),
                                   bth: solph.Flow(nominal_value=100,
                                                   variable_costs=2)})

still does not set these as bus inputs:

    ### inputs/outputs after updating outputs manually
    Node bel
    Outputs:  {'demand_el': <oemof.solph.network.Flow object at 0x7f9eb0b4f7f0>}
    Inputs:  {'chp_gas_working': <oemof.solph.network.Flow object at 0x7f9eb0b4fba8>}

Or did I miss anything!?

ckaldemeyer:
And if they are not set as bus inputs, they are not within the bus balance..

gnn:
Arrgh... wie viele Bugs kann man eigentlich im code haben?
Ich schau mal rein.

gnn:
Hrmpf. It should work the way you want it now, but some `LinearTransformer` tests are failing and I have no idea why. But first, check that it meets your expectations.

uvchik:
> Hrmpf. It should work the way you want it now, but some LinearTransformer tests are failing and I have no idea why.

The test failed because the building commands were missing. Just fixed it by removing the comment signs.

ckaldemeyer:
> Hrmpf. It should work the way you want it now, but some LinearTransformer tests are failing and I have no idea why. But first, check that it meets your expectations.

Works as expected and ported into the outputlib refactoring with https://github.com/oemof/oemof/pull/329/commits/a360a53f22c6cc7fa1738c539adfb5a8bc987170.

@gnn: Thanks a lot!

PR #342, opened 2017-10-18
Features/multi period

simnh:
* [ ] Adapt new results API from #329
* [ ] Check compatibility with csv-reader
* [ ] Provide example with multiple periods
* [ ] Check with custom components?


simnh:
close in favour of #353

PR #345, opened 2017-10-18
solph LOPF 

simnh:
Linear Optimal Powerflow for solph with custom componets

### DIFF
--- a/examples/solph/lopf/3_node_lopf_example.py
+++ b/examples/solph/lopf/3_node_lopf_example.py
@@ -1,0 +1,103 @@
+# -*- coding: utf-8 -*-
+"""
+"""
+# ############################### imports #####################################
+import pandas as pd
+import matplotlib.pyplot as plt
+import logging
+import os
+
+# solph imports
+from oemof.solph import (Sink, Source, LinearTransformer, ElectricalBus, Flow,
+                         OperationalModel, EnergySystem, Bus, ElectricalLine)
+import oemof.outputlib as output
+from oemof.tools import logger
+
+# ####################### data & initialization ###############################
+
+
+def initialise_energysystem(periods=5):
+    """
+    """
+    datetimeindex = pd.date_range('1/1/2012', periods=periods, freq='H')
+
+    return EnergySystem(timeindex=datetimeindex)
+
+
+# ######################### create energysystem components ####################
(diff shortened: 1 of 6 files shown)


ckaldemeyer:
I think this is a nice functionality but shouldn't all nw-classes and blocks be in the `components` or `custom` module.

As I understood it yesterday:

* `network` and `blocks` -> Generic components like Sink, Source, Transformer
* `components` -> Solph components like GenericStorage (+Investment), VariableFractionTransformer, GenericCHP (integrated in solph init)
* `custom` -> Custom stuff like CAES, ....

Meanwhile I think about if integrating `blocks` into `network` would be more consistent..

uvchik:
> Meanwhile I think about if integrating blocks into network would be more consistent..

What about moving the blocks from `blocks` to `network` and removing the `blocks` module afterwards.

simnh:
I wrote this some time ago, @ckaldemeyer you are right, as we now have this structure I just wanted to add this pull-request and maybe someday integrate this in the new structure...

simnh:
I do not know, it would be consistent with components.py, but I do not know if its nice to mix up solph classes with some internal pyomo constraint stuff (however we did it anyway already..so we could do this!) 

then I think we might think about the names of BinaryFlow etc as well..I'd prefer rather something like 'nonconvexities'...

ckaldemeyer:
> Meanwhile I think about if integrating blocks into network would be more consistent..
> What about moving the blocks from blocks to network and removing the blocks module afterwards.

That's exactly what I meant :smile: 

> I do not know, it would be consistent with components.py, but I do not know if its nice to mix up solph classes with some internal pyomo constraint stuff (however we did it anyway already..so we could do this!)

Why not put them in `custom` at first. What internal constraint stuff do you mean? In `custom` it would be fine I guess..

> then I think we might think about the names of BinaryFlow etc as well..I'd prefer rather something like 'nonconvexities'...

It all depends on how we build categories. Based on the impact on the problem type (nonconvexities, ...), on real world representations, on the specific/generic usability, ...

ckaldemeyer:
We just have to be carefull not to blow up the `custom` module up to 20.000 lines with stuff that no one needs and might not even be working anymore. 

Nonetheless, LOPF is an improvement!

simnh:
won't be able to do it for now...

PR #347, opened 2017-10-19
Additional constraints 

simnh:
Hi i added a draft. 

* [x] Find a name (I was bei übergreifend, because I think this is the main structural difference but the name is not very good)
* [ ] decide some general rules (do we want to allow for 'new' flow attributes, e.g. like I did with emission
      I think this would be important to enable highest flexibility

Continues #284


uvchik:
Cool, thank you for your proposal. 

a) What about just `constraints` as we do have `components`.

b) I think we should. Or we pass these values directly e.g. with a second dictionary. But writign the values into the flow might be more lucid.

I do not know if I understand the API. Does it work like this?

```python
energysystem = ...

bgas = solph.Bus(label="natural_gas")
bcoal = solph.Bus(label="hard_coal")

sb_gas = solph.Flow(nominal_value=1000, emission=5)
sb_coal = solph.Flow(nominal_value=1000, emission=8)

sgas = solph.Source(label='rgas', outputs={bgas: sb_gas})
scoal = solph.Source(label='rcoal', outputs={bcoal: sb_coal})

[....]
om = solph.OperationalModel(energysystem)

my_flows = {
    (s_gas, b_gas): sb_gas,
    (s_coal, b_coal): sb_coal,}

emission_limit(om, flows=my_flows, limit=500)

om.solve(solver=....)
```

ckaldemeyer:
> Cool, thank you for your proposal.

:+1: 

> a) What about just constraints as we do have components.

:+1: 




> b) I think we should. Or we pass these values directly e.g. with a second dictionary. But writign the values into the flow might be more lucid.
> 
> I do not know if I understand the API. Does it work like this?
> 
> energysystem = ...
> 
> bgas = solph.Bus(label="natural_gas")
> bcoal = solph.Bus(label="hard_coal")
> 
> sb_gas = solph.Flow(nominal_value=1000, emission=5)
> sb_coal = solph.Flow(nominal_value=1000, emission=8)
> 
> sgas = solph.Source(label='rgas', outputs={bgas: sb_gas})
> scoal = solph.Source(label='rcoal', outputs={bcoal: sb_coal})
> 
> [....]
> om = solph.OperationalModel(energysystem)
> 
> my_flows = {
>     (s_gas, b_gas): sb_gas,
>     (s_coal, b_coal): sb_coal,}
> 
> emission_limit(om, flows=my_flows, limit=500)
> 
> om.solve(solver=....)

I came across this with the `GenericCHP` because I did not want to blow up the class with any attribute for some fancy component. So now its possible to add arbitrary attributes to a Flow thanks to @gnn .
Also see my example for the usage: https://github.com/oemof/oemof/blob/dev/examples/solph/generic_chp/generic_chp.py

Concerning the Flow attributes I would rather stop to integrate them explicitly right after the step from the `components` to the `custom` module. Or already from `network` to `components`. Otherwise we end up having 4711 attributes in the Flow class.

simnh:
I was at the name `constraints` as well, but I think it might make sense to differentiate them from all other constraints? also referring to the paper, that constraints related to components and flows are only referring to its predecessor and successor and the node, and not anything else inside the network..

The API would exactly look like you described @uvchik

ckaldemeyer:
> I was at the name constraints as well, but I think it might make sense to differentiate them from all other constraints?

I think this is already done if someone imports the module and must know what he is doing. Names can also be provided in the functions from my point of view.

By the way as we are going towards more and more modules:

Would it make sense to also go with doctests like for the graph tools (https://github.com/oemof/oemof/blob/dev/oemof/outputlib/graph_tools.py)?

We could to this for the constraint, outputlib.processing module and even with the (development) examples. Of course this is more for the long run..

ckaldemeyer:
> The API would exactly look like you described @uvchik 

I quite like the API as it is now (and proposed by @uvchik).

uvchik:
> By the way as we are going towards more and more modules:

Is this a problem or not pythonic?

uvchik:
> Would it make sense to also go with doctests like for the graph tools (https://github.com/oemof/oemof/blob/dev/oemof/outputlib/graph_tools.py)?

I think doctests are good way to combine usage examples and we already have a few of them. Sometimes it is difficult to combine both so it is good to have normal tests, too. It would be nice to integrate the example tests into the "normal" test to avoid two different structures but we could address that after the release because it will not change the API.

uvchik:
> I came across this with the GenericCHP because I did not want to blow up the class with any attribute for some fancy component. So now its possible to add arbitrary attributes to a Flow thanks to @gnn .
> Also see my example for the usage: https://github.com/oemof/oemof/blob/dev/examples/solph/generic_chp/generic_chp.py
> 
> Concerning the Flow attributes I would rather stop to integrate them explicitly right after the step from the components to the custom module. Or already from network to components. Otherwise we end up having 4711 attributes in the Flow class.

So far we haven't decided what to do with additional attributes.

 * Using `**kwargs` as we do it now. On the downside we won't be able catch wrong attributes.
 * Using our own 'kwargs' e.g. additional_attributes (dict). In that case addtional attributes are seen as those but we have to explain it (oemof solution instead of python solution). We could catch wrong attributes.
 * Using optional attributes. It is more explicite but could lead to exactly (😏) 4711 attributes.
 * @ckaldemeyer approach, which I do not understand 😞, sorry.

ckaldemeyer:
> By the way as we are going towards more and more modules:
> Is this a problem or not pythonic?

This is pythonic and I like it.

> Would it make sense to also go with doctests like for the graph tools (https://github.com/oemof/oemof/blob/dev/oemof/outputlib/graph_tools.py)?
> I think doctests are good way to combine usage examples and we already have a few of them. Sometimes it is difficult to combine both so it is good to have normal tests, too. It would be nice to integrate the example tests into the "normal" test to avoid two different structures but we could address that after the release because it will not change the API.

:+1: 

> So far we haven't decided what to do with additional attributes.
> 
> Using **kwargs as we do it now. On the downside we won't be able catch wrong attributes.
> Using our own 'kwargs' e.g. additional_attributes (dict). In that case addtional attributes are seen as those but we have to explain it (oemof solution instead of python solution). We could catch wrong attributes.
> Using optional attributes. It is more explicite but could lead to exactly () 4711 attributes.
> @ckaldemeyer approach, which I do not understand , sorry.

It is just defining them in the class when they belong to components in the network (or components module) and passing them dynamically for custom components. Using this logic, I would have to integrate the flow attributes for `GenericCHP` within the Flow class.

simnh:
I think the  `kwargs` way like it is now would be fine for me...

uvchik:
I think we could simplify the API by looking for the attribute `emission`.

```python
emission_limit(om, limit=5000):
```
Inside the function we could create a set of all flows that have the attribute emission if `flows is None`.

Then you have both possibilities, passing the flow-set or creating the set by the emission attribute.

I'm fine with this approach and you can merge if you want, I just think that this will make thing much easier and filtering by the attribute shouldn't be too difficult.

simnh:
We can do it, but I like to keep it optionally, because it allows for flexibility, for example only work on a subset of flows that have emission an attribute.

uvchik:
> We can do it, but I like to keep it optionally, because it allows for flexibility, for example only work on a subset of flows that have emission an attribute.

I do agree. It is already an optional parameter so `if flow is None` would be easy.

uvchik:
Just added two tests.

@simnh Could you [check the lp-file](https://github.com/oemof/oemof/blob/features/individual-constraints/tests/lp_files/emission_limit.lp)? Looks fine to me. See code below.

```python
bel = solph.Bus(label='electricityBus')

solph.Source(label='source1', outputs={bel: solph.Flow(
    nominal_value=100, emission=0.5)})
solph.Source(label='source2', outputs={bel: solph.Flow(
    nominal_value=100, emission=0.8)})

om = self.get_om()

solph.constraints.emission_limit(om, om.flows, limit=777)
```

simnh:
lp file looks fine to me!

uvchik:
I added the feature from [above](https://github.com/oemof/oemof/pull/347#issuecomment-339648068).

All tests work fine now. @simnh agreed with the lp-file. For me we can merge.

uvchik:
🎉 @simnh Thank you!

PR #349, opened 2017-10-19
Improve GenericCHP component

ckaldemeyer:


ckaldemeyer:
Memos:

* Empty decision variables should be removed or dealt with in the new RDF!

ckaldemeyer:
@simnh @uvchik : I have provided examples for the usage `GenericCHP` within the solph examples. They are still very lean, not wrapped within a function and not in standard console program style.

Do you think they should be tested, too or can they just live without a test?

Feedback is welcome! Furthermore, this PR contains a fix for the outputlib that removes unused decision variables from custom components.

uvchik:
> Do you think they should be tested, too or can they just live without a test?

I would like to make it mandatory to provide a constraint test for every component that is added to the component or custom module. Example test are optional in my opinion. We could think of moving simple result tests to test module and remove the example tests at all. But this is not urgent.

ckaldemeyer:
> Do you think they should be tested, too or can they just live without a test?
> I would like to make it mandatory to provide a constraint test for every component that is added to the component or custom module.

:+1: 

> Example test are optional in my opinion. We could think of moving simple result tests to test module and remove the example tests at all. But this is not urgent.

As I said, I really like the idea of integrating the tests directly in the examples and not keeping them centrally!

uvchik:
> As I said, I really like the idea of integrating the tests directly in the examples and not keeping them centrally!

Ah, okay. You want to move the full examples into the docstrings. Why not. We could give it a try.

simnh:
I think results test are really great...

ckaldemeyer:
> As I said, I really like the idea of integrating the tests directly in the examples and not keeping them centrally!
> Ah, okay. You want to move the full examples into the docstrings. Why not? We could give it a try.

See: https://github.com/oemof/oemof/pull/349/commits/1f7ae2a766a9e54e5be2b46b796ef0b14a2adfbc 

I have written a simple doctest that checks the results and a simple assertion that checks the lp-files. You can make the checks fail if you change time series or set the number of timesteps <200 (200 is max. and leads to the same lp-files and thus results).

@uvchik , @simnh : What do you think of testing all examples this way and not centrally?

I think it would be pretty easy this way!

uvchik:
I am fine with replacing the result-tests with doctest but we have to find a way to execute them as they are outside the source folder. The following does not work.
```
nosetests --with-doctest -w /my/path/to/oemof
```

We could move the examples into the code folder but I do not know if this is pythonic. I once heard that we should do this anyway but I can't remember who said this and why (@gnn?).

I totally disagree with the constraint test.

 1. There are functions to make the tests more stable e.g. https://github.com/oemof/oemof/pull/357
 2. in the constraint test we reduced the number of time steps to 3 and that is enough to test the constraints

ckaldemeyer:
> I am fine with replacing the result-tests with doctest but we have to find a way to execute them as they are outside the source folder. The following does not work.
> 
> nosetests --with-doctest -w /my/path/to/oemof
> We could move the examples into the code folder but I do not know if this is pythonic. I once heard that we should do this anyway but I can't remember who said this and why (@gnn?).

What about temporarily creating a symbolic link for the examples (in the code folder), executing the test and removing the link afterwards?

Anyhow, there are probably different ways to resolve this but I am happy that you are fine with the result tests.

> I totally disagree with the constraint test.
> 
> There are functions to make the tests more stable e.g. #357

I am actually wondering if this was necessary. If a pyomo-version changes the alignment of the lp-files (which probably does not happen very often), one could check the examples by hand and just create one with the new style.

> in the constraint test we reduced the number of time steps to 3 and that is enough to test the constraints

This would also be possible in single examples by adapting the function arguments, wouldn't it?

Nevertheless, I could go with you if you want to keep the tests centrally. But having self-testing examples feels a bit cleaner, easier and just more "standard python" for me.

ckaldemeyer:
> I am fine with replacing the result-tests with doctest but we have to find a way to execute them as they are outside the source folder. The following does not work.
> 
> nosetests --with-doctest -w /my/path/to/oemof
> We could move the examples into the code folder but I do not know if this is pythonic. I once heard that we should do this anyway but I can't remember who said this and why (@gnn?).
> 
> I totally disagree with the constraint test.
> 
> There are functions to make the tests more stable e.g. #357
> in the constraint test we reduced the number of time steps to 3 and that is enough to test the constraints

It might also be an idea to just have one big (separated) central constraint test and examples with included result tests (doctests).

uvchik:
**Generally:** As we will have an example repository I think of making them test in the first place and moving them into a subfolder of the test-folder. Internally we know that you can have a look at the tests to learn how things work but they are primarily made for testing rather than for showing. Within the test folder you can make the doctest a unit test and that's all.

If you do not like this idea I could live with compromises:

 * I would rather move the example folder than creating a temporary link.

 * A third reason for the constraint tests are that they are reduced to the minimal number of constraints. You cannot even solve them like this but it makes the test quite specific. And, yes you can run them with different attributes and skip the solver (as it is not necessary) but it does not make it cleaner in my opinion.

ckaldemeyer:
> Generally: As we will have an example repository I think of making them test in the first place and moving them into a subfolder of the test-folder. Internally we know that you can have a look at the tests to learn how things work but they are primarily made for testing rather than for showing. Within the test folder you can make the doctest a unit test and that's all.

I could live with this.

> A third reason for the constraint tests are that they are reduced to the minimal number of constraints. You cannot even solve them like this but it makes the test quite specific. And, yes you can run them with different attributes and skip the solver (as it is not necessary) but it does not make it cleaner in my opinion.

To me it is cleaner as it requires less code and user-defined test routines you have to deal with e.g. in https://github.com/oemof/oemof/pull/357 you said that you don't understand the code but trust @gnn :smile: 

But as I said you can keep it central and as it is if you want. I am just wondering if it is really the better solution or not only for historical reasons. Maybe it's also just a matter of taste..

uvchik:
> I am just wondering if it is really the better solution or not only for historical reasons.

One main reason for me is that it is easier to reduce the problem. This makes it a lot easier to find the problem in case of failing tests. A test for a sink could looks like this at the moment, for me this is really lean.

```python
test_bus = solph.Bus(label='test')
demand_el = solph.Sink(label='test_sink', inputs={test_bus: solph.Flow(variable_costs=[3, 5, 9])})
compare_lp_files('sink.lp')
```

The second main reason is that we had some problems with pyomo release (not only once). Each time we had to adapt all tests. Adapting test means to be really sure that the reason for the fail is only the wrong notation and not something else. Because if you do it wrong you consolidate a bug. Function like `def chop_trailing_whitespace` and `def remove` are witnesses of these problems. Actually this could be integrated in your approach by using oemof's `compare_lp_files` instead of python's `filecmp.cmp`.

The proposal to move the examples into the test folder, remove the example test and write unittest or doctest instead to test the results in future should show that I do not hang on old structures due to historical reasons. The example test was my work and with this proposal I would bury them.

The extract of our discussion could be to go to very pure combined result- and constraint tests. One test-block for each component. Each test-block contains of just one component plus additional sinks/sources to make it feasible.

ckaldemeyer:
> One main reason for me is that it is easier to reduce the problem. This makes it a lot easier to find the problem in case of failing tests. A test for a sink could looks like this at the moment, for me this is really lean.
> 
> test_bus = solph.Bus(label='test')
> demand_el = solph.Sink(label='test_sink', inputs={test_bus: solph.Flow(variable_costs=[3, 5, 9])})
> compare_lp_files('sink.lp')

This could also be realized within one (central) example with different components that creates one lp file per component and checks an original against the created one using `filecmp` like in https://github.com/oemof/oemof/pull/349/files#diff-4a4a36616f489dd28746ffb70fad735d. It does not require and user-defined logic and is also very lean.

> The second main reason is that we had some problems with pyomo release (not only once). Each time we had to adapt all tests. Adapting test means to be really sure that the reason for the fail is only the wrong notation and not something else. Because if you do it wrong you consolidate a bug. Function like def chop_trailing_whitespace and def remove are witnesses of these problems. Actually this could be integrated in your approach by using oemof's compare_lp_files instead of python's filecmp.cmp.

I guess if we had one central example that tests all components (like mentioned above) and different result tests (realized as doctests or otherwise) we would notice in both ways if pyomo changes its API and how it changed the API. Normally, all result tests should work then and we could (even explicitly if we use doctests) see the expected and returned results i. e. the difference.

So I am still not convinced that the current way provides advantages ;-)

> The proposal to move the examples into the test folder, remove the example test and write unittest or doctest instead to test the results in future should show that I do not hang on old structures due to historical reasons. The example test was my work and with this proposal I would bury them.
> 
> The extract of our discussion could be to go to very pure combined result- and constraint tests. One test-block for each component. Each test-block contains just one component plus additional sinks/sources to make it feasible.

Maybe we are heading in the same direction but have too much discussion here. Let's just talk directly later :smile:

uvchik:
See #359 for my approach. The constraint test will stay lean but the example tests will become 'normal' tests. I think this is a good start for a compromise. The result-test will now work with travis-ci.

ckaldemeyer:
> See #359 for my approach. The constraint test will stay lean but the example tests will become 'normal' tests. I think this is a good start for a compromise. The result-test will now work with travis-ci.

I am wondering about the best approach to integrate this branch into the new testing/examples branch. Merging the dev into this branch after the feature-example-branch has been finished and merged back into dev!?

ckaldemeyer:
> I am wondering about the best approach to integrate this branch into the new testing/examples branch. Merging the dev into this branch after the feature-example-branch has been finished and merged back into dev!?

Same question for https://github.com/oemof/oemof/pull/358

uvchik:
> I am wondering about the best approach to integrate this branch into the new testing/examples branch. Merging the dev into this branch after the feature-example-branch has been finished and merged back into dev!?

I think you could merge `dev` into your feature branch and move the examples to `tests/test_scripts/test_solph`. You may have to rename them a little and make them tests. If you need any help, let me know. I easily could add the `Expected: 5 Got: 4` functionality if you like.

Than you should add a constraint test. Again, if you need help, let me know.

uvchik:
BTW: You can remove `test_solph` if you want.

ckaldemeyer:
> I am wondering about the best approach to integrate this branch into the new testing/examples branch. Merging the dev into this branch after the feature-example-branch has been finished and merged back into dev!?
> I think you could merge dev into your feature branch and move the examples to tests/test_scripts/test_solph. You may have to rename them a little and make them tests. If you need any help, let me know. I easily could add the Expected: 5 Got: 4 functionality if you like.
> 
> Than you should add a constraint test. Again, if you need help, let me know.

Thanks. I just decided to leave the single examples here since for a mere test its o.k. as you did it. The included fix for an error with outputlib has been put into: https://github.com/oemof/oemof/pull/360/commits/e73328aaaee506ea2e8d66b442183ca8de3c9f48

uvchik:
I do not understand. You will not merge your improvements?

ckaldemeyer:
> I do not understand. You will not merge your improvements?

ckaldemeyer:
Weird...

I'll get up early tmr. and finish all my component PRs!

ckaldemeyer:
> Weird...
> 
> I'll get up early tmr. and finish all my component PRs!

So this one, https://github.com/oemof/oemof/pull/358 and https://github.com/oemof/oemof/issues/404#issuecomment-355996766

ckaldemeyer:
This one has become obsolete as we keep the examples in another repo now. I'll create a new PR for some minor polishing..

Issue #350, opened 2017-10-19
csv dispatch example does not work


uvchik:
The example test of csv dispatch example fails due to a run error.

```python
AttributeError: 'OperationalModel' object has no attribute 'timesteps'
```

simnh:
oh, my fault, I will fix this

uvchik:
Obsolete due to #355.

PR #351, opened 2017-10-19
Features/add transformer

uvchik:
* [x] Shall we keep the TransformerN1 and Transformer1N?
 * [x] How do we want to define the conversion factors (see below)?

I replaced the `LinearTransformer` class by a `Transformer` class, while the new class can handle multiple in- **and** outputs. The API  for all Nx1 and 1xN case is still the same. Furthermore you can combine it to NxM.

The test fail manly because of the new nomenclature. The csv-Reader fails and I don't know why. One problem already exist and is not caused by these changes (see #350). 

I kept the equations to keep it compatible with the old N1 and 1N transformer but I wonder why we decided to define the conversion factors for the input flows like this 
`inflow * conv_factor_inflow = outflow` rather than like this `inflow = outflow * conv_factorinflow`. In the example below this would lead to `0.8` and `0.2` instead of `1.25` and `5`. In the end it makes no big difference but we should decide it now.

```python
    bgas = solph.Bus(label="natural_gas")
    bcoal = solph.Bus(label="hard_coal")
    bel = solph.Bus(label="electricity")
    bheat = solph.Bus(label="heat")

    solph.Transformer(
        label="pp_gas",
        inputs={bgas: solph.Flow(), bcoal: solph.Flow()},
        outputs={bel: solph.Flow(), bheat: solph.Flow()},
        conversion_factors={bel: 0.3, bheat: 0.5,
                            bgas: 1.25, bcoal: 5})
```


ckaldemeyer:
> Shall we keep the TransformerN1 and Transformer1N?

As we are breaking the API anyway and I like the idea "There should only be one way to do it." I would vote for only keeping one Transformer class. Otherwise we have very different examples, subclass from three different classes, ...

I think it is of course more effort to adjust everything but worth it!

> How do we want to define the conversion factors (see below)?

To me `outflow = inflow * conv_factor_outflow` is the most intuitive e.g. `P_el=Q_b*Eta_el`.

ckaldemeyer:
Thanks for doing this! Meanwhile I really like the idea of only keeping one `Transformer` as it is very straight and simple.

uvchik:
> To me outflow = inflow * conv_factor_outflow is the most intuitive e.g. P_el=Q_b*Eta_el.

It's pretty obvious this way. But for the `inflow` it is more ambitious because you cannot call it eta. In a heat pump it is called COP and in a mixed fuel boiler I would call it fraction.

In general this means:

`INflow * conversion_factor_OUTflow = OUTflow * conversion_factor_INflow`

This will break the API a little more because by now the TransformerN1 ist implemented like this:

`OUTflow = INflow * conversion_factor_INflow`.

We have to point this out in the release notes, but I think it is clearer this way.

Something like this?

"As the LinearTransformer is depreciated you have to use the Transformer class instead. Please note that the definition of the conversion factors has changed. If you have used the LinearTransformerN1 by now you have to change the class name from `LinearTransformerN1` to `Transformer` **and** you have to change the **conversion_factors** from `n` to `1/n` (e.g. from `bus1: 2` to `bus1: 0.5`) !"

ckaldemeyer:
> Something like this?
> 
> "As the LinearTransformer is depreciated you have to use the Transformer class instead. Please note that the definition of the conversion factors has changed. If you have used the LinearTransformerN1 by now you have to change the class name from LinearTransformerN1 to Transformer and you have to change the conversion_factors from n to 1/n (e.g. from bus1: 2 to bus1: 0.5) !"

Why should we keep the old ones with a warning? Let's just remove it and change it in all examples, the release nodes and the documentation!?

So we break the API but tell everybody..

ckaldemeyer:
> To me outflow = inflow * conv_factor_outflow is the most intuitive e.g. P_el=Q_b*Eta_el.
> It's pretty obvious this way. But for the inflow it is more ambitious because you cannot call it eta. In a heat pump it is called COP and in a mixed fuel boiler I would call it fraction.
> 
> In general this means:
> 
> INflow * conversion_factor_OUTflow = OUTflow * conversion_factor_INflow
> 
> This will break the API a little more because by now the TransformerN1 ist implemented like this:
> 
> OUTflow = INflow * conversion_factor_INflow.
> 
> We have to point this out in the release notes, but I think it is clearer this way.

I think one can argument in every direction. As long as it is well documented, it's o.k. I guess :smile:

simnh:
> "As the LinearTransformer is depreciated you have to use the Transformer class instead. Please note that the definition of the conversion factors has changed. If you have used the LinearTransformerN1 by now you have to change the class name from LinearTransformerN1 to Transformer and you have to change the conversion_factors from n to 1/n (e.g. from bus1: 2 to bus1: 0.5) !"

I would skip the last part of the explanation of how to do it now and refer to the new documentation and docstring. Because i do not really get it right now.

simnh:
definitely remove old ones. From my perspectives we should be as rigorous as possible with throwing out stuff with v0.2...the rest just creates overhead (maintenance, user is confused on what to use, blocks further developments due to incompatibility)

ckaldemeyer:
> definitely remove old ones. From my perspectives we should be as rigorous as possible with throwing out stuff with v0.2...the rest just creates overhead (maintenance, user is confused on what to use, blocks further developments due to incompatibility)

:+1:

uvchik:
> I would skip the last part of the explanation of how to do it now and refer to the new documentation and docstring.

👍

uvchik:
Thank you for you feedback.

1. The Transformer class replaces LinearTransformer etc. (old stuff will be removed)
2. The equation looks like this:
     `flow(i)/conv_factor(i) = flow(o)/conv_factor(o) with o of outflows and i of inflows`
3. Examples and test are adapted and a new test and a new doctest are created.

The last thing I will do is to create a minimal example but I could merge this one soon to avoid merge conflicts in the future as it effects a lot of modules and add the example in a new branch.

@ckaldemeyer , @simnh @gnn Shall I merge now or do you need more time to review?

uvchik:
...furthermore I have to adapt the documentation but I still think we should merge soon. I could adapt the documentation together with the minimal example in a new branch.

ckaldemeyer:
> Thank you for you feedback.
> 
> The Transformer class replaces LinearTransformer etc. (old stuff will be removed)
> The equation looks like this:
> flow(i)/conv_factor(i) = flow(o)/conv_factor(o) with o of outflows and i of inflows
> Examples and test are adapted and a new test and a new doctest are created.
> The last thing I will do is to create a minimal example but I could merge this one soon to avoid merge conflicts in the future as it effects a lot of modules and add the example in a new branch.
> 
> @ckaldemeyer , @simnh @gnn Shall I merge now or do you need more time to review?

If tests are running well I would merge. Very cool! Thanks for your effort :smile:

uvchik:
equation:
http://oemof.readthedocs.io/en/features-add_transformer/api/oemof.solph.html#oemof.solph.blocks.Transformer

network
http://oemof.readthedocs.io/en/features-add_transformer/api/oemof.solph.html#oemof.solph.network.Transformer

documentation
http://oemof.readthedocs.io/en/features-add_transformer/oemof_solph.html#transformer

ckaldemeyer:
> equation:
> http://oemof.readthedocs.io/en/features-add_transformer/api/oemof.solph.html#oemof.solph.blocks.Transformer
> 
> network
> http://oemof.readthedocs.io/en/features-add_transformer/api/oemof.solph.html#oemof.solph.network.Transformer
> 
> documentation
> http://oemof.readthedocs.io/en/features-add_transformer/oemof_solph.html#transformer

Thanks a lot.

> network
> http://oemof.readthedocs.io/en/features-add_transformer/api/oemof.solph.html#oemof.solph.network.Transformer

I just discovered that some code is not formatted to be aligned as example:

`Defining an linear transformer: >>> from oemof import solph >>> bgas = solph.Bus(label=”natural_gas”) >>> bcoal = solph.Bus(label=”hard_coal”) >>> bel = solph.Bus(label=”electricity”) >>> bheat = solph.Bus(label=”heat”)
`

uvchik:
> I just discovered that some code is not formatted to be aligned as example:

Fixed it.

ckaldemeyer:
> Looks good to me. Thanks!

Sorry, one more suggestion (I should have looked at it before):

```
b_gas = solph.Bus(label='natural_gas')
b_coal = solph.Bus(label='hard_coal')
b_el = solph.Bus(label='electricity')
b_th = solph.Bus(label='heat')

solph.LinearTransformer(
    label='pp_chp',
    inputs={b_gas: Flow()},
    outputs={b_el: Flow(nominal_value=30),
             b_th: Flow(nominal_value=40)},
    conversion_factors={b_el: 0.3, b_th: 0.4,
                        b_coal: 0.7, b_gas: 0.3})
```

I think one input is missing here, isn't it?

uvchik:
> Sorry, one more suggestion (I should have looked at it before):

Fixed. Thank you.

uvchik:
@simnh, @gnn May I merge?

uvchik:
I talked to @simnh aside a web conference and orally approved this PR.

Issue #344, opened 2017-10-18
Add gradient  (ramp) costs


simnh:
I just noted that it's not possible to add costs associated with gradients / ramping up and down..

Maybe with the new functionality to just add attributes to the flow we can find a way to easily provide this constraint if such an attribute is present?

ckaldemeyer:
> I just noted that it's not possible to add costs associated with gradients / ramping up and down..
> 
> Maybe with the new functionality to just add attributes to the flow we can find a way to easily provide this constraint if such an attribute is present?

I think this would be of value! But if it is a standard functionality I would define these attributes explicitly for the Flow class. The functionality of adding arbitrary attributes should (from my point of view) be used for real fancy costum components only as it is quite convenient. Otherwise we just have almost empty classes and automatically set args and kwargs in combination with some pyomo-logic behind it ;-)

ckaldemeyer:
Basically if a gradient (https://github.com/oemof/oemof/pull/329/commits/f93f94d03f0e7442f71fee5307a094877294b3a2) and related costs are set, these just have to generate an objective expression, or not?!

simnh:
yeah, it's prettey simple...just a question of where to put it and not ' blow up ' the flow class..but we can put it there.. (or provide a api like this: 

Flow(nominal_value=10, pos_gradient={'ub': 0.5, 'cost': 5}, .. )

ckaldemeyer:
> yeah, it's prettey simple...just a question of where to put it and not ' blow up ' the flow class..but we can put it there.. (or provide a api like this:
> 
> Flow(nominal_value=10, positive_gradient={'ub': 0.5, 'cost': 5}, .. )
> 
> ub for upper bound variable bound
> or
> 
> Flow(nominal_value=10, gradient={'up': 0.5, 'down': 10, 'cost': 5}, .. )
> 
> but then there would be no possibility to different costs....I don't know..I can make a proposal

I would prefer:

    Flow(nominal_value=10, positive_gradient={'ub': 0.5, 'cost': 5}, .. )

ckaldemeyer:
Or something like:

Flow(nominal_value=10, positive_gradient=0.5, positive_gradient_costs=5)

But the dict variant is a bit cleaner I guess..

simnh:
mh..dict would work, but would cause ''problems'' inside the csv-reader I guess..

csv-reader is annoying. 

I don't know...

ckaldemeyer:
[...]

PR #352, opened 2017-10-19
New flow gradient API

simnh:
I gave it a shot but I don't know if its a good idea, kind of hackish...

from #344


ckaldemeyer:
> I gave it a shot but I don't know if its a good idea, kind of hackish...
> 
> from #344

Why do you think it is hacky?

simnh:
Just feels like it right now :), I guess @gnn would improve this. I liked the new generic setting of attributes of the flow and now I started with a dict and messed this nice code up again..but I of course works..and the internal logic can be understood (I think)

simnh:
As we decided to remove the csv-reader, this one can be merged, if the API is ok for everybody...please do so, if you agree @ckaldemeyer , @uvchik

simnh:
I am not sure, when I will have the time...probably in two weeks...

uvchik:
I added a test but either I got the API wrong or this features does not work properly.

I created the following Source object and expected the costs to be part of the objective function. You can find the resulting lp-file [here](https://github.com/oemof/oemof/blob/features/gradient-costs/tests/lp_files/source_with_gradient.lp).

```python
        bel = solph.Bus(label='electricityBus')

        solph.Source(label='powerplant', outputs={bel: solph.Flow(
            nominal_value=999, variable_costs=23,
            positive_gradient={'ub': 0.03, 'costs': 7},
            negative_gradient={'ub': 0.05, 'costs': 8})})
```

uvchik:
@simnh What about this one?

simnh:
i think it should be fine

uvchik:
> i think it should be fine

You mean the `lp-file`?

And what do you think about building the objective parts?

uvchik:
I am fine now. 

@gnn Could you have a look at the attribute definition as discussed above.

@simnh Could you have a look at commit b5a234c. These lines seem to be obsolete to me.

gnn:
Hi there. I don't think it's (too) hackish. It's actually readable and self explanatory.
The only small "gripe" I had at first was the fact that explicit `None`s where stored in the defaults, because that is what `.get`'s differing behaviour from `[]` is for. But after thinking about it, the explicit `None`s are a good way of communicating through the code how to specify no upper bound and that this is the default.

PR #353, opened 2017-10-20
Multi Period

simnh:
this introduces TIMEINDEX as a set (period,timestep) for flows.

### DIFF
--- a/examples/oemof_full_check.py
+++ b/examples/oemof_full_check.py
@@ -121,16 +121,16 @@ def run_example_checks():
         results = None
 
     test_results = {
-        (('wind', 'bel'), 'flow'): 19504.415509800005,
-        (('pv', 'bel'), 'flow'): 4205.0081990299996,
-        (('bel', 'demand_el'), 'flow'): 95893.690674837926,
-        (('bel', 'excess_el'), 'flow'): 353.30769076999997,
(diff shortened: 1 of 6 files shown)


simnh:
Can you have a look here, and check if'ts ok for you. basically for now it's just internal handling, so for all current problems only the index should change inside the lp-files.

ckaldemeyer:
> Can you have a look here, and check if'ts ok for you. basically for now it's just internal handling, so for all current problems only the index should change inside the lp-files.

Could you provide a minimal example for a multi-period usage? Or is it not working?

> I did not yet adapt CAES and GenericCHP

There are no tests so far ;-)

simnh:
yeah because before I write test, provide an example and adapt existing test and fix all bugs, I like to now if the general idea  / direction of where the code is goiing is ok for you?

ckaldemeyer:
> yeah because before I write test, provide an example and adapt existing test and fix all bugs, I like to now if the general idea / direction of where the code is goiing is ok for you?

As long as everything works I am fine. By having a quick look it looks good for me!

simnh:
I won't have the time to do it for the next version. therefore I moved the milestone, to 'nice-to-have'

p-snft:
Closed in favour of #750, which continues and expands what has been started here.

PR #357, opened 2017-10-20
Normalize lp-file equations to non-negative RHSs

### LINKED ISSUE #254: New pyomo version breaks oemof tests

gnn:
It's not beautiful but it fixes #254 by always normalizing equations in lp-files to have non-negative right hand sides when comparing them.


uvchik:
We removed the `doc/whatsnew/v0-1-5.rst` as we directly heading v0.2.0. 

@gnn Please make sure your changes will appear in the `doc/whatsnew/v0-2-0.rst` file.

uvchik:
As I understand the workflow:

@gnn fix
@uchick approve
@gnn merge

Or do I misunderstand something.

uvchik:
🎉 Thank you !!

PR #358, opened 2017-10-23
Features/custom component caes

ckaldemeyer:
This adds a generic model for CAES as a first custom component and provides a simple example.

It is not meant to be made prominent but shows how one can integrate user defined models for technologies!


ckaldemeyer:
@uvchik @simnh : Would it also be o.k. for you to go this direction (integrating custom components and making not all prominent)?

uvchik:
What implication does "not making it prominent" have?

ckaldemeyer:
> What implication does "not making it prominent" have?

Not updating WHATSNEW or having to provide examples except development examples

uvchik:
We should collect the minimal requirements for a PR with a new component. And we should be a good example 😏

ckaldemeyer:
> We should collect the minimal requirements for a PR with a new component. And we should be a good example

I would like to format the examples in https://github.com/oemof/oemof/pull/349 and this PR in the same style and include a results tests as doctest (we can discuss if we skip the lp-file-check via telephone). If you and @simnh are fine, I would merge asap..

simnh:
I am fine @ckaldemeyer

uvchik:
I am fine with this, too. I just think we should provide a list for developer who wants to add a custom component. Therefore I thought @ckaldemeyer could just write down what he did.

ckaldemeyer:
> I am fine with this, too.

:+1: 

> I just think we should provide a list for developer who wants to add a custom component. Therefore I thought @ckaldemeyer could just write down what he did.

What do you mean exactly? Wouldn't it be sufficient to just look at the code?

uvchik:
We can do that at the meeting. I just thought of a list: "What do I have to do if I want to add a new component to solph". As you just added a component it might be easy for you to start such a list.

But this has nothing to do with your component and the ability to merge it.

ckaldemeyer:
> We can do that at the meeting. I just thought of a list: "What do I have to do if I want to add a new component to solph". As you just added a component it might be easy for you to start such a list.
> 
> But this has nothing to do with your component and the ability to merge it.

Good idea.

ckaldemeyer:
https://github.com/oemof/oemof/issues/365

ckaldemeyer:
I'll care about this on the dev-meeting next week.

PR #359, opened 2017-10-24
Restructure of the test environment

uvchik:
Restructure of the test environment (thanks to @ckaldemeyer  for the feedback)

 * reduce examples to test (only nosetests, but feel free to write doctest instead)
 * @birgits will restore the examples in the examples package of oemof
 * reduce test-examples and csv-files to a minimal number of time steps to reduce the execution time
 * replace the "oemof_examples" script with "oemof_installation_test" (from now on only the installation test can be execute with a script)
 * have a look at the installation test and add more critical parts that could be test and write a feedback statement for the user

Now you can execute **all** test (including the results tests) in one line.

```sh
nosetests --with-doctest -w your/path/to/oemof
```

The result should look like this 😄 
```pyhton
.....................................................
----------------------------------------------------------------------
Ran 53 tests in 5.083s

OK
```


uvchik:
Now every new component has to provide only two tests:

**ConstraintTest** (as a method in the constraint_test module)

```python
def test_my_component
    bus1 = solph.Bus('Bus1')
    bus2 = solph.Bus('Bus2')
    etc.

    MyNewComponent(
            label='name_of_component',
            inputs={bus1: solph.Flow(...), ...},
            outputs={bus2: solph.Flow(...), ...},
            my_attribute=5)
    self.compare_lp_files('my_component.lp')
```
Now you can run it once, check the new lp-file. If it looks right you can copy it to the `lp-files` to make the test permanent. That's all.

**ResultTest** (as a module in the test_scripts folder)

You just have to write a little script to produce results and compare the results within an assert test. There are already some examples that will help you to understand how it works (see [test_generic_chp](https://github.com/oemof/oemof/blob/features/restructure_test_environment/tests/test_scripts/test_solph/test_generic_chp/test_generic_chp.py)). This may look pretty similar to your own test script when you developed the component, so it should not be much work.

@gnn, @simnh, @ckaldemeyer What do you think?

simnh:
Go ahead, I like the ideas!

ckaldemeyer:
> Go ahead, I like the ideas!

I am also fine with it. It is much cleaner now and I can live without doctests! ;)

uvchik:
So be it 🎉  @ckaldemeyer Thank you very much for your ideas and contribution.

PR #366, opened 2017-11-01
Connect investment variables - constraint

uvchik:
Connecting investment variables ([see example](https://github.com/oemof/oemof/blob/features/connect_investment_variables_constraint/tests/test_scripts/test_solph/test_connect_invest/test_connect_invest.py)).

I am unsure with the API. At the moment you have to pass a list of tuples. The tuple consists of the investment variable and a factor. The factor is needed to make e.g. 1 * inv_var1 == 2 * inv_var2 etc.. 

```python
om = solph.Model(energysystem)

invest = [(om.InvestmentFlow.invest[line12, bel2], 1),
          (om.InvestmentFlow.invest[line21, bel1], 0.5),
          (om.InvestmentFlow.invest[lineg1, bel1], 1),
          (om.GenericInvestmentStorageBlock.invest[storage], 1)]

solph.constraints.connect_investment_variables(om, invest)
```

@oemof/oemof-solph What do think about the API. Any better ideas? A dictionary?

 * [x] Add tests.
 * [x] Add equation to docstring


ckaldemeyer:
> Connecting investment variables (see example).
> 
> I am unsure with the API. At the moment you have to pass a list of tuples. The tuple consists of the investment variable and a factor. The factor is needed to make e.g. 1 * inv_var1 == 2 * inv_var2 etc..
> 
> om = solph.Model(energysystem)
> 
> invest = [(om.InvestmentFlow.invest[line12, bel2], 1),
>           (om.InvestmentFlow.invest[line21, bel1], 0.5),
>           (om.InvestmentFlow.invest[lineg1, bel1], 1),
>           (om.GenericInvestmentStorageBlock.invest[storage], 1)]
> 
> solph.constraints.connect_investment_variables(om, invest)
> @oemof/oemof-solph What do think about the API. Any better ideas? A dictionary?

From the code, I do not understand which ivestment variables are connected here and find it a bit confusing.

Wouldn't it be more specific/self-explaining if you define let's say a function `set_equal(var1, var2, factor)` and use this to  create constraints. Just as an idea..

But maybe I just didn't understand the intention or got something wrong ;-)

uvchik:
I was thinking of a function to equalise more than one variable. But I think you are right, it might be a lot clearer if I just add function to equalise two variables. This will cover most cases and still gives you the opportunity to write your own loop if you have more than one variables. 

I think of renaming `factor` to `factor1` to make it clearer that it will be multiplicated with `var1`.

henhuy:
Hi @uvchik,
I looked at your example and would join @ckaldemeyer recommendation.
Perhaps it would be nice to use namedtuple (i like those!) instead of dict?
Here is an example:
```
from collections import namedtuple
from oemof.solph import Flow, Transformer, Investment

InvestmentRelation = namedtuple(
    'InvestmentRelation',
    ['investment1', 'investment2', 'fraction']
)

line12_invest = Investment(ep_costs=20)
line12 = Transformer(
    label="line12",
    inputs={bel1: Flow()},
    outputs={bel2: Flow(investment=line12_invest)})

line21_invest = Investment(ep_costs=20)
line21 = Transformer(
    label="line21",
    inputs={bel2: Flow()},
    outputs={bel1: Flow(investment=line21_invest)})

invest_relation12 = InvestmentRelation(
    line12_invest, line21_invest, fraction=0.3)

solph.constraints.connect_investment_variables(om, [invest_relation12])
```
Thereby, you already would have an str-representation (i.e. `InvestmentRelation(investment1='inv1', investment2='inv2', fraction='0.5')`) of the relation and you could nicely get the attributes via `invest_relation12.investment1`. Also, you could override `__str__` of the namedtuple (as suggested here: https://stackoverflow.com/a/7914193/5804947) to have an even nicer representation (i.e. "invest1 = 0.5 x invest2").

uvchik:
@henhuy Thank you for the feedback, but I rather would like to keep it as simple as possible. That was the reason why I did not like my first approach. As I wrote above I will follow @ckaldemeyer's idea.

uvchik:
I changed the API to set only two variables to equal. I had to define a name each time the function is called to avoid the constraint to be overwritten by itself, so made it possible to define you own name even though it is not necessary. New API:

```python
equate_variables(om, var1, var2, factor1=1, name=None)
```

The example from my first post now looks like this:

```python
om = solph.Model(energysystem)

solph.constraints.equate_variables(
    om, om.InvestmentFlow.invest[line12, bel2], om.InvestmentFlow.invest[line21, bel1], 2)
solph.constraints.equate_variables(
    om, om.InvestmentFlow.invest[line12, bel2], om.InvestmentFlow.invest[lineg1, bel1])
....
```

@ckaldemeyer, @simnh  Please review.

uvchik:
If everything else is fine I will shorten the script test before merging.

PR #377, opened 2017-11-24
add a new component GenericEngineGenerator with a linearized relation

mloenneberga:
I added a new component with class.Transformer being its parent class. I changed the input_output_relation of class.GenericEngineGenerator compared with class.Transformer. It allows for a linear approximation of the input (fuel consumption) in relation to the output of the GenericEngineTransformer. This component helps to model combustion engine generators e.g. diesel generators for micro-grids. 

What do you think?


uvchik:
I like the approach but I think we should do two things.

1. The general idea extends our existing transformer by a second factor which could be also part of the Transformer class.
2. We can still keep you component as a specialist component for micro grids.

```diff
- flow(i,n,t) / conv_factor(n,i,t) = flow(n,o,t) / conv_factor(n,o,t)
+ flow(i,n,t) / conv_factor(n,i,t) + offset(n,i,t) = flow(n,o,t) / conv_factor(n,o,t) + offset(n,o,t)
```

If the offset is zero we do have the actual equation of our transformer. So we could extend it to the general form with extending the API of the Transformer class. For example passing a tuple instead of one value. If one value is passed the offset is assumed to be zero (nothing will change) but if a tuple is passed the first value will be the conv_factor and the second value will be the offset.

ckaldemeyer:
> I like the approach but I think we should do two things.
> 
> The general idea extends our existing transformer by a second factor which could be also part of the Transformer class.
> We can still keep you component as a specialist component for micro grids.
> - flow(i,n,t) / conv_factor(n,i,t) = flow(n,o,t) / conv_factor(n,o,t)
> + flow(i,n,t) / conv_factor(n,i,t) + offset(n,i,t) = flow(n,o,t) / conv_factor(n,o,t) + offset(n,o,t)
> If the offset is zero we do have the actual equation of our transformer. So we could extend it to the general form with extending the API of the Transformer class. For example passing a tuple instead of one value. If one value is passed the offset is assumed to be zero (nothing will change) but if a tuple is passed the first value will be the conv_factor and the second value will be the offset.
> Merge state

I agree. And one more question: Do you have any objective expressions e.g. fixed costs, etc. for your component? I guess this should be the case in a similar way to the GenericCHP component or other plants.

Thanks for contributing!

simnh:
Does the component also work with all 'blocks', e.g. Investment, NonConvex etc? 

I guess some thoughts may be put into how nonconvex P-min constraints (Big-M) are working together with this component (or rather it's flows and the conversionfactor)

ckaldemeyer:
> I guess some thoughts may be put into how nonconvex P-min constraints (Big-M) are working together with this component (or rather it's flows and the conversionfactor)

Shouldn't we discuss this in combination with (additional) constraints for different types of Flow?

jnnr:
It seems to me that this PR still needs some discussion? Please correct me if I am wrong!

simnh:
Yes I think, this component has not been developed completely.

uvchik:
I wonder if it is obsolete with #424? 

@mloenneberga  What do you think?

mloenneberga:
Indeed, PR #377 is obsolete due to #424, which can be regarded as a more generic formulation of #377

PR #378, opened 2017-11-28
revise tools sub-package

uvchik:
Revision of the tools sub-package.

* remove config.py -> oemof does not use config files
* remove time logging function -> use datetime.datetime.now() instead.
* revise logger function -> straight forward, easily customisable but with default values for easy access
* remove default config files -> not needed any more
* one unused helper function is removed -> not needed any more
* add remaining modules to init file -> resolves import problems


uvchik:
DOCUMENTATION

The main purpose of this function is to provide a logger with well set default values but with the opportunity to change the most important parameters if you know what you want after a while. This is what most new users (or users who do not want to care about loggers) need.
If you are an advanced user with your own ideas it might be easier to copy the whole function to your application and adapt it to your own wishes.

```python
def define_logging(logpath=None, logfile='oemof.log', file_format=None,
                   screen_format=None, file_datefmt=None, screen_datefmt=None,
                   screen_level=logging.INFO, file_level=logging.DEBUG,
                   log_version=True, timed_rotating=None):
```

By default down to INFO all messages are written on the screen and down to DEBUG all messages are written in the file. The file is placed in `$HOME/.oemof/log_files` as `oemof.log`. But you can easily pass your own path and your own filename. You can also change the logging level (screen/file) by changing the `screen_level` or the `file_level` to logging.DEBUG, logging.INFO, logging.WARNING.... . You can stop the logger from logging the oemof version or commit with `log_version=False` and you can change the format on the screen and in the file according to the [python logging documentation](https://docs.python.org/3/library/logging.html#logrecord-attributes). You can also change the used time format according to [this documentation](https://docs.python.org/3/library/time.html#time.strftime).

```python
file_format = "%(asctime)s - %(levelname)s - %(module)s - %(message)s"
file_datefmt = "%x - %X"
screen_format = "%(asctime)s-%(levelname)s-%(message)s"
screen_datefmt = "%H:%M:%S"
```

You can also change the behaviour of the file handling (TimedRotatingFileHandler) by passing a dictionary with your own options (timed_rotating).

See the API-doc for more information.

ckaldemeyer:
> thx! I think this cleans up the code significantly!

I agree. Thanks! :+1: 

@simnh You have to approve!

PR #380, opened 2017-11-30
Features/node flows

henhuy:
Hi everyone,

I talked to @uvchik today and we discussed how to add the Flow components to node results.
As I worked on that lately, I added my solution here.
Now, if `get_flows` is set to `True` in `node` function, flows are added as new dictionary containing the  tuple `((from_node, to_node), flow_name)` from `sequences` columns as keys and a namedtuple containing `Flow` and `FlowType` as values.
Here is an example how to get all input Flows of a storage node:
```
storage_results = views.node(results, storage, get_flows=True)
print(
    list(
        nodes for (nodes, flow_name), flow in
        storage_results['flows'].items()
        if flow.type == 'input'
    )
)
```

Additionally I added two functions I often need to filter nodes from results or to get node via label.

Are you fine with this approach? Please give me feedback on this!

### DIFF
--- a/oemof/outputlib/views.py
+++ b/oemof/outputlib/views.py
@@ -118,3 +118,18 @@ def filter_nodes(results, option=NodeOption.All, exclude_busses=False):
         return {n for n in nodes if not isinstance(n, Bus)}
     else:
         return nodes
+
+
+def get_node_by_name(results, *names):
+    """
+    Searches results for nodes
+
+    Names are looked up in nodes from results and either returned single node
+    (in case only one name is given) or as list of nodes. If name is not found,
+    None is returned.
+    """
+    nodes = filter_nodes(results)
+    if len(names) == 1:
+        return next(filter(lambda x: str(x) == names[0], nodes), None)
+    else:
+        return [n if str(n) in names else None for n in nodes]
--- a/oemof/outputlib/views.py
+++ b/oemof/outputlib/views.py
@@ -8,8 +8,23 @@ Information about the possible usage is provided within the examples.
 from enum import Enum
 from itertools import zip_longest
 import pandas as pd
+from collections import namedtuple
 
 from oemof.network import Bus
+
+
+# NodeFlow contains oemof.network.Flow or oemof.network.Node component and
+# related FlowType (see below):
+NodeFlow = namedtuple('NodeFlow', ['component', 'type'])
+
+
+class FlowType(str, Enum):
+    """
+    Gives information on flow type
+    """
+    Single = 'single'
+    Input = 'input'
+    Output = 'output'
 
 
 def convert_keys_to_strings(results):
@@ -24,7 +39,7 @@ def convert_keys_to_strings(results):
     return converted
 
 
-def node(results, node):
+def node(results, node, get_flows=False):
     """
     Obtain results for a single node e.g. a Bus or Component.
 
@@ -67,6 +82,12 @@ def node(results, node):
         cols = [c for sublist in cols for c in sublist]
         filtered['sequences'].columns = cols
         filtered['sequences'].sort_index(axis=1, inplace=True)
+
+    if get_flows:
+        filtered['flows'] = {}
+        for nodes, flow_name in filtered['sequences']:
+            filtered['flows'][(nodes, flow_name)] = __get_flow_component(
+                nodes, node)
 
     return filtered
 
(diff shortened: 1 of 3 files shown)


ckaldemeyer:
> Hi everyone,
> 
> I talked to @uvchik today and we discussed how to add the Flow components to node results.
> As I worked on that lately, I added my solution here.
> Now, if get_flows is set to True in node function, flows are added as new dictionary containing the tuple ((from_node, to_node), flow_name) from sequences columns as keys and a namedtuple containing Flow and FlowType as values.
> Here is an example how to get all input Flows of a storage node:
> 
> storage_results = views.node(results, storage, get_flows=True)
> print(
>     list(
>         nodes for (nodes, flow_name), flow in
>         storage_results['flows'].items()
>         if flow.type == 'input'
>     )
> )
> Additionally I added two functions I often need to filter nodes from results or to get node via label.
> 
> Are you fine with this approach? Please give me feedback on this!

Sorry, but I do not really understand what it is doing. Could you provide a more illustrative examples?

Anyway, thanks already for contributing!

ckaldemeyer:
> Now, if get_flows is set to True in node function, flows are added as new dictionary containing the tuple ((from_node, to_node), flow_name) from sequences columns as keys and a namedtuple containing Flow and FlowType as values.

Could you maybe describe the newly created data struxcture?

henhuy:
> Could you maybe describe the newly created data structure?

Of course!
Calling `node` function gives you back a dictionary containing 'scalars' and 'sequences'. The sequences entry itself contains a `pandas.DataFrame` which holds timeseries/sequences of current node - the index is a timeindex and column headers are in the form `((from_node, to_node), "flow")`.

Given the parameter `get_flows=True` an additional key 'flows' is added to the node results dictionary having the upper mentioned column headers as keys and a new structure `NodeFlow` as value. Thereby, `NodeFlow` is a simple `namedtuple` from pythons `collections` package with attributes `component` and `type`. The `component` attribute holds the related `oemof.network.Flow` for given node tuple, while `type` describes if this Flow is a "single", "input" or "output" flow. Internally, the FlowType is inherited by pythons `Enum`, but can be accessed as normal `str` as well. 

Another example maybe:
```
storage_results = views.node(results, storage, get_flows=True)
flow = storage_results['flows'][((b_el, storage), 'flow')]
print(type(flow.component))
>>> oemof.network.Flow
print(flow.type)
>>>'output'
```

A special case is given for the storage capacity, as its column header equals `((node, ) 'capacity')` - in this case resulting `flow.component` holds the node itself and `flow.type` is "single".

I hope this makes it clearer...

uvchik:
Does this work? (I think for examples `>>>` is the input, but I may be wrong)
```python
>>> storage_results = views.node(results, storage, get_flows=True)
>>> flow = storage_results['flows'][((b_el, storage), 'flow')]
>>> print(flow.component.variable_costs)
25
```
If it works like this it is a good way to get all information around one component.

But if I want to  have the sequence of all input flows of a bus it is still kind of complicated.
```python
bus_results = views.node(results, 'my_bus', get_flows=True)
in_cols = [nodes for (nodes, flow_name), flow in bus_results['flows'].items()
           if flow.type == 'input']
df_input_flow_sequences = bus_results['sequences'][in_cols]
```

Maybe we could find a way to make that more simple because I think it is not a rare case. We could add an attribute that will return MulitIndex columns for the sequences or we just write a new function based on your internal logic.
```python
bus_results = views.node(results, 'my_bus', get_flows=True)
df_with_multiindex_columns = split_sequences(bus_results['sequences'])
df_with_outflow_sequences = df_with_multiindex_columns['out']
df_with_inflow_sequences = df_with_multiindex_columns['in']
```

henhuy:
> Does the following work?
> ```python
> >>> storage_results = views.node(results, storage, get_flows=True)
> >>> flow = storage_results['flows'][((b_el, storage), 'flow')]
> >>> print(flow.component.variable_costs)
> 25
> ```
Yes, this is how it's supposed to work.

Maybe it's good to define a function `flows(node, flow_type=None)` which either returns all flows like in my above implementation if flow_type is None, or it returns all flows of certain kind ('single', 'input' or 'output') as dictionary with nodes+flow_name as key and flow itself as value. Additionally, the `get_flows` attribute of the `node` function could call the `flows` function with `flow_type=None`. Then you could do something like the following to get fuel costs of a generator for example:
```
gen_results = views.node(results, generator)
inflows = flows(generator, flow_type='input')
# you can also use Enum-typing: inflows = flows(generator, flow_type=FlowType.Input)
fuel_costs = 0.0
for nodes, flow in inflows.items():
    fuel_costs += gen_results['sequences'][nodes] * flow.variable_costs
```

For me your MultiIndex Dataframe approach sounds a bit overdone - but this is a personal feeling...

uvchik:
> For me your MultiIndex Dataframe approach sounds a bit overdone - but this is a personal feeling...

This was just an idea, the flows function is fine for me, too 😄

henhuy:
@uvchik can you check if this fits your needs?

ckaldemeyer:
I think this is postponed/obsolete as @henhuy and me had a telephone meeting today and will implement a single dict object for all params in the same style as the variable result dict.

PR #383, opened 2017-12-04
Add lopf to solph

simnh:


simnh:
Great news! Minimal lopf example is running! checkout oemof examples repository feature branch

https://github.com/oemof/oemof_examples/blob/features/lopf/examples/oemof_0.2/electrical/simple_lopf.py

example reproduces pypsa minimal example results: 

https://www.pypsa.org/examples/minimal_example_lopf.html

oemof/solph/models.py
```diff
@@ -96,7 +96,8 @@ def __init__(self, es, **kwargs):
 
         # non-negative pyomo variable for all existing flows in energysystem
```
uvchik: I think these lines will break existing models.

Would it be okay to add a flow option like "NonNegative=True" and you can set it to False to make lopf possible?

oemof/solph/models.py
```diff
@@ -96,7 +96,8 @@ def __init__(self, es, **kwargs):
...
         self.flow = po.Var(self.FLOWS, self.TIMESTEPS,
-                           within=po.NonNegativeReals)
```
uvchik: I think these lines will break existing models.

Would it be okay to add a flow option like "NonNegative=True" and you can set it to False to make lopf possible?

oemof/solph/models.py
```diff
@@ -96,7 +96,8 @@ def __init__(self, es, **kwargs):
...
+                           within=po.Reals)
```
uvchik: I think these lines will break existing models.

Would it be okay to add a flow option like "NonNegative=True" and you can set it to False to make lopf possible?

simnh: I added a condition to set the lower bound to zero. Acutally, I think it would be great if generally we allow to be the variables within Reals and then restrict it by bounds (min, max etc). Actually I think it should not have broken model results but only lp-files before

uvchik: > I think it would be great if generally we allow to be the variables within Reals and then restrict it by bounds (min, max etc)

I thought the default value of min is "0" anyway shouldn't it work without your hack? If min=0 is set we do not need my proposal above. But somehow it does not seem to work.

simnh: yes thats what i meant, but it ll definitely break lp files. the thing is i guess this may be a problem for unrestricted flows, e.g. from a gas source to a gas bus. in that case no bounds are set and the change from NonNegReals to Reals changes the flow bounds for these flows. but in that case one line should also be sufficient : 

`self.flow[o, i, t].setlb(0)`

This would mean: first set all flows lower bound to '0' and then maybe reset if `min` is negative

uvchik: If I understand you correctly this is similar to what I was thinking of this morning 😄 

1. At the moment the default value of min is "0". That means that we could check if any value within the min sequence is below 0. If this is the case we could set the variable to `Reals` otherwise to `NonNegReals`.

2. In the next step we allow components to overwrite the default value of the connected flows. For example the "electrical_line" could loop over all flows and set `min` to -1 if `min` is `None`.

In that case the changes are not somewhere in the code but directly within the definition of the component. I think this is more explicit.

We could of course discuss about what is the default behaviour and what is the exception in the future but by now I would keep as described above.

uvchik: BTW : Breaking the lp-files is no problem for me as long as the the models still work.

simnh: Not possible to set Reals for some flows and NonReals for some other flows, either the one or the other for all variables. In my opinion we should just do flow as reals and set lower bounds based on flow objects and keep default value of 0 for all flows. we just need to make that work.

uvchik: > Not possible to set Reals for some flows and NonReals for some other flows, either the one or the other for all variables.

Did not know that.

> we just need to make that work.

That's right 😏 Does `self.flow[o, i, t].setlb(min[t])` work? Will this make the problem bigger/slower? In that case we could think of a NonNegativeFlowModel as special child Model.

simnh:
Now it should not break other models anymore, Right? If so, thats strange. 

I will deal with the rest of the problems regarding electrical flows or flows that may take negative values somewhere else...

oemof/solph/models.py
simnh: not bigger or slower to solve, maybe some milliseconds slower to construct. 

but of course it would need to be min * nominal_value ...therefore I like the idea that I implemented below: just make the default `0`, but allow for negative values. This would not change anything  (API or models etc) for anyone but allow for more developing flexibility for us

simnh:
Ok, I again rewrote the block and changed a check inside the investment flow. Though internally it's a little hacky the components ElectricalLine should now be working, both for investment and dispatch. I assume also for NonConvex, but I will have to check. 

@uvchik: Now, does it break any models, I don't think so, right?

simnh:
I added default value of -1 to flows of electrical lines, now we just need the check in the models.py @uvchik, as you had an idea, can you try to implement it?

uvchik:
I fixed the merge conflicts but we still have some strange behaviour....

uvchik:
I fixed some tests:

1. I think aa51efa was a typo.
2. I changed it because it does not work 90516a4. Maybe because of empty sets.
3. Just following the new rules (af662ae, ec858f1)

Now there are only constraint tests failing. I will have a look at them and fix them afterwards.

uvchik:
YES!

I think I fixed the constraint test. The low bound fixed the variable to >= 0 but there was an additional constraint: flow >= 0. Commit  004ce3 prevents the InvestmentFlow from creating these obsolete constraints.

uvchik:
Sorry for messing around with the imports (last four commits). Now all tests are fine.

uvchik:
@nesnoj Is this feature interesting for you?

simnh:
@uvchik thanks a lot. I just realized that simply using `min` is kind of not nice. For BidirectionalFlows we would rather need something like: min = (forward, backward) , max = (forward, backward). 

But as for custom components, this may be alright for now. I will think about adding bidi-edges, which will probably have another API than `Flow`. In that processes I just realized that `Flow` should actually be `Edge`, maybe something for version 0.3...I guess there is lots of renaming necessary actually. But this is of topix here. I'll add an issue

uvchik:
> For BidirectionalFlows we would rather need something like: min = (forward, backward) , max = (forward, backward).

Also a good idea 😄  `if isinstance(min, tuple) -> BIDIRECTIONAL else UNI`

We can test it and improve it a while. So far it does not even effect the existing lp-files.

simnh:
nice, thank you ;) ....I am actually already working on the next object, so its good to merge the lopf

simnh:
@ @gplssm @nesnoj can somebody of you electricity experts (maybe someone else from eGo) check this functionality. We could also maybe compare to eGo-lopf...not sure how complicated that would be...

simnh:
@gnn there is one problem left: if the user does not set a 'slack' bus, its set automatically. This should be the same for the same problem if possible. Can you do this, or someone else?

oemof/solph/custom.py
```diff
@@ -1,18 +1,191 @@
...
+        self.ELECTRICAL_BUSES = po.Set(initialize=[n for n in m.es.nodes
+                                       if isinstance(n, ElectricalBus)])
+
+        def _voltage_angle_bounds(block, b, t):
+            return b.v_min, b.v_max
+        self.voltage_angle = po.Var(self.ELECTRICAL_BUSES, m.TIMESTEPS,
+                                    bounds=_voltage_angle_bounds)
+
+        if True not in [b.slack for b in self.ELECTRICAL_BUSES]:
+            # TODO: Make this robust to select the same slack bus for
```
simnh: This TODO is maybe a good idea before merge

uvchik:
It is not necessary for this PR but as it is a big new feature we should soon create some test. Otherwise changes within the solph code base may break these components.

ckaldemeyer:
Should I review, again now or are you still busy?

uvchik:
Thank you for this feature. May we delete the branch?

ckaldemeyer:
> Thank you for this feature. May we delete the branch?

Oh, I haven't seen that you have already merged. Thanks also from my side!

Issue #387, opened 2017-12-04
Unpacking of results nodes-tuple


henhuy:
I stumbled upon the nodes-tuple in results output or `views.node` function.
The "problem" is that I want to extract the `sequences` in a nice way, like this:
```
node_results = views.node(results, node)
for ((from_node, to_node), flow_name), timeseries in node_result.items()
    ...
```
This works nicely, in case no single node-tuple is involved. As soon as I do this for a storage, I ran into the following error:
`Value Error: not enough values to unpack`
because inside it the capacity series is build as (storage,)-tuple, which cannot be unpacked into two values. Possible workaround is as follows:
```
node_results = views.node(results, node)
for (nodes, flow_name), timeseries in node_result.items()
    from_node = nodes[0]
    to_node = nodes[1] if len(nodes) > 1 else None
    ...
```
My question is, if it would be possible to set 2nd node to `None` in case of single-node-tuples (ie. storage) in the **results creation**? Or does this lead to further complications? Or is it to late for the release?

ckaldemeyer:
> I stumbled upon the nodes-tuple in results output or views.node function.
> The "problem" is that I want to extract the sequences in a nice way, like this:
> 
> node_results = views.node(results, node)
> for ((from_node, to_node), flow_name), timeseries in node_result.items()
>     ...
> This works nicely, in case no single node-tuple is involved. As soon as I do this for a storage, I ran into the following error:
> Value Error: not enough values to unpack
> because inside it the capacity series is build as (storage,)-tuple, which cannot be unpacked into two values. Possible workaround is as follows:
> 
> node_results = views.node(results, node)
> for (nodes, flow_name), timeseries in node_result.items()
>     from_node = nodes[0]
>     to_node = nodes[1] if len(nodes) > 1 else None
>     ...
> My question is, if it would be possible to set 2nd node to None in case of single-node-tuples (ie. storage) in the results creation? Or does this lead to further complications? Or is it to late for the release?

This is no problem and would work really quickly. We had been discussing this in the respective PR and the discussion was about whether it is `(my_node,)` or `(my_node, my_node)`. The decision was made for `(my_node,)` but not for a particular reason as it was rather a matter of taste.

But obviously the "empty" tuple element makes tasks like slicing a bit harder, induces more code and we should change it then.

The following options (including your option, which I also like), would be very quickly to implement:

* `(my_node, my_node)`
* `(my_node, None)`

[...]

PR #389, opened 2017-12-05
Unpacking of results nodes-tuple #387

henhuy:
Is related to #387 

I adapted `processing.results` function in order to return `(node, None)`-tuples as keys in case of single node result.
This change led to an error in `views.node` function:
"`AttributeError: 'NoneType' object has no attribute 'label'`"
as the `sort_index` function does not work with `None` as comparison, due to `Nodes` internal implementation of the `__lt__` function. 
Therefore, I commented the `sort_index` line out.

Everyone okay with that, or is the sorted index crucial for any implementation?

### DIFF
--- a/oemof/network.py
+++ b/oemof/network.py
@@ -217,6 +217,8 @@ class Node:
         return id(self) == id(other)
 
     def __lt__(self, other):
+        if other is None:
+            return False
         return self.label < other.label
 
     def __hash__(self):
--- a/oemof/outputlib/processing.py
+++ b/oemof/outputlib/processing.py
@@ -98,19 +98,20 @@ def results(om):
     Results from Pyomo are written into a dictionary of pandas objects where
     a Series holds all scalar values and a dataframe all sequences for nodes
     and flows.
-    The dictionary is keyed by the nodes e.g. `results[(n,)]['scalars']`
+    The dictionary is keyed by the nodes e.g. `results[(n, None)]['scalars']`
     and flows e.g. `results[(n,n)]['sequences']`.
     """
     df = create_dataframe(om)
 
     # create a dict of dataframes keyed by oemof tuples
-    df_dict = {k: v[['timestep', 'variable_name', 'value']]
+    df_dict = {k if len(k) > 1 else (k[0], None):
+               v[['timestep', 'variable_name', 'value']]
                for k, v in df.groupby('oemof_tuple')}
 
     # create final result dictionary by splitting up the dataframes in the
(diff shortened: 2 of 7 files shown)


henhuy:
okay - unfortunately the tests are using fixed sorted index!
I will change the tests one more time to dynamically get results...

ckaldemeyer:
> Is related to #387
> 
> I adapted processing.results function in order to return (node, None)-tuples as keys in case of single node result.
> This change led to an error in views.node function:
> "AttributeError: 'NoneType' object has no attribute 'label'"
> as the sort_index function does not work with None as comparison, due to Nodes internal implementation of the __lt__ function.
> Therefore, I commented the sort_index line out.

Let's just fix the _lt_ solution in the `Node` class! Sorting is needed and we can then leave the tests as they are. @gnn Would this be a problem?

henhuy:
Okay - I can do this as well once @gnn approved your suggestion.
I would easily check if attribute `label` exists and otherwise set None to be "smaller" than any existing label?!

gnn:
Approved. Except:  please don't check whether the `label` attribute exists, but explicitly check for `None`, if possible. If that's not possible, check for the `label` attribute.
Reason: there might be other objects with a `label` attribute that we really don't want to be able to be compared to `Node`s, as we don't have any idea whether the `label` attribute means the same for the other objects as for `Node`s. So only enabling comparison with `None` is just a precaution to still error out early instead of hiding weird buggy corner cases. But as these might not ever occur, getting this working has higher priority, so if the cautious solution isn't possible, so be it.

Slightly nitpicky remark: a better commit message for the first commit would've been something along the lines of

  >  Change `processing.results`' keys to always be pairs
  >
  > Done by inserting `None` as the second component, i.e. the target, of every key which doesn't have one.
  > Fixes #387.

Reasons:

  * First line shorter than 50 characters and properly capitalized.
  * First line immediately conveys what's happening, rest of the messages explains why it's happening.
  * Relation to #387 is stored in the commit message. Ideally there would be a regression test for the issue, explaining the wrong, desired and final behaviour, which can then easily be related to the commit.
  * #387 will automatically be closed upon merging into `dev`.

But this really is just nitpicking for the future. ;) The new code is awesome and changing the ordering of `Node`s to include comparison with `None` is the obvious and straightforward solution.

gnn:
@henhuy: wanted to try something out, so already made the changes. Look at the PR in your fork, check out the changes and merge or amend them as you see fit.

Issue #334, opened 2017-08-28
Fixed costs for Flows


ckaldemeyer:
I am just double-checking some solph expressions for a calculation which includes fixed costs.

In the [documentation](http://oemof.readthedocs.io/en/stable/api/oemof.solph.html#oemof.solph.network.Flow) it says 

> fixed_costs (numeric) – The costs of the whole period associated with the absolute nominal_value of the flow.

which made me think that I have to pass the overall value for fixed costs e.g. for an absolute value of `P_el=90 MW` with costs of `1000 EUR/MW` the fixed costs should be set to `90,000 EUR`.

This made me skeptical as I thought (and only find it logically) to provide specific costs per unit e.g. `1000 EUR/MW` in this case.

Actually, luckily this is also [how it is implemented currently](https://github.com/oemof/oemof/blob/dev/oemof/solph/blocks.py#L509) and it was just not documented clearly.

@oemof/oemof-solph : Do you also think that the description is misleading?

Then, I would adapt it!

birgits:
When I first read the explanation for fixed_costs I thought it should be the 1000 EUR/MW but reading how you understood it made me question it as well. So I guess it should be clarified. I can't think of anything good though. Maybe we could explain it using an example.

ckaldemeyer:
> When I first read the explanation for fixed_costs I thought it should be the 1000 EUR/MW but reading how you understood it made me question it as well. So I guess it should be clarified. I can't think of anything good though. Maybe we could explain it using an example.

What do you think of
> fixed_costs (numeric) – The costs associated with one unit of the absolute nominal_value of the flow for the whole period.

or

> fixed_costs (numeric) – The specific costs for one unit of the absolute nominal_value of the flow for the whole period.

?

ckaldemeyer:
Actually ` absolute nominal_value` is redundant for me. The `absolute` could be removed for me.

ckaldemeyer:
We could also be more specific
> fixed_costs (numeric) – The costs associated with one unit of the absolute nominal_value of the flow for the whole period e.g. fixed OPEX.

ckaldemeyer:
But maybe being too specific could also be misleading..

birgits:
I like your last suggestion best!

ckaldemeyer:
> I like your last suggestion best!

:+1:

ckaldemeyer:
I'll do it after my holiday..

simnh:
What about this one?

uvchik:
[...]

PR #394, opened 2017-12-06
Update documentation according to issue #334

SabineHaas:
Could you close the issue please if you think it is solved?


ckaldemeyer:
@oemof/oemof-solph : Didn't we have a discussion about removing fixed costs!?

I couldn't find it but depending on it we could merge/omit this PR.

uvchik:
See #407.

PR #396, opened 2017-12-06
Add investment limit

simnh:
this adds an investment limit constraint. It should work, maybe do the following: 

* [x] Check functionality 
* [x] Add some documentation 
* [x] Add the functionality to an example (like the sdewes examples of v0.2)


simnh:
>  You have fixed_cost for existing components as well and the fixed_costs have not influence on the investment process.

First part: yes, second part: No 

We can discuss the 'No', but implemented is a "No", i.e. look here 

https://github.com/oemof/oemof/blob/dev/oemof/solph/blocks.py line 441 ff

That way, what is true: fixed_cost do not change anything for non-investment model except the absolution objective function value, but not the solution..

uvchik:
> second part: No

You are right 😄.

> We can discuss the 'No', but implemented is a "No", i.e. look here

I would leave it like this.

But still I would not add fixed_costs to the investment limit. Making an investment model with fixed_costs will be "dangerous". Otherwise you need shutdown cost as well and a possibility to save fixed costs if you e.g. shut down a power plant. I guess this is becoming a long discussion 😏

But if you want to implement it like this you should point that out in the docstring or the name. At least I would not expect the fixed_cost to be part of it.

mloenneberga:
I thought about the investment limit constraint again, after being through your conversation. I totally get your point about the investment limit having an influence on fixed cost and investment cost. Though, at least for my application, i'm interested in a constraint that limitis the initial capital. so, in that particular case, fixed cost should be left out. Do you have thoughts about example cases where both, fixed costs and investment cost limits are of interest?

simnh:
I think that depends on the definition of fixed costs, but if they are dependent of the capacity of a unit, then for me it just makes sense to include these, as they have to be earned every year (even if the unit does not run a single hour)....no?

uvchik:
You still have the chance to put your fixed costs on top of your epc if you like.

simnh:
I know, but then i would would to remove fixed costs completely. in investment you can add it there. in dispatch you can add it afterwards...so whats the purpose? (BTW: i totally like that, because in my view objective values should not be the point of interest anyway, but thats another discussion)

uvchik:
> I think that depends on the definition of fixed costs, but if they are dependent of the capacity of a unit, then for me it just makes sense to include these, as they have to be earned every year (even if the unit does not run a single hour)....no?

You are right but you will get a discrepancy because you do not have the chance to replace an existing power plant. It is a difficult discussion and I really do see that both ways have their arguments but the main point is that you have the chance to put them on top the epc but not the other way round.

uvchik:
> I know, but then i would would to remove fixed costs completely. in investment you can add it there. in dispatch you can add it afterwards...so whats the purpose? (BTW: i totally like that, because in my view objective values should not be the point of interest anyway, but thats another discussion)

I totally agree in that point. But there are people (at least were people) who like to use the objective function as an outcome and therefore we decided to add fixed_costs. I remember the discussion but for me it was okay to add them, I will not use them anyway.

simnh:
> the chance to replace an existing power plant
I do not understand that..how can we replace a existing plant within our implementation? Can you give an example (But I agree that its a long discussion)....

I vote for removing fixed costs...maybe split costs in the Investment class into capex and fixed_costs or something like that, but as we have epc...this does not matter anyway.

uvchik:
> BTW: i totally like that,

Even though I agree, I will not start the discussion... 😄

simnh:
yes but with our ethical mission of removing the objective-function-value-fetish of energy system modellers, I guess removing fixed costs would be great to avoid "Tatbestand: Fahrlässige Optimierung" ;)

uvchik:
:wave::wave:

ckaldemeyer:
> yes but with our ethical mission of removing the objective-function-value-fetish of energy system modellers, I guess removing fixed costs would be great. also to prevent "Tatbestand: Fahrlässige Optimierung" ;)

:+1:

simnh:
Yeah, but whats the decision here, can we delete fixed costs from all objective expression and only keep the investment_costs for investment models? I would then implement this and add entry in whats new.

ckaldemeyer:
> Yeah, but whats the decision here, can we delete fixed costs from all objective expression and only keep the investment_costs for investment models? I would then implement this and add entry in whats new.

I would vote for yes

simnh:
and then rename variable_costs to costs ;)

simnh:
Ok I removed the fixed caused from solph, now examples need adaptation. 

But what is worse I saw an error (I think): variable costs are not added for investment models

https://github.com/oemof/oemof/blob/dev/oemof/solph/blocks.pyL441-L445

And as the Flow-block is not called, there are no var costs in investment models

uvchik:
Even though I endorse the content I think this is not a very transparent process.

In my opinion we cannot remove fixed_costs in a PR called 'Add investment limit'. We should add a new PR/issue and give other developers the chance to object at least until the end of the week.

simnh:
You are right, my fault, but I think we can just inform them here by mentioning this, as it only is for @oemof/oemof-solph

uvchik:
The InvestmentFlow is just an add-on. First the FlowBlock is called for all flows and than the InvestmentFlowBlock is called for InvestmentFlows. At least this is how it should work if it doesn't work like this there is a bug. The best thing is to check the lp-file. 

I do not have time because just realised another bug caused by the lopf module I have to fix.

simnh:
By the way maybe there was not a bug, but this is due to the fact that Flow and InvestmentFlow are not two distinct groups (which I would like to change ->  #405

uvchik:
If I remember it right this was caused by the fact that the Flows are initiated in the Model class. We found it easier that way at that time.

simnh:
mh ok, however I think it should be distinct...I will adapt the tests for the fixed costs removal. I think we should definitely do it and users will not miss any functionality..

uvchik:
I opened a new branch and PR (#407) for the "fixed_costs" topic and reverted the associated commits.

ckaldemeyer:
I don't have time for reviews this week... Sorry :(

simnh:
Can we merge this?!

uvchik:
I added a test could you have a look -> [lp-file](https://github.com/oemof/oemof/pull/396/commits/baa0e15d44ac3c357a17bc857986e600341fb42d#diff-cba63a907c7bc08da09d8fe72e04abd5)

simnh:
looks great, I was just working on my first test :-( :)

uvchik:
Sorry, you should have mentioned that... I just did not want to annoy again 😄

uvchik:
@gnn said that there is an API in pyomo to get the existing constraints. That could help making constraint test faster by comparing just the important constraints and not the whole file...

simnh:
Ok, but we can merge now, right?

uvchik:
> @gnn said that there is an API in pyomo to get the existing constraints. That could help making constraint test faster by comparing just the important constraints and not the whole file...

This is just a "nice to have" and if you as an pyomo expert have an Idea how to do it we could open a not-urgent pull request.

> Ok, but we can merge now, right?

Yes

PR #399, opened 2017-12-08
Mini guide for contribution

c-moeller:
I would also like to add something like this on the first page of oemof.org. 
What do you think? Who can I ask for proofreading?


uvchik:
I fine with it.

I know it is not your topic but I think "Got further questions regarding oemof?" needs some clarification. In which case should I open an issue and in which case should I open a thread in the forum.

uvchik:
Thank you for your Off-Topic-Contribution 😄

PR #401, opened 2017-12-08
Add Link component and block

simnh:
I am not sure, but I think  / hope this functionality is not included in the transformer already...as for the lopf there is an link example an the branch with the same name in the example repository

### DIFF
--- a/oemof/solph/custom.py
+++ b/oemof/solph/custom.py
@@ -115,6 +115,16 @@ class ElectricalLineBlock(SimpleBlock):
             voltage_angle(i(n), t) - volatage_angle(o(n), t), \\
             \forall t \\in \\textrm{TIMESTEPS}, \\
             \forall n \\in \\textrm{ELECTRICAL\_LINES}.
+
+    TODO: Add equate constraint of flows
+
+    **The following variable are created:**
+
+    TODO: Add voltage angle variable
+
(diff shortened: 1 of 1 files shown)


uvchik:
No, it is not implemented. By now it is only possible to get this behaviour by adding two transformers (one for each direction). With the new Flows (min=-1) it is possible possible to define a bidirectional transformer but this would lead into nonsense because the conversion_factor will be `conversion_factor` for one direction and `1/conversion_factor` for the other direction. It is a good way to gain energy 😏

I hope people will understand this once they noticed that flows can be bidirectional.

simnh:
yeah I know, i was thinking about: 'How can I add a component with one flow (actually two) that has losses for the one direction and for the other...but that does only work with additional binary variables...I think the link component can be a good basis for a facade with investment, so that you do not have to link the investment variables...

however, I guess this already may simplify modelling

uvchik:
I already thought a hundred times about powerline facade. Just define two buses and an efficiency and in the background two transformers were created. The only thing is that I do not know how people will find their results. At least I haven't sufficiently thought about it.

simnh:
I know this is annoying, but we move slowly. I ll merge this...at least now you only need one component instead of two for many problems where you have two transformers with one in one output each.

uvchik:
> I know this is annoying, but we move slowly.

That's not what I meant to say 😄 Thank you for implementing 🎉

PR #403, opened 2017-12-10
Add sos2 costs 

simnh:
Hi I transferred the old sos stuff to this branch. I should work, but only for solvers that can handle sos-type constraints (e.g. gurobi). Right now the implementation works if you provide a list of tuples with base points to the investment.ep_costs attribute. (It does not work for storages so far,)

The same code would probably be needed for all nl-approx. costs (e.g. storage investment). Therefore it might be good to find a better place to drop the code. 

* [ ] Check functionality and interference with existing implementations
* [ ] Adapt documentation 
* [ ] Find the right location for the code
* [ ] Add an example
* [ ] Decide on API for investment
* [ ] Add check (for current implementation check if either scalar or list of tuples is passed) 

I will delete the other branch, hopefully this is a good starting point for you @jnnr !

### DIFF
--- a/oemof/solph/blocks.py
+++ b/oemof/solph/blocks.py
@@ -8,7 +8,8 @@ __copyright__ = "oemof developer group"
 __license__ = "GPLv3"
 
 from pyomo.core import (Var, Set, Constraint, BuildAction, Expression,
-                        NonNegativeReals, Binary, NonNegativeIntegers)
+                        NonNegativeReals, Binary, NonNegativeIntegers,
+                        SOSConstraint)
 from pyomo.core.base.block import SimpleBlock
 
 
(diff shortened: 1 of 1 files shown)


jnnr:
Hey simnh! Sorry for this late reply. And thanks for your code and this well structured introduction. I will dive into it as soon as I can find some time. As soon as I got something, I will let you know.

uvchik:
@jnnr Is it still on your list?

p-snft:
This is far from being merged. So, I released the requested reviewers and will postpone the feature request.

uvchik:
@jnnr Is it still on your list?

jnnr:
Thanks for asking, @uvchik. To be honest, I will not have time for this in the near future. To answer your question: This is rather not on my list at the moment.

uvchik:
Everybody is free to reopen it or use the code for further development. By now it is only possible to define an offset for the invest costs (see #636).

PR #405, opened 2017-12-12
Add gradient constraints to investment models

simnh:
I had to change the grouping (I thought it was already this way, that Flow and InvestmentFlow block are two distinct groups, but they weren't) I would like to have it this way, because there are no shared constraints anyway...

* [x] Add equations 
* [x] Check changes due to grouping
* [ ] Check timeincrement functionality


simnh:
I could add a test, there are still some problems with the grouping but I ll try to fix this...

uvchik:
A test would be great to keep the system robust.

It seems to me, that it does not break backwards compatibility. 

Is this feature urgent or shall we set the milestone to v0.2.1?

simnh:
Actually I think it'll break the API, therefore I woul like to add it to 0.2, as I will require to set the investment attribute for storages with investment from now on...which is, I think, straightforward anyway. I still try to fix some test...

uvchik:
You do not like it to add the investment object to the flows automatically? 
For me it is annoying to add three Investment objects. Furthermore you get into trouble if users add different attributes to them. They have to be empty.

simnh:
I causes trouble, because the Flows get grouped twice (once as flow and once as investment flow). I know it's annoying, but as I said, I think it's straightforward, and me might rather add facades as the storage is kind of hackisch and might at some point get changed anyway, no?

simnh:
Somehow the Flow-Block is called (or rather it's objective-functions expressions) but I don't know why, because actually this block should not be called as there are only investment flows that should end up somewhere else :- /

uvchik:
I thought about it and actually I do not care adding the Investment objects manually or not. Both ways have advantages and disadvantages.

simnh:
Ok, but can you figure out why there is this error? I quite don't get how the model is constructed within the constraint tests... :-/

simnh:
Puh...have it! I think this one only misses some polishing regarding the gradient-timeincrement problem

ckaldemeyer:
> Puh...have it! I think this one only misses some polishing regarding the gradient-timeincrement problem

Did the e-mail from today clarify things!?

simnh:
I did now: 

* Multiply upper bound by timeincrement
* Constraint simply calculates the gradient value within the given bounds
* In the objective timeincrement is added
* I also made the gradient upper bound timedependent of the nominal max[t] value

I hope this is correct, but I really cannot get my head around that...

If it is correct correct, I will add the correct documentation (math for constraints etc) and this branch is ready to merge (after resolving conflicts)

ckaldemeyer:
> I did now:
> 
> Multiply upper bound of gradients by timeincrement
> Constraint simply calculates the gradient value within the given bounds
> In the objective timeincrement is added
> I also made the gradient upper bound timedependent of the nominal max[t] value
> I hope this is correct, but I really cannot get my head around that...
> 
> If it is correct correct, I will add the correct documentation (math for constraints etc) and this branch is ready to merge (after resolving conflicts)

Thanks a lot for that. I think a testing example would clarify if it works correctly. I could help with the test if you want!

ckaldemeyer:
> I did now:
> 
> Multiply upper bound of gradients by timeincrement
> Constraint simply calculates the gradient value within the given bounds
> In the objective timeincrement is added
> I also made the gradient upper bound timedependent of the nominal max[t] value
> I hope this is correct, but I really cannot get my head around that...
> 
> If it is correct correct, I will add the correct documentation (math for constraints etc) and this branch is ready to merge (after resolving conflicts)
> 
> Thanks a lot for that. I think a testing example would clarify if it works correctly. I could help with the test if you want!

If you drop some example lines here, I can also create it

simnh:
An example would be great, but I do not have the time to add this at the moment. We can also extend an existing one...

uvchik:
Is it still valid? It is not in line with the actual code.

simnh:
I am not sure...jsut postpone if nobody needs this feature

uvchik:
Parts of this PR seems to be bug fixes rather than a feature. Is it right?

For example line [140ff](https://github.com/oemof/oemof/pull/405/files#diff-5ccc7aaea84212594325204da4fa2ee9R140).

uvchik:
Still some action here?

uvchik:
I brought the code in line with the v0.4.x structure, so it is possible to finish this PR soon.

I added a basic test, but we have to check it very accurately before we merge it. Therefore I think v0.4.1 is a realistic milestone.

uvchik:
Still constraint tests are missing!

simnh:
I think this can be closed - reopen / create new again, once required / needed.

p-snft:
Okay, I'll close this.

PR #407, opened 2017-12-12
Remove fixed_costs attribute from Flow class

uvchik:
As this attribute is not necessarily needed but might be confusing in investment models we would like to remove it. This came up in PR #396 but I think it is cleaner to create a separate PR.

@oemof/oemof-solph Any objections?

### DIFF
--- a/oemof/solph/components.py
+++ b/oemof/solph/components.py
@@ -93,8 +93,7 @@ class GenericStorage(network.Transformer):
     ...     nominal_input_capacity_ratio=1/6,
     ...     nominal_output_capacity_ratio=1/6,
     ...     inflow_conversion_factor=0.9,
-    ...     outflow_conversion_factor=0.93,
-    ...     fixed_costs=35)
+    ...     outflow_conversion_factor=0.93)
 
     >>> my_investment_storage = solph.components.GenericStorage(
     ...     label='storage',
@@ -126,7 +125,6 @@ class GenericStorage(network.Transformer):
                 'outflow_conversion_factor', 1))
         self.capacity_max = solph_sequence(kwargs.get('capacity_max', 1))
         self.capacity_min = solph_sequence(kwargs.get('capacity_min', 0))
-        self.fixed_costs = kwargs.get('fixed_costs')
         self.investment = kwargs.get('investment')
 
         # General error messages
@@ -208,11 +206,6 @@ class GenericStorageBlock(SimpleBlock):
 
     **The following parts of the objective function are created:**
 
-    If :attr:`fixed_costs` is set by the user:
-        .. math:: \sum_n nominal\_capacity(n, t) \cdot fixed\_costs(n)
-
-    The fixed costs expression can be accessed by `om.Storage.fixed_costs`
-    and their value after optimization by: `om.Storage.fixed_costs()`.
     """
 
     CONSTRAINT_GROUP = True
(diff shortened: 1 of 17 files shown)


uvchik:
I would raise an error if fixed_costs are still used. Because of  `**kwargs` additional keyword arguments are allowed and people may not notice the removal of this attribute.

uvchik:
> But I do not understand the hurry.

According to the semantic versioning rules we are only allowed to change the API in mayor releases. In our cases this is v0.2, v0.3... That's why we have some last minute discussions about API changes right now.

> Do they need to have this issue in mind?

I think removing fixed_cost takes about one minute so that should not be the point.

oemof/solph/components.py
```diff
@@ -820,25 +796,13 @@ def _Q_min_res_rule(block, n, t):
...
-        for n in self.GENERICCHPS:
-            if n.fixed_costs is not None:
-                P_max = [list(n.electrical_output.values())[0].P_max_woDH[t]
-                         for t in m.TIMESTEPS]
-                fixed_costs += max(P_max) * n.fixed_costs
-
-        self.fixed_costs = Expression(expr=fixed_costs)
-
-        return fixed_costs
+        return 0
```
ckaldemeyer: If you agree, we could also remove the objective expression completely as it holds nothing!?

ckaldemeyer: It's your decision @uvchik. For the rest, I am fine. Thanks!

simnh: @ckaldemeyer  you can remove the 'empty' method, just make sure that no error occurs, as for all blocks the method is automatically called

ckaldemeyer: > @ckaldemeyer you can remove the 'empty' method, just make sure that no error occurs, as for all blocks the method is automatically called

Let's wait what @uvchik says. But I can do it!

uvchik:
I just moved this stuff from one branch to another to make it more transparent. I actually did not start these changes.

In the end we have to search for all occurrences of `fixed_costs` anywhere in the code and the documentation and remove it. If all tests work and our test are robust we will not have any problems.

I can help but do not feel responsible for the whole PR.

uvchik:
Because of the `kwargs` nothing will happen if people define fixed_costs. They will be silently ignored by solph. So people may not notice that we removed `fixed_cost`, keep on using them and wonder why the results are different.

@oemof/oemof-developer-group What shall we do about it? Warning? Error?

simnh:
A created a branch in oemof_examples repository with the same name where I removed all the fixed costs. Somebody might review this on the pull request:

https://github.com/oemof/oemof_examples/pull/9

simnh:
I wonder why all test pass... :-/, should the lp-files not be broke?

uvchik:
You forgot one comma in line 56 in test_components.py.

If the test is not about the fixed_costs attribute it should not fail if you remove it. So everything is fine. The constraint test failed but I fixed them.

simnh:
But you did not push this, right? :)

uvchik:
Travis-Ci test will merge and then test. As I added test for the meta results the objective function was tested and the objective function will be different if fixed_costs are removed.

simnh:
Thanks @uvchik! So what about can we merge?

uvchik:
Could we first answer the question [above](https://github.com/oemof/oemof/pull/407#issuecomment-351449967).

Because of the `kwargs` nothing...

simnh:
I added an attribute error...hope this is ok

simnh:
For me a depreciated warning is fine (and if the program stops). If you want to change it feel free, otherwise I would merge this one?

uvchik:
I just wonder because it is not in line with common practise.

Your first variant (32c1ebb) was fine I just wonder why you changed it because now it is not in line with common practise.

> Warning messages are typically issued in situations where it is useful to alert the user of some condition in a program, where that condition (normally) doesn’t warrant raising an exception and terminating the program.[1]

[1] https://docs.python.org/3/library/warnings.html#module-warnings

simnh:
Ok the lets revert it...

uvchik:
I will write a small test. Wait with you merge....

uvchik:
Okay, done.

uvchik:
Approve (but I am not a reviewer) :smile:

The error message could contain an advise but we can merge it and anybody who wants could change it.

uvchik:
@simnh According to yesterdays talk: 851a4d7, simple but effective.

PR #409, opened 2017-12-13
Adaptions of the docu to the new version of the outputlib.

jnnr:


ckaldemeyer:
Thanks for your effort!

Should we add some remarks on the structure of the result dictionary (my_results[(n1,n2)/(n1,None)]['sequences'/'scalars']?

jnnr:
Thanks for reviewing! Indeed we should add a more detailed description.

I had to get into working with outputlib myself. I try to learn by playing with the examples. Am I right that you have to give a name to the node when you create it to be able to pick it later in the results dictionary later by [(n1, n2)] or [(n1, None)] ? In the examples, not all nodes are assigned variable names.

uvchik:
Both is possible. You can use the name (label) or the object itself.

By default it is the object but @ckaldemeyer wrote a transformation function that converts the objects in the keys to their labels.

jnnr:
I checked it again. Indeed both is possible, but only in views.node(). Here you can either do views.node(results, 'electricity') 
or 
views.node(results, b_el) 
for a node that is defined as
bel = solph.Bus(label='electricity')

Working with results dict directly it does only work with the variable names. Does that mean that one usually would not use the results dictionary provided by processing.results(om) directly, but only by requesting views of certain nodes?

henhuy:
You can use strings to access results as well - therefore, you have to convert the keys of result dict to strings via `views.convert_keys_to_strings(result)` first.
The same happens in `views.node` automatically if you insert strings.
But attention! Output of `Views.node` differs on input! If you enter strings, all outputs of node results will be strings as well (no `Node` objects anymore!) - so you are losing information thereby...

uvchik:
To access the objects you can also use `groups` of the EnergySystem.

```python
>>> es = solph.EnergySystem(....)
[....]
>>> storage = es.groups['storage_label']
>>> type(storage)
<class 'oemof.solph.components.GenericStorage'>
```

ckaldemeyer:
> You can use strings to access results as well - therefore, you have to convert the keys of result dict to strings via views.convert_keys_to_strings(result) first.
> The same happens in views.node automatically if you insert strings.
> But attention! Output of Views.node differs on input! If you enter strings, all outputs of node results will be strings as well (no Node objects anymore!) - so you are losing information thereby...

Well, this was just how everything started during the redisign of the outputlib. If you have the feeling that a different functionality (e.g. always converting to string keys, ...) makes more sense, we can still change it.

jnnr:
I adapted the documentation according to your comments. It seems that the most handy way to use the outputlib as it is now is using views.node() to access data belonging to nodes or groupings. 

I found the documentation of the groupings rather confusing. Searching the examples, I did not find one that uses the groupings. I will play around with this to learn how it works as soon I have the time.

uvchik:
> I adapted the documentation according to your comments. It seems that the most handy way to use the outputlib as it is now is using views.node() to access data belonging to nodes or groupings.

I would not say that. The `views.node()` function is very convenient but I do need the direct API as well e.g. if you want to compare a special flow between regions. In that case the `views.node()` function does not help. Therefore I would describe both ways.

> I found the documentation of the groupings rather confusing. Searching the examples, I did not find one that uses the groupings. I will play around with this to learn how it works as soon I have the time.

One easy thing is that each Node object is a group itself identified by its label. So you can get any Node by using the group method `es.group['label']`. I think this is the most important functionality for users.

uvchik:
I think processing is a better place for the conversion function (keys to labels) #417 .

jnnr:
I expanded and reformulated the explanation again after talking with Uwe. I hope things are clearer now?

jnnr:
@cord @uvchik: Are you ok with this? With your approval, I would merge soon.

uvchik:
I think you can just merge. We still may fix things later.

jnnr:
Ok, thanks!

ckaldemeyer:
ohh.. to late ;)

PR #410, opened 2017-12-13
Improve developer welcome

jnnr:
I added a new section on developing in README.txt and extended developing_oemof.rst:
Now there is a welcome sentence to tell about the possibilities to become a developer.
I also expanded section on issues and put it directly under pull requests as those are the two ways to communicate.

Apart from this, I wonder if the section on pull requests still describes the workflow how we understand it now. We could add a bullet point on how to invite reviewers (how many, whom,..) Also, is the section on tests still up to date?

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -49,11 +49,16 @@ You can find a repository with all currently available examples for oemof on git
 
 You are welcome to contribute your own examples via a `pull request <https://github.com/oemof/examples/pulls>`_ or by sending us an e-mail (see `here <https://oemof.org/contact/>`_ for contact information).
 
-Got further questions regarding oemof? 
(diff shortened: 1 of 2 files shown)


mloenneberga:
I like it! It might be nice to add a short note on developer meetings, since i think that these are a good way of getting in touch with the community if you want to start  contributing.

c-moeller:
I like your changes, too!

Because I added some contribution notes, too, there now might be some double information (in documentation and readme), such as here:
http://oemof.readthedocs.io/en/latest/about_oemof.html#why-should-i-contribute

Maybe we should adapt that some day, or maybe it's even ok like that.

jnnr:
Thanks for your opinions!

PR #411, opened 2017-12-14
Add BaseModel to solph 

simnh:
Hey I like to restructure solph models internally, to add a multi-period model at some point, which I won't merge with the existing stuff. I think this will add flexibility...

Does not change anything really for anyone except core-solph-developers ;)

I really like to merge this once for v0.2 ;)

@gnn : Would be very interested in your opinion on my code!

### DIFF
--- a/oemof/outputlib/graph_tools.py
+++ b/oemof/outputlib/graph_tools.py
@@ -97,7 +97,7 @@ def graph(energy_system, optimization_model=None, edge_labels=True,
     ...                                                variable_costs=40)},
     ...                            conversion_factors={b_el: 0.5})
     >>> es.add(b_gas, b_el, demand_el, pp_gas)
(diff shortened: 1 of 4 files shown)


uvchik:
> Does not change anything really for anyone except core-solph-developers ;)

This announcement was misleading, because you silently broke the API.

Please be clearer if you rename a central attribute. I took me a while to find the error 😞

henhuy:
Why did the tests pass if API was broken?

uvchik:
> Why did the tests pass if API was broken?

Good point, I will add a regression test.

uvchik:
> Why did the tests pass if API was broken?

I do not have to add tests because there are already some test regarding this issue. They were fixed.

Only examples/apps failed afterwards.

PR #414, opened 2017-12-14
Features/doku api changes

c-moeller:
I updated the documentation parts oemof-network and oemof-solph. Did I miss something?

In oemof-solph I only did some revisions which are not related to the api changes, because I didn't see any api description.


c-moeller:
Some remarks:
- The Flow class is not described in oemof-network. Is this on purpose? Or did we just forget in the past?
- The image in oemof-solph shows a LinearTransformer. I'm going to adapt this.
- I need to add the link to the custom components description after I merged the corresponding PR.

doc/oemof_network.rst
```diff
@@ -6,7 +6,13 @@ oemof-network
...
+A *node* is either a *bus* or a *component*. A bus is always connected with one or several components. Likewise components are always connected with one or several buses. Based on their characteristics components are divided into several sub types.
+
+*Transformers* have input and output, e.g. a gas turbine takes from a bus of type 'gas' and feeds into a bus of type 'electricity'. With additional information like parameters and transfer functions input and output can be specified. Using the example of a gas turbine, the resource consumption (input) is related to the provided end energy (output) by means of an conversion factor. Components of type *transformer* can also be used to model transmission lines.
+
+A *sink* has only an input but no output. With *sink* consumers like households can be modeled. But also for modelling excess energy you wouls use a *sink*.
```
uvchik: wouls != would

doc/oemof_network.rst
```diff
@@ -6,7 +6,13 @@ oemof-network
...
+A *node* is either a *bus* or a *component*. A bus is always connected with one or several components. Likewise components are always connected with one or several buses. Based on their characteristics components are divided into several sub types.
+
+*Transformers* have input and output, e.g. a gas turbine takes from a bus of type 'gas' and feeds into a bus of type 'electricity'. With additional information like parameters and transfer functions input and output can be specified. Using the example of a gas turbine, the resource consumption (input) is related to the provided end energy (output) by means of an conversion factor. Components of type *transformer* can also be used to model transmission lines.
```
uvchik: ...*Transformers* have any number of inputs and outputs... 

e.g. CHP one input two outputs

doc/oemof_solph.rst
```diff
@@ -46,17 +46,16 @@ Now you can start to add the components of the network.
...
 
-If you have defined an instance of the EnergySystem class all components you define will automatically be added to your EnergySystem.
+After defining an instance of the EnergySystem class you have to add all components you define in the following to your EnergySystem.
 
 Basically, there are four types of Nodes and every node has to be connected with one or more buses. The connection between a component and a bus is the flow.
 
  * Sink (one input, no output)
  * Source (one output, no input)
- * LinearTransformer (one input, n outputs)
+ * LinearTransformer (n inputs, n outputs)
```
uvchik: Transformer not LinearTransformer

doc/oemof_solph.rst
```diff
@@ -46,17 +46,16 @@ Now you can start to add the components of the network.
 Add your components to the energy system
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-If you have defined an instance of the EnergySystem class all components you define will automatically be added to your EnergySystem.
+After defining an instance of the EnergySystem class you have to add all components you define in the following to your EnergySystem.
```
uvchik: Components is right in trivial language but in our case we have to say nodes. Nodes can be buses or components and components can be sources, sinks,.... see https://github.com/oemof/oemof/blob/dev/oemof/network.py#L248

doc/oemof_solph.rst
```diff
@@ -46,17 +46,16 @@ Now you can start to add the components of the network.
...
 
 Basically, there are four types of Nodes and every node has to be connected with one or more buses. The connection between a component and a bus is the flow.
 
  * Sink (one input, no output)
  * Source (one output, no input)
- * LinearTransformer (one input, n outputs)
+ * LinearTransformer (n inputs, n outputs)
  * Storage (one input, one output)
```
uvchik: The storage is not a network class.

uvchik: Sorry, I did not noticed that we are already in the solph documentation.

In solph we do have three groups:

1. Components directly inherited from the network class (pure graph concept). They can be found in the network module and the blocks in the blocks module (We should change that because it is different to anything else).
2. Proofed solph components which can be found in the components module.
3. Experimental or very specific stuff in the custom module.

doc/oemof_solph.rst
```diff
@@ -256,14 +260,14 @@ Furthermore, an efficiency for loading, unloading and a capacity loss per time i
 
 .. code-block:: python
 
-    solph.Storage(
+    my_energysystem.add(solph.Storage(
```
uvchik: `solph.components.GenericStorage()`

c-moeller:
Thanks for your revisons! Will include them on Monday.

doc/oemof_solph.rst
c-moeller: Shall I add your explanation about the three groups to the docu?

uvchik: Why not, it helps to understand the structure.

jnnr:
Should we not point out more clearly that oemof-network is part of oemofs core and contains the base classes that are used in solph? It took me some time in the beginning that I do not have to directly use oemof-network to build an energy system. We could adapt the first sentence to make that clear. Also, the example confuses me a lot. Apart from it being not very beautiful it confuses me because it suggests oemof-network being something I would want to get my hands in which is not true for beginners, I guess. Or am I wrong?

uvchik:
> Apart from it being not very beautiful it confuses me because it suggests oemof-network being something I would want to get my hands in which is not true for beginners, I guess. Or am I wrong?

No, you are not. At the moment it is completely useless to define an energy system on the network level. But it was planed differently for the v0.1 release that's why this section was added. I think it is a good idea to point that out. All components can be found in solph, too. Most of the users will have no contact to the network level.

You may want to inherit from network components if you want to create new components, but that's all at the moment.

@oemof/oemof-solph Or do I miss something?

c-moeller:
> Also, the example confuses me a lot.

In my opinion we can remove the example, because we have a picture explaining nearly the same in the oemof-solph section.

c-moeller:
I removed the example from the network documentation and did some other small changes.
@uvchik, @jnnr can I merge?

jnnr:
Yes, that looks better. In my opinion, this can be merged now. Thank you!

uvchik:
I talked to @c-moeller. After fixing some discrepancies in the `Bus` chapter you may merge.

PR #416, opened 2017-12-15
Add Node parameters to result object

ckaldemeyer:
This feature aims at adding all node parameters to a data structure having the same structure as the result object. Doing this, both should be easily mergeable afterwards..


ckaldemeyer:
@henhuy : I have started with the idea but not finished yet. But you could have a look at the general idea and me/we can continue next week ;-)

ckaldemeyer:
@henhuy: `separate_node_attrs(om)` is not finished yet and we could also put redundant code into other functions e.g. `get_iterable_attrs` or similar. It's just a sketch..

henhuy:
Hey @ckaldemeyer,
thanks for your work! Finally, I found some time to look at your code.
I've got some questions/remarks on it:
1. I think as the functions `separate_node_attrs` and `separate_flow_attrs` are almost the same they should be transferred into one generic function.
2. I would change the index for node param_results from `(node, None)` to simple `node` as this is very intuitive (I think the `(node, None)` solution is only necessary in case of mixed tuples having both `(node, node)` and `(node, None)` occurrences?!)
3. I like your flattening of dicts. Nevertheless, this leads to "sequences" which consist of scalars. For example in case of gradient `positive_gradient = {'ub': [None], 'costs': 0}` this leads to the "sequence" `positive_gradient_costs = 0` which is only an integer. Maybe we could think about this one more time?
4. Eventually, I would neglect Scalars or Sequences containing None or [None] respectively, as they don't really give additional information. What do you say?

If you are fine with some of these suggestions, I can implement them...

henhuy:
To 4.: Maybe, this could be a users choice. This means, adding a parameter `neglect_none=False` or similar!

henhuy:
I implemented my above suggestions.
Following features/fixes are integrated:
1. `None` values can be excluded
2. Created sequences are double-checked for scalars; found scalars are moved to scalars (i.e. negative_gradient_costs = 1 are moved to scalar after flattening Sequence). **Not sure if this behaviour is wanted** - but on the other hand I found it strange to have scalars in sequences dict?!
3. `Investment` is unpacked (by recursive call of the function) and added with `investment_`- prefix
4. Keys can also be converted to strings

Please check and confirm PR if everything is okay :)

henhuy:
hmm. Python3.4 check fails due to circular import, which is partially allowed in new versions.
I'll find a workaround...

uvchik:
This is not only a question of Python 3.4 but of oemof.

As I understand the structure the outputlib should not depend on the solph library.

@oemof/oemof-core, @oemof/oemof-outputlib Is this right?

uvchik:
@henhuy Thank you for the solution. But in the long run we have discuss the topic above. But not in this PR ;-)

ckaldemeyer:
@henhuy : I just went through the code and the test. Thanks for your effort!

> I implemented my above suggestions.
> Following features/fixes are integrated:
> 
>     None values can be excluded

:+1: 

> Created sequences are double-checked for scalars; found scalars are moved to scalars (i.e. negative_gradient_costs = 1 are moved to scalar after flattening Sequence). Not sure if this behaviour is wanted - but on the other hand I found it strange to have scalars in sequences dict?!

This was not wanted so I prefer you solution.

> Investment is unpacked (by recursive call of the function) and added with investment_- prefix

:+1: 

> Keys can also be converted to strings

:+1: 

When I look at `processing.param_results` I am struggling with the structure (which should be easily changeable):

    """
    Create a result dictionary containing node parameters.

    Results are written into a dictionary of pandas objects where
    a Series holds all scalar values and a dataframe all sequences for nodes
    and flows.
    The dictionary is keyed by the nodes e.g.
    `results['nodes'][idx]['scalars']`
    and flows e.g. `results['flows'][(n,n)]['sequences']`.
    """

My idea was to have the `param_results` 100 % compatible with the optimization results to make them easily mergeable e.g. by appending the param series to the opt. results series or the dataframes similarly.

This should work by putting flows params in `results[(n,n)]['sequences'/'scalars']` and node params `results[(n, None)]['sequences'/'scalars']` as it is in the regular result object.

@henhuy : What do you think?

ckaldemeyer:
> This is not only a question of Python 3.4 but of oemof.
> 
> As I understand the layer structure the outputlib should not depend on the solph library.
> 
> Both can inherit from the core but not from each other...I am confused.
> 
> @oemof/oemof-core, @oemof/oemof-outputlib May anybody help?

Yes, we somehow have to deal with this and getting some feedback would be nice. So I'll add @simnh and @gnn

ckaldemeyer:
> Yes, we somehow have to deal with this and getting some feedback would be nice. So I'll add @simnh and @gnn

Meaning some feedback from your side would be nice ;-)

henhuy:
Hey @ckaldemeyer  thanks for your review!

> This should work by putting flows params in `results[(n,n)]['sequences'/'scalars']` and node params `results[(n, None)]['sequences'/'scalars']` as it is in the regular result object.

Yeah, great - I just wanted to suggest the same thing, struggling over it myself.
Cool! I'll change it and then we are ready to merge, I think...

henhuy:
**FYI**:
The circular import is due to one single import of the `processing` module in `oemof.solph.models`. Within this module the `processing.results` function is used to return results by calling `BaseModel.results()`. I don't know if this function is important (often used) - just wanted to point out where "the circle" starts.

simnh:
I really like to see you going "crazy" about this. But I kind of lost track of all functionality. Maybe it would be good to provide one rather extensive example where the various ways to extract result are shown (extensive in terms of result processing and not components etc.)

For new users I think this would help a lot...

simnh:
I think BaseModel.results() could be removed, results are already stored in this model if it has been solved. We want to move this function where you can get result to the energysystem in the near future...
An hopefully at some point back to the components...but that a way to go...

simnh:
> As I understand the layer structure the outputlib should not depend on the solph library.

That's the way I think it (not sure if we can implement it right now but for the near future. Result structure should be in its basic form something in core (as e.g. there will always be flows, either exogenous or endogenous). 

However I think it this way (long term) the energy system will not have results but just a set of information that may be coming from anywhere but also from model results. so in this thinking models have results, energy systems don't. But we have to discuss this and I know this contradicts what I said above regarding the location of a `results()` method....

ckaldemeyer:
> I really like to see you going "crazy" about this. But I kind of lost track of all functionality. Maybe it would be good to provide one rather extensive example where the various ways to extract result are shown (extensive in terms of result processing and not components etc.)
> 
> For new users I think this would help a lot...

:+1: We could do this in the examples repo. But I don't have time.

> I think BaseModel.results() could be removed, results are already stored in this model if it has been solved. We want to move this function where you can get result to the energysystem in the near future...
> An hopefully at some point back to the components...but that a way to go...

I don't know who added the method but go ahead.

ckaldemeyer:
> Hey @ckaldemeyer thanks for your review!
> 
>     This should work by putting flows params in results[(n,n)]['sequences'/'scalars'] and node params results[(n, None)]['sequences'/'scalars'] as it is in the regular result object.
> 
> Yeah, great - I just wanted to suggest the same thing, struggling over it myself.
> Cool! I'll change it and then we are ready to merge, I think...

:+1:

ckaldemeyer:
> henhuy merged commit f7264ec into dev 3 hours ago

No offense but I did not review but only commented :smile: I would like to do some testing before as @simnh proposed and some minor polishing!

@henhuy : Would that be fine and we revert until then!?

henhuy:
Oh sorry! Yes, no problem.

ckaldemeyer:
> Oh sorry! Yes, no problem.

But it looks as it is still in `dev` :smile:

ckaldemeyer:
> But it looks as it is still in dev :smile:

Anyway, it doesn't hurt. We could also merge the `dev` into https://github.com/oemof/oemof/pull/433 and adapt changes from there as you contributions do not intefer with other functionalities.

henhuy:
Yes, I'm not sure how to revert. Simply pressing "Revert" or do I have to keep something in mind? (sorry - not quite familiar with the github stuff)

ckaldemeyer:
> Yes, I'm not sure how to revert. Simply pressing "Revert" or do I have to keep something in mind? (sorry - not quite familiar with the github stuff)

I guess the easiest thing would be to just merge dev into https://github.com/oemof/oemof/pull/433/files and work from there. If we generally agree that we want to have result_params it should be fine!

PR #417, opened 2017-12-18
move key-conversion-function from views to processing module

uvchik:
The conversion of the results keys belongs to processing rather than to views.

### DIFF
--- a/oemof/outputlib/processing.py
+++ b/oemof/outputlib/processing.py
@@ -144,6 +144,18 @@ def results(om):
     return result
 
 
+def convert_keys_to_strings(result):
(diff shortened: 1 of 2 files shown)


jnnr:
I agree. Converting the keys of the results dictionary from object names to labels is something generic. You may want to convert keys to strings without using the views module.

uvchik:
> I agree. Converting the keys of the results dictionary from object names to labels is something generic. You may want to convert keys to strings without using the views module.

I guess that is an implicit approval?

I will wait for an approval of @ckaldemeyer because he wrote this function.

PR #421, opened 2018-01-04
Correct desctribtion of pandas Dataframe

jnnr:
I corrected a detail in the docu "Using oemof". The pandas object we use is a pandas dataframe, not a MultiIndex dataframe.


uvchik:
As I understand it the overall structure is a dictionary. Only the time depended variables are stored in DataFrames within the dictionary structure.

PR #424, opened 2018-01-08
Features/add nonconvex offset component

ckaldemeyer:
This implements the simple case of (yet) a 1:1 Transformer that has a binary inflow with load ranges, etc. (compatible with gradients, ...) and one outflow. We are using it to model heat pumps, ...

I haven't provided a test yet but can write a quick one if you agree that we want this component in the code base.


ckaldemeyer:
And it includes a fix within the graph tools: https://github.com/oemof/oemof/pull/424/files#diff-4832c6033aedd30fc4bb618296f12add

Issue #418, opened 2017-12-18
Shall matplotlib stay a requirement of oemof?


uvchik:
At the moment the matplotlib is a requirement of oemof.

As we decided that the plotting functionality shall be moved to oemof-visio there is only one function that needs matplotlib: [graph](https://github.com/oemof/oemof/blob/dev/oemof/outputlib/graph_tools.py#L12).

The graph function also need networkx and pygraphviz, which are not requirements of oemof so it does not work out of the box any way. Furthermore the plotting functionality is just one feature of the graph function.

So, I would vote for the removal of the matplotlib as an requirement of oemof.

@oemof/oemof-solph What do you think?

uvchik:
@birgits knows a way to add optional requirements. Would that be an option?

uvchik:
@birgits As nobody else cares could you make a PR with an optional installation of the matplotlib?

uvchik:
Closed in favour of #425.

PR #425, opened 2018-01-08
remove matplotlib requirement and adapt graph tools

uvchik:
* matplotlib was removed from the requirements because plotting function was moved to oemof-visio
* do not throw warnings unless the function is used 
     - matplotlib warning only if `plot=True`
     - graph-tool warnings only if graph function is used
* Test import of each package separately and throw separate logging-error
* clean code
* integrate changes from #424 
* Related to #418


uvchik:
Sorry, I messed around a little bit because the tests work locally but not with Travis, so I had to push every little change and try again.

Now everything works but I think we should split this function.

One function that creates the network graph. One can make network analyses with this graph and I am not sure if the outputlib is the right place.

A second function that draws the graph. This function could be part of the outputlib or even oemof-visio because we decided that we do not want to have plotting function within oemof.

@ckaldemeyer @oemof/oemof-solph What do you think?

ckaldemeyer:
> Sorry, I messed around a little bit because the tests work locally but not with Travis, so I had to push every little change and try again.
> 
> Now everything works but I think we should split this function.
> 
> One function that creates the network graph. One can make network analyses with this graph and I am not sure if the outputlib is the right place. The tools module might be a better place.
> 
> A second function that draws the graph. This function could be part of the outputlib or even part of oemof-visio because we decided that we do not want to have plotting functions within oemof.
> 
> @ckaldemeyer @oemof/oemof-solph What do you think?

Thanks for your effort! I strongly support this.

uvchik:
After I talked to @ckaldemeyer I implemented the suggestions from above.

* The graph function is divided into `graph()` and `draw_graph()`.
* `draw_graph()` is deprecated and will be removed with the next major release (or with this release if we get a fast decision - see below)
* The `networkx` is now a requirement of oemof, while `matplotlib` was removed as a requirement.

Open questions:

* Is the outputlib a good place for a graph tool, if it can be used to locate unconnected nodes etc.? I think we should move it into the tools or the core section.
* Where shall we move the `draw_graph()` function? oemof-visio, examples....

A good place for the graph tools might be interesting to clearly differentiate the model part from the graph part.

@oemof/oemof-core / @oemof/oemof-solph 
If get a fast decision we can move it within the v0.2. release. Otherwise we will do it later.

ckaldemeyer:
> After I talked to @ckaldemeyer I implemented the suggestions from above.
> 
>     The graph function is divided into graph() and draw_graph().
>     draw_graph() is deprecated and will be removed with the next major release (or with this release if we get a fast decision - see below)
>     The networkx is now a requirement of oemof, while matplotlib was removed as a requirement.

Thanks!

> Is the outputlib a good place for a graph tool, if it can be used to locate unconnected nodes etc.? I think we should move it into the tools or the core section

I would also vote for tools.

> Where shall we move the draw_graph() function? oemof-visio, examples....

Depending on the usage frequency start with examples and move to visio if people need it more often.

> @oemof/oemof-core / @oemof/oemof-solph
> If get a fast decision we can move it within the v0.2. release. Otherwise we will do it later.

I vote for v.0.2!

uvchik:
> I vote for v.0.2!

Okay, let's wait for a response of @simnh.

simnh:
I am  fine, though this could also just be a function that is integrated in  the core energysystem as a method...but even therefore it can stay in the outputlib

simnh:
I think also it make sense to name it: create_nx_grahp or somthing, as the graph is already defined in a oemof logic...

ckaldemeyer:
> I am fine, though this could also just be a function that is integrated in the core energysystem as a method...but even therefore it can stay in the outputlib

We could easily use the function from `tools` within a method of the  `es` class first and transfer it to the class afterwards. What do you think?

ckaldemeyer:
> I am fine, though this could also just be a function that is integrated in the core energysystem as a method...but even therefore it can stay in the outputlib
> 
> We could easily use the function from tools within a method of the es class first and transfer it to the class afterwards. What do you think?

The graph creation would also work at core level and we could place it there and have one more example for our layer structure..

uvchik:
So be it....

PR #433, opened 2018-01-12
Features/add cost calculations

henhuy:
# Adds cost calculations to oemof
**Located in**: `oemof.tools.economics`
**Functionality**:
- Basic cost results dictionary can be calculated via function `oemof.tools.economics.cost_results`.
- Node flows can be categorized by flow type into "single", "input" & "output" flow via function `oemof.outputlib.views.get_flow_type` (see information below at "Share your knowledge")
- LCOE can be calculated for nodes

**Usage**:
Different costs can be calculated dynamically (and extended) by function `cost_results` - two default cost calculations are implemented (invest & variable_costs). Cost calculation is based on `Cost` variables which define which parameter key is multiplied with which flow key. For example invest costs are calculated by: 
`invest = param_results[nodes]['scalars']['invest'] * results[nodes]['scalars']['invest']`
and variable_costs by (sequences are first transformed into pandas.Series object by reading `data` attributes):
`variable_costs = pandas.Series(param_results[nodes]['sequences']['variable_costs'].data) * pandas.Series(results[nodes]['sequences']['variable_costs'].data)`
Once cost results dictionary is calculated, LCOE for nodes can be calculated.

## API:
- Cost results can be get by simply calling:
`cost_results = economics.cost_results(results, param_results)`
- LCOE are calculated via (here for node storage):
`storage_lcoe = economics.calculate_lcoe(storage, results, cost_results)`
- Cost calculations can be extended/changed by defining custom costs (here "real_invest_costs" are calculated, multiplying investment with real costs instead of ep_costs):
```
custom_costs = [
    economics.Costs(
        name='real_invest_costs',
        flow_key=economics.AttributeKey(
            dimension='scalars',
            name='invest'
        ),
        param_key=economics.AttributeKey(
            dimension='scalars',
            name='real_invest_costs'
        )
    )
]
custom_cost_results = economics.cost_results(
    results, param_results, custom_costs)
```

## Share your knowledge:
Function `get_flow_type` is needed to categorize node flows and to further distinguish between OPEX or fuel costs etc.

## Other comments and questions:
Which cost calculations are useful as well?


henhuy:
See `test_economics` for further API examples.

simnh:
Ok some thoughts on this: 

- In my view in `oemof.tools.economics` (which might at some point be `oemof.economics` ) should only hold standard calculations (like the once of NREL) which are completely independt of oemof. 
- this functionality should be used to process results / information of energy systems. In my view this should be independent of solph. If it will be solph dependent, it should be located in the solph package. 
- In the future we might have a result structure on the core level which can be used in combination with the general economic calculation.

uvchik:
To take @simnh comment into account I would suggest to move it to `oemof.solph.economics`.

If it works fine for everybody we could later extract the general part into a separate module/package and keep just the adapter in solph. This won't break the API but makes the economic calculations available for others (if it's worth).

ckaldemeyer:
> To take @simnh comment into account I would suggest to move it to oemof.solph.economics.
> 
> If it works fine for everybody we could later extract the general part into a separate module/package and keep just the adapter in solph. This won't break the API but makes the economic calculations available for others (if it's worth).

I guess that's a good compromise not hindering ourselves with "some day" features.

ckaldemeyer:
Maybe I couldn't see it at on glance, but is the structure of `param_results` and `results` now equal?

For me it would be important to have it that way e.g. flow attributes in `param_results` should be accessible via `param_results[(n1, n2)]['sequences/scalars]` and component attributes via  `param_results[(n, None)]['sequences/scalars]`.

An easy test would be if both structures could be merged somehow e.g. the Series and Dataframes with the same keys are concatenated.

henhuy:
Yes, the structure is now the same!
This indeed makes calculations etc. very comfortable.

ckaldemeyer:
> Yes, the structure is now the same!
> This indeed makes calculations etc. very comfortable.

Will you merge dev in this branch or should I? I think (combined with a simple example in the examples repository) this would be a good proof of concept.

jnnr:
Can this be merged before the release v0.2.1 (issue #447)? To me it seems that we only need an example to finish it.

uvchik:
@henhuy , @simnh @jnnr : Postponed as I it is not ready to merge.

henhuy:
This is a very old try and no longer useful!

PR #437, opened 2018-01-15
change attribut group to dictionary

uvchik:
With the negative/positive gradient we added dictionaries as attributes to the Flow class.

In the constructor they are named "sequences". I just stumbled over it searching for something else, but naming types correctly make things clearer.

If we have more dictionaries in the future we can distinguish between them within the if statement:

```python
if attribute in dictionaries:
    if attribute in typ_a_dictionary:
        do something
    elif attribute in typ_b_dictionary:
        do something else
elif:
    ....
```
As long as we do have just one type of dictionaries the second if statement is left out.

### DIFF
--- a/oemof/solph/network.py
+++ b/oemof/solph/network.py
@@ -68,12 +68,18 @@ class Flow:
         nominal\_value to get the absolute value. If fixed attr is set to True
         the flow variable will be fixed to actual_value * :attr:`nominal_value`
         , I.e. this value is set exogenous.
-    positive_gradient : numeric (sequence or scalar)
-        The normed maximal positive difference (flow[t-1] < flow[t])
-        of two consecutive flow values.
-    negative_gradient : numeric (sequence or scalar)
-        The normed maximum negative difference (from[t-1] > flow[t]) of two
-        consecutive timesteps.
+    positive_gradient : dictionary
+        Two obligate keys:
+        'ub': numeric (sequence, scalar or None), the normed maximal positive
+         difference (flow[t-1] < flow[t]) of two consecutive flow values
+         (ub = upper bound).
+        'costs': numeric (scalar or None), the gradient cost per unit.
+    negative_gradient : dictionary
+        Two obligate keys:
+        'ub': numeric (sequence, scalar or None), the normed maximal negative
+         difference (flow[t-1] > flow[t]) of two consecutive flow values
+         (ub = upper bound).
+        'costs': numeric (scalar or None), the gradient cost per unit.
     summed_max : numeric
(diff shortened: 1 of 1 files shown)


simnh:
I would approve this one, but can be finalize this one first (together with it) #405 =

henhuy:
That was something @ckaldemeyer and I also stumbled upon while creating the param_results in #416.
First, those were added to sequences, as it was an iterable - then we recursively flattened all dicts, so that `{'negative_gradient': {'ub': x, 'costs': y}}` became {'negative_gradient_ub': x, 'negative_gradient_costs': y}` - and finally looked through all sequences and moved "false"-sequences into scalars-dict.
Now, eith your suggestion maybe it makes sense to add an additional key `dictionaries` to the param_results dict?!

uvchik:
My commits do not change anything, I just renamed the attribute-groups, because they were named sequences but they were dictionaries. 

Furthermore, we could of course discuss the idea of dictionary-attributes but this was not the intention of this PR.

uvchik:
@simnh You can set the review on hold with "request changes" and approve it when #405 is merged. In that way it is visible for everybody.

uvchik:
```
positive_gradient : dictionary
       Two obligate keys:
       'ub': numeric (sequence, scalar or None), the normed maximal positive
        difference (flow[t-1] < flow[t]) of two consecutive flow values
        (ub = upper bound).
       'costs': numeric (scalar or None), the gradient cost per unit.
negative_gradient : dictionary
       Two obligate keys:
       'ub': numeric (sequence, scalar or None), the normed maximal negative
        difference (flow[t-1] > flow[t]) of two consecutive flow values
        (ub = upper bound).
        'costs': numeric (scalar or None), the gradient cost per unit.
```

I revised the docstring. Does it look okay for you.

uvchik:
@simnh Don't you think we should merge this into dev and into #405.

uvchik:
This is more or less a fix and does not have effect on the output. Shouldn't we just merge it?

jnnr:
I had a look at the code. In my opinion, it is a good idea to move the gradients out of the sequences list. However, why create a new dictionary and not put it into scalars? Maybe you have a reason for that which I do not see now. In my opinion, gradients are attributes that are belong in the same category as nominal_value.

 Any other comments on this PR before the release of v0.2.1?

uvchik:
It is kind of "blocked" by @simnh but he has not reacted for seven weeks so maybe we just merge it if nothing happens the next days.

jnnr:
Any comments? Should we merge this?

Issue #436, opened 2018-01-15
Default _Sequence causes multiple list extensions. Performance/memory issue? 


henhuy:
I just saw that a sequence with a default value generated by `plumbing._Sequence` does not create a list at first, but extends the list every time an item with higher index is demanded. This leads to the following behavior: Say I have an energysystem with 8760 timesteps, while building this energysystem every sequence which was build by default calls it `__getitem__` function 8760 times and thus the function extends the internal data list every time by only one value. Maybe this is not very good for performance and memory reasons...

Two suggestions:
1. If no additional value is set via `__setitem__` every time only default is returned (no list extension)
2. If length of sequence is known, list is pre-built with `[default] * length` items

What do you say?

uvchik:
I do not understand the implications for the performance but if everything works I am fine with it.

I think you should wait for a comment of @gnn or create a PR and set @gnn as reviewer because he wrote the function.

simnh:
I think 2. never happens. How should the length be know, if so, it will be a iterable object, so there is no problem anyway. 

For the 1. proposal: I think this is a good idea, however, as we have a lot of optional attributes that may be a sequence but do not have to, we implemented it this way as constraints etc. are always constructed for the all timesteps.

So you need to find a solution when a index is provided, as a list is assumend, e.g. `flow_object.min[5]`

Maybe you already know this, then I did not understand your proposal....

One other option might be to do the following instead: If the default value is set as an attribute, find the timeindex of the energy-system-object and do what you said something like `[default] * len(time_index)`

jnnr:
The respective PR has been merged, so I will close this issue.

PR #438, opened 2018-01-16
Revision/expanding sequences Sequences with default value get not expanded

henhuy:
### This revision aims to reduce unnecessary list extensions while creating the model - see related issue #436. 

**Related issue**: #436 
**Located in**: Major change was made to `oemof.solph.plumbing`
**Functionality**:
While building the sequence an internal attribute `self.default_changed` is set to False. Whenever `__getitem__` of sequence is called, only the default value is returned. Before, every time an internal list was extended to requested index. Only, if `__setitem__` is called attribute `self.default_changed` is set to True and an internal list is build, as the sequence no longer consists of only one default value.
**Outcome**:
This revision leads to three things:
1. Testing run-time decreased! (for me, from 8sec to 5sec!) :)
2. `__len__` function of Sequences with only one default value now returns 0 - this is no problem for all tests, but users looking into solph components may get different results.
3. I changed param_results-API, so that Sequences containing only the default value are now added to the 'scalars' dictionary.

**Not implemented**:
The other idea from #436, setting up a list containing entries for every time-step is not implemented and maybe even unnecessary, due to new implementation!

I think these changes are good, improving oemof solph! What do you think?

### DIFF
--- a/oemof/solph/plumbing.py
+++ b/oemof/solph/plumbing.py
@@ -96,11 +96,16 @@ class _Sequence(UserList):
     def __init_list(self):
         self.data = [self.default] * (self.highest_index + 1)
 
+    def __repr__(self):
+        if self.default_changed:
+            return super(_Sequence, self).__repr__()
+        return str([i for i in self])
+
     def __len__(self):
         return max(len(self.data), self.highest_index + 1)
 
     def __iter__(self):
         if self.default_changed:
-            super(_Sequence, self).__iter__()
+            return super(_Sequence, self).__iter__()
         else:
             return repeat(self.default, self.highest_index + 1)
--- a/oemof/solph/plumbing.py
+++ b/oemof/solph/plumbing.py
@@ -7,6 +7,7 @@ __copyright__ = "oemof developer group"
 __license__ = "GPLv3"
 
 from collections import abc, UserList
+from itertools import repeat
 
 
 def sequence(sequence_or_scalar):
@@ -66,9 +67,11 @@ class _Sequence(UserList):
     def __init__(self, *args, **kwargs):
         self.default = kwargs["default"]
         self.default_changed = False
+        self.highest_index = -1
         super().__init__(*args)
 
     def __getitem__(self, key):
+        self.highest_index = max(self.highest_index, key)
         if not self.default_changed:
             return self.default
         try:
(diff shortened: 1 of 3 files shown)


henhuy:
As the param_results dict was changed (to the good, I think) this will have implications for #433. If this PR is accepted, I will change related calculations...

henhuy:
I wonder why there are no travis checks for this PR? Can you explain me?

oemof/solph/plumbing.py
```diff
@@ -65,16 +65,20 @@ class _Sequence(UserList):
     """
     def __init__(self, *args, **kwargs):
         self.default = kwargs["default"]
+        self.default_changed = False
         super().__init__(*args)
 
     def __getitem__(self, key):
+        if not self.default_changed:
+            return self.default
```
simnh: Do I understand this right, that no iterable object but a scalar is returned, if ` not default_changed`? I think, it kind of strange to have a sequence class that does not return a iterable object. My idea was rather to move this to the flow/node classed instead.

simnh:
Sorry, I think I did not read correctly...Nevertheless I am not sure, whether we should mix the idea of default values with the sequence class..

uvchik:
> I wonder why there are no travis checks for this PR? Can you explain me?

That's strange because the test has been done: https://travis-ci.org/oemof/oemof/builds/329396739

The test failed by the way :-P

henhuy:
Bit confused @simnh : Is your request still open, or did you change your mind and approved?

> Nevertheless I am not sure, whether we should mix the idea of default values with the sequence class..

I don't really see the advantage of creating a variable length list, if `solph` only needs the default value every time...
> The test failed by the way :-P

Sorry, i did not check for the docstring tests (I will change my testing parameters to take these into account for the future). The test fails, because no iterable list is created internally. 

So, there is an decision to make, if this behaviour is wanted or not. Personally, I prefer my implementation as oemof is set up faster and I don't see the point in creating a list of 8760 times the same value... 
But no offense there - it was just an idea!

henhuy:
I fixed sequences __len__, __setitem__ and __iter__ function in order to emulate typical list behavior without actually creating list internally. 
Maybe this resolves your last concerns @simnh?

henhuy:
What do you mean with `plumbing` is not in `core` atm?

simnh:
> Is your request still open, or did you change your mind and approved?

I already approved this one, as I changed my mind ;-)

> What do you mean with plumbing is not in core atm?

the module is not located on the core level but in the solph package...because you said above 'this will improve oemof core'

henhuy:
> plumbing is not in core atm...
Ah okay! (I had to re-read my own text) Thanks for the hint - changed PR description...

PR #439, opened 2018-01-17
Revise/nx graph

uvchik:
**Revision of the graph module**

* I could not see the advantage of using a solph model instead of the energy system as the solph model gets its information from the energy system. That's why I removed the option to pass a model instead of an energy system.
* I added a label to the nodes which helps to show the description in external tools
* I added the weight to edges if an energy system is passed
* I added the option to write the graph to disc as an `.graphml` file.

**Things we may want to discuss**

* Do you like the integrated "write" option or do want to leave it to the user?
* At them moment no "weight" attribute is written if the weight is None. We also could add an empty attribute. I think both ways have advantages.

### DIFF
--- a/oemof/graph.py
+++ b/oemof/graph.py
@@ -8,7 +8,7 @@ __license__ = "GPLv3"
 import networkx as nx
 
 
-def create_nx_graph(energy_system=None, remove_nodes=None,
+def create_nx_graph(energy_system=None, remove_nodes=None, filename=None,
                     remove_nodes_with_substrings=None, remove_edges=None):
     """
     Create a `networkx.DiGraph` for the passed energy system and plot it.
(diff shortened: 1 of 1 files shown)


uvchik:
Here are the Travis results: https://travis-ci.org/oemof/oemof/builds/330299021

jnnr:
Looks good and gives nice results. Thanks! I like the labeling and weights. The integrated write function can be useful sometimes. If not needed, one just does not pass a filename, so that's perfectly ok.

uvchik:
> The docstring could be revised. Under "Parameters", "filename" appears twice. Should I fix it?

Are you sure you have the latest commit?

jnnr:
Somehow I didn't. Strange. But now its ok.

PR #440, opened 2018-01-22
Create CODE_OF_CONDUCT.md

uvchik:
A friend told that it is good to have such a code before any problem occurs.

I do not have any time so I just used the github templated with three clicks.

Maybe we do not need it but as it was so easy to create and may not harm we could just use it.


uvchik:
If nobody answers I will merge it next week as "silently accepted" .

Issue #406, opened 2017-12-12
licensing under "oemof developer group"


Ludee:
I'm sorry to bring you more or less bad news.
After an intensive meeting with FSFE in Berlin we have certainty that "oemof developer group" **is not a legal option to host the copyright!** I put this issue on the agenda because I think this is quite important for further developments.

Decided [on the dev meeting](https://github.com/oemof/oemof/issues/218#issuecomment-265700459):
> Decision: Authorship lies with the "oemof developer group"

Short answer: It does not lie there! It has to be a real or legal person.

Here is an extract from the mail Robbie and I send over the openmod:

> Legal matters / Project location
the hosting of projects on neutral terrain is advisable, particularly if several institutes are involved
> 1. one approach is to set up a legal entity for the project (perhaps a German eingetragener Verein (eV) or incorporated society) and use contributor license agreements (CLA) to assign rights
[my comment: notwithstanding, there are downsides to this approach, including substantially reduced appeal for casual contributors, increased administrative overhead, more formalized procedures, and the need to canvas and define legal purposes at the outset] [LH: I totally agree with all arguments. An advantage would be an totally independent developer group. Perhaps there are even more? ]
> 2. Second approach is to keep the copyright with each institute (or company/ persons) in the code (header of each script). But use the “developer group” for communication (GitHub, Website, etc.) [LH: This was the preferred approach by FSFE, because it combines advantages]
> 3. Additional option is to give the copyright to an already existing entity that operates as an umbrella organisation. One example given by Matthias is the Apache Software Foundation (https://apache.org/index.html#projects-list) [LH: Is there any experience with this kind of approach? Pros and Cons?]

uvchik:
So we have to found a foundation or association 😄

simnh:
we already decided, that we don't care about that for now, because we will never charge anyone anyway...

ckaldemeyer:
> we already decided, that we don't care about that for now, because we will never charge anyone anyway...
> 
> but a association like: 'oemof e.V.' would be nice !

I guess the question is different: is our license still valid at all if we do not have a legal body?

In this case, the code legally must/will be treated as if there was no license at all.

jnnr:
Even if it is still half a year to go and we may proceed with this issue in the meantime I already put this as a first topic on the agenda of the dev-meeting in May 2018.

ckaldemeyer:
[...]

PR #442, opened 2018-01-23
change licence into REUSE format

uvchik:
According to #406 I changed the license file with find/replace. This leads to the same copyright in every file but that's all I will do to end this annoying topic.

Feel free to change it for individual files (add your name or remove a name), but remove names with care.

```python
"""
This module is about something....

Copyright (c) 2014-2018 Cord Kaldemeyer, Simon Hilpert, Stefan Günther,
                        Uwe Krien

SPDX-License-Identifier: GPL-3.0-or-later
"""
```

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -55,7 +55,7 @@ master_doc = 'index'
 
 # General information about the project.
 project = u'oemof'
-copyright = u'2014, oemof-Team'
+copyright = u'2014-2018, oemof-developer-team'
 author = u'oemof-Team'
 
 # The version info for the project you're documenting, acts as replacement for
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -55,7 +55,7 @@ master_doc = 'index'
 
 # General information about the project.
 project = u'oemof'
-copyright = u'2014-2018, oemof-developer-team'
+copyright = u'2014-2018, oemof-developer-group'
 author = u'oemof-Team'
 
 # The version info for the project you're documenting, acts as replacement for
--- a/oemof/energy_system.py
+++ b/oemof/energy_system.py
@@ -2,8 +2,9 @@
 
 """Basic EnergySystem class
 
-Copyright (c) 2014-2018 Cord Kaldemeyer, Simon Hilpert, Stefan Günther,
-                        Uwe Krien
+This file is part of project oemof. It's copyrighted by the contributors
+recorded in the version control history of the file, available from its original
+location https://github.com/oemof/oemof/energy_system.py
 
 SPDX-License-Identifier: GPL-3.0-or-later
 """
--- a/oemof/energy_system.py
+++ b/oemof/energy_system.py
@@ -1,10 +1,12 @@
 # -*- coding: utf-8 -*-
 
 """Basic EnergySystem class
-"""
 
-__copyright__ = "oemof developer group"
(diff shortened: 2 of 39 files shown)


uvchik:
Okay I read the reuse text again and found something that exactly meets our needs. The following paragraph comes from (https://reuse.software/practices/2.0/).


> For a project using the version control system to convey information about copyright, you must:
> 
> - make sure the version control system is publicly accessible and take steps1 to ensure it will remain so, 
> - provide a link back to the original file in version control from each header,
> - ensure that if someone copies the source repository without version control system metadata (such as if they make a tar-file of it), then the version control system metadata relevant for determining copyright must still be included, for example in the form of a bill of materials (see later), an automatically generated log file or similar,
> - make sure the commit metadata reflect the actual copyright (this is particularly important if a project accepts code contributed through mailing lists, bug trackers, or similar, where the original copyright holder is not the one pushing code to version control, or where the copyright is held by an organisation rather than the author.)
> 
> An appropriate header in this case would be:
```
 /*
  * This file is part of project X. It's copyrighted by the contributors
  * recorded in the version control history of the file, available from
  * its original location http://git.example.com/X/filename.c
  * 
  * SPDX-License-Identifier: GPL-3.0
  */
```

 In our case e.g. for the graph.py module this would be:

```python
"""
Some module text.....

This file is part of project oemof. It's copyrighted by the contributors
recorded in the version control history of the file, available from
its original location https://github.com/oemof/oemof/graph.py

SPDX-License-Identifier: GPL-3.0
""""

**Comment by uvchik** [comment]
> Some module text..... This file is part of project oemof. It's copyrighted by the contributors recorded in the version control history of the file, available from its original location https://github.com/oemof/oemof/graph.py SPDX-License-Identifier: GPL-3.0

I adapted the files to the text above.

I added the main authors as reviewers. If they agree I will merge this branch to avoid more merge conflicts as necessary in the future.

**Comment by uvchik** [comment]
I did not add the license for the documentation. I think this is something else for somebody else.

@Ludee What you create a PR for the documentation?

**Comment by Ludee** [comment]
> @Ludee What you create a PR for the documentation?

I can change the Wiki template in https://github.com/oemof/oemof/wiki/Header-and-License

Or do you mean an open license on the documentation readthedocs?

**Comment by uvchik** [comment]
Okay, I talked to @Ludee. and we decided to change the copyright of the documentation from `2014 oemof-team' to `2014-2018 oemof-developer-group` just to make it right.

Then we will start a new PR to change it to a different license such as [creative commons](https://creativecommons.org/).

I would prefer a simple "by" license but other developer may think different.

**Comment by Ludee** [comment]
The generated url to the files are not working. They need the `blob/master/` part as well.
When putting long url names (long filename in folder) PEP8 is complaining.

**Comment by uvchik** [comment]
1. PEP8 allows long URLs.

2. If we use the blob we have to change it for every branch which is kind of annoying. What we can do is to provide a basic link `github.com/oemof/oemof` and the relative link `oemof/oemof/solph/network.py` which is shown on top of the site within github.

```
This file is part of project oemof (github.com/oemof/oemof). It's copyrighted by the 
contributors recorded in the version control history of the file, available from its
original location oemof/oemof/solph/network.py

SPDX-License-Identifier: GPL-3.0-or-later
```

oeli

This is valid for every branch!

PR #445, opened 2018-02-02
Added energysystem compatibility for param_results

henhuy:
Just a little addition to former _add-node-param-obj-results_ PR:

Now, not only (Operational-)Model, but also the energysystem can be
given as input parameter for processing.param_results function


henhuy:
It's only a simple feature - I did not know if I can push directly to dev...

oemof/outputlib/processing.py
```diff
@@ -305,10 +309,22 @@ def param_results(om, exclude_none=True, keys_as_str=False):
     The dictionary is keyed by the nodes e.g.
     `results['nodes'][idx]['scalars']`
     and flows e.g. `results['flows'][(n,n)]['sequences']`.
+
+    Parameters
+    ----------
+    system: Model or energysystem
```
ckaldemeyer: I would suggest "Model or EnergySystem" or "optimization model or energy system"

uvchik:
I would say `solph.model` or `solph.EnergySystem`, because it is meant an instance of the `solph.EnergySystem` class and not any energy system.

By the way I would not allow passing a model but only(!) `EnergySystem`. Models gets their informations from the `EnergySystem` so this is the source. Models might vary in the future so it will be additional work to keep both APIs in line. So I would not add but replace the option to pass a Model.

henhuy:
> By the way I would not allow passing a model but only(!) EnergySystem. 
I thought I shouldn't change the API after a release?
That's the only reason I allowed both options...

uvchik:
Ah okay, I did not know that this is already part of v0.2.0, I thought this is something new.

Anyhow, we should talk about the usage of Models and Systems on the next developer meeting.

Issue #448, opened 2018-03-06
oemof does not work with pyomo 5.4.x at the moment


uvchik:
I do not know if it is a pyomo bug or an API change but we should find out.

ckaldemeyer:
https://github.com/oemof/oemof/issues/444

ckaldemeyer:
This is annoying since I need version 5.4 for some additional stuff..

ckaldemeyer:
I have tried to locate the problem and from my opinion it seems to be related to changes within the Pyomo API.

See the following thread for my question on the Pyomo mailing list: https://groups.google.com/forum/#!topic/pyomo-forum/MQbSL2BOOG0

ckaldemeyer:
The quick release history for the latest release underpins our assumption: https://github.com/Pyomo/pyomo/releases

Let's wait for their answer ;-)

ckaldemeyer:
> The quick release history for the latest release underpins our assumption: https://github.com/Pyomo/pyomo/releases
> 
> Let's wait for their answer ;-)

Here's the answer:

```
Cord:

There haven't been API changes, and the expression system didn't change.  This looks like an issue with the index value used to define the constraint group.

There were changes in the logic for testing such values.  

--Bill
```

ckaldemeyer:
This shows how to fix it so we can think about fixing it and then switch to Pyomo >= 5.4.

uvchik:
I just talked to @ckaldemeyer and he will try to fix it this week. I think we should wait with the release so that the new release works with the actual oemof version.

jnnr:
This is a real bug, so I am o.k. with postponing the release a bit to include the fix.

simnh:
I will try to fix this today...

simnh:
Please check out my pull request #449

uvchik:
Closed in favour of #449.

PR #449, opened 2018-03-06
Features/update contraints to new pyomo api

simnh:
  * Adapt internal contraint creation to new pyomo api
  ...

* Related issues:
#448 

See commit message and discussion here:

https://groups.google.com/forum/#!topic/pyomo-forum/MQbSL2BOOG0

There is still an open question as one result check 'fails', but this seems to be a numerical decimal precision issue, as far as I see it. I adpted the corresponding test.

I did not test with old pyomo version. Though I guess there should not be a problem. 

I think this might slow down contraint creation a little maybe only marginally. But I do not see an other way right now, as it is the standard way to deal with this problem. Speeding up oemof, is a different topic. For me this is ok right now.

### DIFF
--- a/oemof/graph.py
+++ b/oemof/graph.py
@@ -110,7 +110,8 @@ def create_nx_graph(energy_system=None, optimization_model=None,
     # passed or undirected edge otherwise
     for n in energy_system.nodes:
         for i in n.inputs.keys():
-            weight = energy_system.flows()[(i, n)].nominal_value
+            weight = getattr(energy_system.flows()[(i, n)],
+                             'nominal_value', None)
             if weight is None:
                 grph.add_edge(i.label, n.label)
             else:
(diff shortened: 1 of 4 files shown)


simnh:
I also adapted the test, please check if that's ok or if i missed something...

simnh:
Oh, I just saw that I pushed something not related to this PR, however, I think this should go into dev as well. For the sake of time just merge this together, I do not have the time right now to revert the commit

uvchik:
It works with 5.3 only the storage test fails with an older pyomo version so we have to test with the most actual version.

uvchik:
Thanks a lot for fixing it :smile: :tada:

uvchik:
I think commit 301b216e45b57223740 is the one that should not be part of this PR, but as it makes things more stable I think we should just keep it....

jnnr:
Thanks!

Issue #450, opened 2018-03-08
Add datapackage-reader


simnh:
We are currently adding functionality to build energy system from a datapackage structure. 

I guess we will be aiming for full oemof-solph functionality without `solph.options` for the start. 

If two complicated we will use facades module, but we will try to be as generic as possible. There is a pull request open: #451

simnh:
Example can be found in the oemof examples repo:

https://github.com/oemof/oemof_examples/tree/features/data-package-reader/examples/oemof_0.2/datapackage-reader

uvchik:
PR #451 is merged and the link does not work any more. I think we can close it.

PR #451, opened 2018-03-08
Features/data package reader

simnh:
**Functionality**
Adds a method to the energy system class to build oemof energy system from datapackage with a predefined structure. 

**Location**
Main functionality will be located on the core level in `oemof.energy_system`

  **API**
   Something like 
```
    es = EnergySystem.from_datapackage(datapackage.json'),
         attributemap={
            object: {"capacity": "nominal_capacity"},
            Flow: {"ub": "nominal_value"}},
        typemap={
            'volatile-generator': Source,
            'bus': Bus,
           'storage': GenericStorage,
          'demand': Sink,
        FLOW_TYPE: Flow})
```

**Related issues**
#450

**Testing Example**
https://github.com/oemof/oemof_examples/tree/features/data-package-reader/examples/oemof_0.2/datapackage-reader

**Notes**
Will first be implement for solph without options. 

**Open questions**
* [ ] How to set the timeindex 

**ToDo's / Errors:**
* [x] typemap requires `'bus': Bus` otherwise error, strange... (error can be produced by altering existing example )
* [x] sequences are written as pd.Series but only there values should be written
* [x] ForeignKey to object reference with string in datapackage does not work inside node_parameters
* [x] empty edge_parameters in datapackage row entry causes an error(error can be produced by altering existing example )
* [x] write arbitrary fields as node attributes (including bus str to object / sequence reference for these entries) this is important for correct working with `oemof.solph.facades`... **NOT IMPORTANT FOR NOW**
* [ ] Documentation
* [ ] Tests


simnh:
Thanks, most things run already as expected. One thing does not work correctly at the moment: 

If you add a Transfomer (1in, 2out) object with node parameters `{"conversion_factors": {"electricity-bus1": 0.4, "heat-bus": 0.5}}`

The conversion factor is also created for the coal bus (even it is not specified above): 

```
In [41]: for k,v in es.groups['BP-chp'].conversion_factors.items(): 
    ...:        print(k.label, v)
    ...:        
    ...:        
heat-bus [0.5, 0.5, 0.5]
global-coal [1, 1, 1]
electricity-bus1 [0.4, 0.4, 0.4]

```



Full object specification:

`BP-chp;backpressure-turbine;{"conversion_factors": {"electricity-bus1": 0.4, "heat-bus": 0.5}};global-coal;electricity-bus1, heat-bus;{"ub": [null, 60, 90], "cost": [12,null,null], "type": ["coal", "electricity", "heat"]}
`

simnh:
Reminder for switching to facades: 
* One tabular resource per Facade (e.g. demands.csv, generators.csv,...)
* Every resource associated with a facade will have a header with mandatory arguments 
  of the class. It should be possible to parse optional arguments. Basically, every field 
  of the resource should be tried to set as an attribute of the facade instance
* attribute mapper and type mapper should stay as they are right now
* If the field-name (column-name) is a resource in *sequences* directroy and the value in this     field points to a sequences column name, the sequence should be set for the argument (works similar right now) 
* Similarly, if it's pointed to a node with an string this should be created (If predecessors/ successors is removed, we need to figure out how we find the total set of possible nodes to search for)
* edge_parameters should stay possible protected arguments for individual parametrization of edges. 
  (need to figure how to point to a specific edge). 

 I will provide small example chunks and  correct associated output using the examples directory. We can use these as test later on...

simnh:
Nice, facades seem to work now except the sequence foreign key does not add the timeseries, but the name of the column of the sequence.

simnh:
Ok functionality should work now as expected...I will start with the documentation...this will be awesome!

simnh:
@gnn Can we finalize this one with the existing functionality? Whats missing?

oemof/tools/datapackage.py
```diff
@@ -64,6 +64,15 @@ def read_facade(facade, facades, create, typemap, data, objects,
...
     """
+    # TODO: Make sure the class var 'required' exists
+    if hasattr(typemap[facade['type']], 'required'):
+        for r in typemap[facade['type']].required:
+            if facade[r] is None:
+                raise AttributeError(
+                    ("Missing attribute `{}` for facade of type `{}` with" +
+                     " name/label `{}`.").\
+                        format(str(r), facade['type'], facade['name']))
+
```
gnn: Hey. Unless I'm missing something here, there's a better way to do this. See my comments in c034689.

PR #453, opened 2018-03-09
Add console test and clean installation test

uvchik:
  * Clean code according to developer rules
  * Console script added to test oemof including all nosetests and doctests. I did not add anything to the documentation because I think we can test how it works and if it works fine I will add a line there.
  * Just use `test_oemof` within your virtual environment.
  * I think it should be added it to the v0.2.1 release, because it is mostly code-clean-up and a new hidden function that could be removed in the next release if it does not work as intended. No risk in my opinion :smile:


jnnr:
Looks good. Thanks!

PR #460, opened 2018-03-12
Make deserialized nodes serializable again

gnn:
@ckaldemeyer discovered a bug which meant that `Node`s could not be dumped again after being restored. He discovered this by using `copy.deepcopy` which uses the same protocol as dumping/restoring via `pickle`.
I fixed the issue and would like to have the fix in the next release.
@ckaldemeyer  also made an [example](https://github.com/oemof/oemof_examples/blob/5c120d70f326d8784e7395d22011ea72d5659df0/examples/oemof_0.2/generic_chp/bpt.py) which gets fixed by this branch, so I know my fix works.

### DIFF
--- a/doc/whatsnew/v0-2-1.rst
+++ b/doc/whatsnew/v0-2-1.rst
@@ -37,6 +37,14 @@ New features
     (`Issue #445 <https://github.com/oemof/oemof/issues/445>`_)
 
 
+Fixes
+#####
+
+  * Deserializing a :class:`Node <oemof.network.Node>` lead to an object which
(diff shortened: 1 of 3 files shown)


ckaldemeyer:
@jnnr : Can we integrate this into `v.0.2.1`? I would really appreciate that as it was a bug and it is a feature that I need for my work.

Thanks a lot @gnn ! :tada:

ckaldemeyer:
> @ckaldemeyer also made an example which gets fixed by this branch, so I know my fix works.

I will delete the branch as it was only to illustrate the error..

PR #462, opened 2018-03-12
Add option for multiindex dataframe on view

simnh:
Adds the possiblity to return dataframe with multiindex column / index for a view


ckaldemeyer:
Would it be an idea to shift the code that's applied twice into a function? Nevertheless, I am fine. You can decide if you leave it like this or not..

simnh:
Yes this would be an option, however its not the exact same code....which kind of complicates things...I will think about this!

Issue #464, opened 2018-03-15
Error: Cannot run example "simple_dispatch"


henhuy:
We just tried to simply run:
_oemof_examples/examples/oemof_2.0/simple_dispatch/simple_dispatch.py_ 
within a fresh environment and oemof installed via 
`pip install oemof`
But the following error occured:
`"Index '(<oemof.solph.network.Bus object at 0x7f1681a639a8>, 0)' is not valid for indexed component 'Bus.balance'"`
Any hints?

uvchik:
Which commit?

uvchik:
Sorry, if you use `pip install` you have v0.2.0 installed.

This is an Pyomo issue. Use `pip install pyomo==5.3`.

uvchik:
This issuee will be fixed with the release of v0.2.1 which is delayed due to illness. I will be released in a few days. Afterwards you can update pyomo again using `pip install pyomo --upgrade`.

henhuy:
Works again! thank you!

uvchik:
I reopen this as a reminder.

I suggest to limit the version of each required package in the setup.py to the most actual (working) version on every oemof release. The reason is that otherwise `pip install oemof` will always install the most actual version of each package at the moment of the installation. This behaviour may lead to errors (as seen above) if the the API of one required package changed since the last oemof release.

@henhuy and @ckaldemeyer already agree on that.

uvchik:
Closed in favour of #465.

PR #465, opened 2018-03-16
limit version of required packages to the most actual version

uvchik:
This PR will limit the version of each required package in the setup.py to the most actual (working) version for the v0.2.1 release.

The reason is that otherwise pip install oemof will always install the most actual version of each package at the moment of the installation. This behaviour may lead to errors (see #464) if the the API of one required package changed since the last oemof release.

You may check the versions right before the release if I missed an actual important update.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -29,12 +29,12 @@ setup(name='oemof',
       packages=find_packages(),
       package_data={'oemof': [
           os.path.join('tools', 'default_files', '*.ini')]},
-      install_requires=['dill',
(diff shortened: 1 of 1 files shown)


simnh:
Why don't we set pyomo to >= 5.4  with the next release?

uvchik:
> Why don't we set pyomo to >= 5.4 with the next release?

I think it is not necessary to limit it so narrowly, because solph works fine with 4.4.

PR #466, opened 2018-03-18
Add constraint group methods to solph classes

simnh:
**Description**
  * Adds methods for easier coupling between solph classes and their constraints.
  * Only changes internal solph logic
  * Two test that fail need to be adapted

* [ ] Fix tests 
* [x] Check if bus.balanced works correct (maybe add test for this-> not now)

### DIFF
--- a/oemof/outputlib/processing.py
+++ b/oemof/outputlib/processing.py
@@ -206,7 +206,8 @@ def __separate_attrs(system, get_flows=False, exclude_none=True):
     def detect_scalars_and_sequences(com):
         com_data = {'scalars': {}, 'sequences': {}}
 
-        exclusions = ('__', '_', 'registry', 'inputs', 'outputs')
+        exclusions = ('__', '_', 'registry', 'inputs', 'outputs',
+                      'constraint_group')
         attrs = [i for i in dir(com)
                  if not (callable(i) or i.startswith(exclusions))]
 
--- a/oemof/solph/components.py
+++ b/oemof/solph/components.py
@@ -168,6 +168,11 @@ class GenericStorage(network.Transformer):
                 if not isinstance(flow.investment, Investment):
                     flow.investment = Investment()
 
+    def constraint_group(self):
+        if isinstance(self.investment, Investment):
+            return GenericInvestmentStorageBlock
+        else:
+            return GenericStorageBlock
 
 class GenericStorageBlock(SimpleBlock):
     r"""Storage without an :class:`.Investment` object.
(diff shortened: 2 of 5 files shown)


uvchik:
Could you briefly explain, why your changes effect the results and how the tests has to be adapted.

simnh:
Two test that fail are located in `test_processing.Parameter_Result_Tests.test_nodes_without_none_exclusion`. There the attributes of objects are compared, now there is a new object in the comparison...Which would actually enhance the test, as the grouping will also be checked: 

```
======================================================================
FAIL: test_processing.Parameter_Result_Tests.test_nodes_without_none_exclusion
----------------------------------------------------------------------
  File "user/lib/python3.5/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/user/oemof/tests/test_processing.py", line 178, in test_nodes_without_none_exclusion
    'conversion_factors_b_diesel': 1,
AssertionError: {'label': 'diesel', 'constraint_group': <bound method Transformer.constraint_group of <oemof.solph.network.Transformer object at 0x7fa43d5ec6d8>>, 'conversion_factors_b_diesel': 1, 'conversion_factors_b_el1': 2} != {'label': 'diesel', 'conversion_factors_b_diesel': 1, 'conversion_factors_b_el1': 2}
```

simnh:
Ok I fixed the first problem by adding the new method to the `exclusion` variable. Now there is only the temrination condition left...

henhuy:
Considering the resulting new parameter in `test_processing.Parameter_Result_Tests.test_nodes_without_none_exclusion` - I recently had a solph component stored in the parameter results, as by now only "private" members (`self.__member`) and the inputs and outputs are excluded from resulting dict. For myself, I changed the code to exclude all solph components from being stored in parameter results. I think, maybe we should exclude those constraints too - as solph components and constraints are neither "scalars" nor "sequences", but complex structures which aren't simple input parameters...
What do you think?

simnh:
> What do you think?

See above, I just added the `constraint_group` to exclusion

henhuy:
> I fixed the first problem by adding the new method to the exclusion variable
This is what I wanted to say...

simnh:
Ok, does anyone know why the termination condition changed, though the lp-file is the same (I assume, because these tests work)

uvchik:
All tests are fine for me (locally) and Travis seems to be satisfied, too :smile:

simnh:
Ok I will merge then!

Issue #476, opened 2018-04-17
Keep the sequence after mathematical calculations


uvchik:
I have two sequence attributes (`a` and `b`) and want to multiply them to get `c`. Is there a way that `c` is a sequences as well?

henhuy:
Yes - that should be the normal case.
I will add this functionalit

henhuy:
Sorry, I pressed the wrong button...

uvchik:
An example that shows the problem can be found in the `ExtractionTurbineCHP` class.
A loop over all TIMESTEPS is used to calculate the `main_flow_loss_index`, just in case the sequence is a real time series. This produces a real time series even if all attributes are scalars (pseudo lists).

```python
n.main_flow_loss_index = [
    (n.conversion_factor_full_condensation_sq[t] -
     n.conversion_factors[m.es.groups[n.main_output.label]][t]) /
     n.conversion_factors[m.es.groups[n.tapped_output.label]][t]
    for t in m.TIMESTEPS]
```
 [see full example](https://github.com/oemof/oemof/blob/dev/oemof/solph/components.py#L935)

It would be nice if it would work like this:

```python
n.main_flow_loss_index = (
    (n.conversion_factor_full_condensation_sq -
     n.conversion_factors[m.es.groups[n.main_output.label]]) /
     n.conversion_factors[m.es.groups[n.tapped_output.label]]
    )
```

If it is not possible like this we could test whether one of the attributes is a list and loop only if a real list is present.

uvchik:
Closed in favour of #477.

PR #477, opened 2018-04-19
Mathematical operations on sequences

henhuy:
Now it is possible to perform mathematical operations on two Sequences (see Issue #476).
To do so, I had to change inner logic of `oemof.solph.plumbing.sequence` function to wrap real lists into `_Sequence` as well.

**Note:**
Now, `sequence([1, 2, 3])` will return a `_Sequence` containing the list instead the list itself! 
All Tests pass, but if `sequence` is used elsewhere, expecting to return a real `list` this might cause errors... Opinions?

* Examples:
```
from oemof.solph.plumbing import sequence
>>> a = sequence(3)
>>> a + sequence(5)
8
>>> a[2] = 4
>>> a + sequence(5)
[8, 8, 9]
```
See `tests.test_solph_plumbing` for further examples.

* Notes:
I did not implement _i_-operations, like `a += 1`

* Related issues: #476


uvchik:
Thank you for work. I will have a look soon and test it in my classes.

uvchik:
I add some test and adapted the calculations in the `ExtractionTurbineCHP` class.

For me it works as intended.

@gnn What do you think.

gnn:
If the argument always gets wrapped, there's no need to differentiate between
`sequence` and `_Sequence` anymore. The whole point of having both was that
`sequence` could inspect its argument and return any list like (non-str)
argument unchanged, while changing scalars to transparently behave like lists
of unknown (potentially infinite) length in order for them to interact
correctly with real list by adjusting to those. That was also the reason why
there was no `__len__` or `__iter__` method as, e.g. `zip(sequence(0), [1])`
was supposed to yield `[(0, 1)]` instead of `[]` like it does now. Obviously I
should've written a test or documentation for that... but I digress. Long story
short, if we keep the implementation the way it is, I suggest to distribute the
functionality to a new class `Sequence` because then it's explicit and
transparent that the result is always a `Sequence` instance and we have the
option to handle the case where everything gets boxed differently from boxing
only scalars.

BUT it also looks like what gets implemented are mathematical operations the
way `numpy.ndarray` supports them? If so, maybe the simplest thing would be to
subclass from `ndarray` instead of `UserList` and keep everything else the way
it is.

uvchik:
@henhuy @gnn, I think this will never happen. Shall we close it?

henhuy:
yes, I think you can close it...

Issue #311, opened 2017-05-09
Implementation of base storage and adaption of the current one: @uvchik @birgits  


ckaldemeyer:
As discussed on the user meeting..

birgits:
Just found this issue again but don't at all remember what it's about... Does any of you remember @uvchik @ckaldemeyer?

uvchik:
At the moment the storage contains a storage component and two transformers that manage the input efficiency and output efficiency. The idea was to create a basic storage component without these transformers. People sometimes want to connect things in a different way and such a component would be flexible and less complicated.

ckaldemeyer:
>> Just found this issue again but don't at all remember what it's about... Does any of you remember @uvchik @ckaldemeyer?

> At the moment the storage contains a storage component and two transformers that manage the input efficiency and output efficiency. The idea was to create a basic storage component without these transformers. People sometimes want to connect things in a different way and such a component would be flexible and less complicated.

The idea was also to make the storage the first custom component: https://github.com/oemof/oemof/pull/332

But let's discuss it if the PRs are ready code-wise :-)

uvchik:
This issue has been addressed by PR #480, which is already part of the v0.2.2 release.

PR #478, opened 2018-04-23
Move flow creation in storage to method

simnh:
Just moves code to method, as subclassing was kind of diffcult, when so many things are going on inside the __init__

Please approve ;-)

### DIFF
--- a/oemof/solph/components.py
+++ b/oemof/solph/components.py
@@ -127,24 +127,33 @@ class GenericStorage(network.Transformer):
         self.investment = kwargs.get('investment')
 
         # General error messages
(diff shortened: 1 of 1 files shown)


uvchik:
I am fine with the changes.

The plans to revise the storage #311 came up again with [openmod thread 745](https://forum.openmod-initiative.org/t/power-investment-with-the-flow-class/745).

Are you planing to sub-class in your app or do you want to add a new component?

simnh:
We are subclassing on app level...but we may also need to add another component soon. 

Thanks for approving!

PR #481, opened 2018-05-02
Avoid error for flow objects without investment/nonconvex attribute

simnh:


uvchik:
Is this a bug-fix? Shall I merge it? I will release tomorrow.

simnh:
Yes you can merge...just a minor improvement to avoid errors when subclassing and flexible modelling...

PR #482, opened 2018-05-03
Features/analyzer

henhuy:
This feature aims to support flexible analysis on oemof results.
Therefore, three new classes are introduced:
* `Analyzer` (abstract): The class is used to create new analyzer performing different actions given a node tuple from _results_ or _param_results_. Additionally, analyzers can depend on other analyzers and use their output to prevent duplicate calculations (for example summation of flows is performed in `SequenceFlowSumAnalyzer` and can later be used by `BusBalanceAnalyzer`, but also by further `Analyzers` if needed). Furthermore, `Analyzer` objects can check if needed depending `Analyzers` are added and if required `Iterator` is used.
* `Iterator` (abstract): Different iterators can be implemented to perform various iterations on _results_ or _param_results_
* `Analysis`: Can hold multiple analyzers. When starting analysis, an iterator is used to send node_tuples to all connected analyzers

Current implemented analyzers:
* `SequenceFlowSumAnalyzer`: Calculates results[current_tuple]['sequence']['flows'].sum()
* `InvestAnalyzer`: Calculates investment multiplied by ep_costs
* `VariableCostAnalyzer`: Calculates variable_costs multiplied by flow sum
* `FlowTypeAnalyzer`: Categorizes flows by input and output
* `NodeBalanceAnalyzer`: Returns node balance (input and output sums) of current node
* `BusBalanceAnalyzer`: Same as NodeBalanceAnalyzer, but only for Busses
* `LCOEAnalyzer`: Calculates LCOE by adding results from `InvestAnalyzer` and `VariableCostAnalyzer` divided by flow sum of all demand outputs (therefore, `LCOEAnalyzer` has to be initialized with list of demand outputs!)

**Example: (Updated API)**
For the user the whole analysis is kept simple:
```
import analyzer
analysis = analyzer.Analysis(results, param_results, iterator=analyzer.FlowNodeIterator)  # optional auch non-standard iterator angeben
seq = analyzer.SequenceFlowSumAnalyzer()
ft = analyzer.FlowTypeAnalyzer()
bb_analyzer = analyzer.BusBalanceAnalyzer()
analysis.add_analyzer(seq)
analysis.add_analyzer(ft)
analysis.add_analyzer(bb)
analyzer.analyze()
balance = bb_analyzer.result[(bus, None)]  # Beispielzugriff auf ein Ergebnis
```
See `oemof.tests.test_solph_analyzer` for more examples

**Insights/Remarks:**
For this module I made highly use of _Design Patterns_:
* `Analisys` is implemented as _Singleton Pattern_
* `Iterator` uses _Iterator Pattern_
* `Analyzer` is derived from _Visitor Pattern_ (I changed some aspects)


henhuy:
I also made an UML to show underlying class structure, but I don't know where to put it that you can see it...!?

uvchik:
hendrik_uml

henhuy:
danke @uvchik

uvchik:
I like the new features and would suggest to add it to make it possible to use and test. Could we add something like "This feature is works fine but is still in a testing state, be aware that the location or API will change in the future...." or something like this. We could remove this warning after a while. It is just because it is kind of a complicated feature and I do not know who will be able to fix it if you are not available.

uvchik:
I would like to test it could we move it to the outputlib and merge it. We could add a logging warning that it is a "beta-feature". In that way we can test it and if everything works fine we can announce it officially in the 0.3 release.

henhuy:
As I there wasn't enough time/ to much concurrency to present my workshop on analyzer feature, I put my jupyter notebook into cloud folder:
https://cloud.rl-institut.de/index.php/s/a2ETAX6Yw48Nyjn?path=%2Foemof_dev_analyzer

Feedback wanted - decision wanted ;) Thanks!

jakob-wo:
Hi,
I "won" this Issue at the Issue auction (dev meeting 2019)... and now I would like to advance the Issue and this nice analysing feature.

> As there wasn't enough time/ to much concurrency to present my workshop on analyzer feature, I put my jupyter notebook into cloud folder (better idea, where to put it?):
> https://cloud.rl-institut.de/index.php/s/a2ETAX6Yw48Nyjn?path=%2Foemof_dev_analyzer
> 
> Feedback wanted - decision wanted ;) 

@henhuy with the link to your jupyter notebook I end up with "502 Bad Gateway" and without notebooks. Could you provide the notebook or an example again. That would be great!

uvchik:
The views module has been moved from the outputlib to solph. The outputlib no longer exists.

henhuy:
Did not see that something was happening here...
I will add it to my TODO-List and hopefully get the time to work on it again!

uvchik:
If you have a problem with the merge conflicts I can help you.

uvchik:
The "Analyzer" is interesting for more concrete models because it does not work for the generic concept of solph.

PR #483, opened 2018-05-05
Add option pre-select block vars for result data structure

ckaldemeyer:
As described, this implements the possibility to pre-select block vars which can be passed as a parameter e.g. `blockvars=['GenericStorageBlock.capacity', 'GenericCHPBlock.Y', ...]`.

If this is of interest, we can integrate it. Otherwise, I'll keep it locally..

### DIFF
--- a/oemof/outputlib/processing.py
+++ b/oemof/outputlib/processing.py
@@ -60,7 +60,7 @@ def remove_timestep(x):
         return x[:-1]
 
 
-def create_dataframe(om):
+def create_dataframe(om, blockvars=None):
     """
     Create a result dataframe with all optimization data.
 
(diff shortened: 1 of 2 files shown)


ckaldemeyer:
I'll improve this a bit an then assign you, again. Nevertheless, let me know if you think if this feature is of interest.

uvchik:
:+1: The additional "solvers" are really nice to have even though I do not know anything about the differences.

:question: I think we should keep the official oemof output API because all the views rely on this API. 

:+1: Adding a filter that only fetches some results does not break the API and will be a really useful feature.

ckaldemeyer:
This is outdated and will probably be replaced with the changes to mutable parameters..

PR #484, opened 2018-05-06
Add option for objective function timestep  weighting indepent of con…

simnh:
**Description**
 * Add possiblity to only set weighting for objective function expressions (independent of constraints)
 * For example useful for temporal aggregation where only economic weighting is wanted 
 
Just pass an array.
```
Model(es, objective_weighting=[10, 10, ...])
```

If kw is not passed, the time increment will be used. Therefore, nothing changes for users.

### DIFF
--- a/oemof/solph/blocks.py
+++ b/oemof/solph/blocks.py
@@ -219,7 +219,7 @@ class Flow(SimpleBlock):
         for i, o in m.FLOWS:
             if m.flows[i, o].variable_costs[0] is not None:
                 for t in m.TIMESTEPS:
-                    variable_costs += (m.flow[i, o, t] * m.timeincrement[t] *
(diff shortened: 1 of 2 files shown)


ckaldemeyer:
> Description
> 
>     Add possiblity to only set weighting for objective function expressions (independent of constraints)
>     For example useful for temporal aggregation where only economic weighting is wanted
> 
> Just pass an array.
> 
> Model(es, objective_weighting=[10, 10, ...])
> 
> If kw is not passed, the time increment will be used. Therefore, nothing changes for users.

Thanks for implementing this! I think it is a nice add-on and can be very helpful. Should we integrate a test?

ckaldemeyer:
Well, I'll leave it up to you. As the feature is easy to grasp and optional I would be fine also without a test!

Issue #475, opened 2018-04-17
node view does not work for parameters


uvchik:
I heard that the `param_results()` function produces the same structure as the `results()` function. But the following code does not work.

```python
results = outputlib.processing.results(model)
param = outputlib.processing.param_results(model)
res_view = outputlib.views.node(results, 'my_bus')
param_view = outputlib.views.node(param, 'my_bus')
```

The last line throws out the following error:
`AttributeError: 'dict' object has no attribute 'empty'`

If I use a different code (the key instead of the label) I do not get an error anymore but an empty dictionary.
```python
>>> outputlib.views.node(params, ('my_bus', None))
{}
```
Do I think the wrong way or is this a bug.

henhuy:
I just looked into code and found underlying problem - 
We have different data types in `results` and `param_results`:
In `results`, `scalars` and `sequences` are `pandas.Series` or `pandas.DataFrame` respectively.
In `param_results` we made it dicts...
oh no

uvchik:
> So I think, we have to change param_results to use pandas.

I agree because it would be nice to have the same structure for both.

In the next step we could add a merge_results function.

`all_results = merge_results(results, param_resutls)`

Or an `parameter`attribute to the results function.

`all_results = resutls(model, parameter=True)`

What do you think?

uvchik:
Solved with PR #494.

Issue #470, opened 2018-03-22
Making the result object a class


simnh:
I know we already thought about this. But I think a class with method instead of functions for the processing and particularly views as wells as writing results to csv for example could really help...should not be too much of an effort...

What do you think?

ckaldemeyer:
> I know we already thought about this. But I think a class with method instead of functions for the processing and particularly views as wells as writing results to csv for example could really help...should not be too much of an effort...
> 
> What do you think?

I could go with it as long as we do not blow it up with specific plotting code!

ckaldemeyer:
But I won't spend time on it..

simnh:
No, no plotting, just methods for user-friendly result processing :-) ..

ckaldemeyer:
> No, no plotting, just methods for user-friendly result processing :-) ..

:+1:

uvchik:
As there are always different ways to a good solution I am fine with the procedural approach as well as an object oriented approach. The new proposal should be at least as good as the existing functionality :smile: 

@henhuy thought in the same direction maybe you could talk to him to share your ideas.

@ckaldemeyer In December we moved the plotting module to an extra repository (oemof-visio) and in January we removed the plotting functions from the graph module. Since v0.2.0 the matplotlib is no longer a requirement of oemof as decided on the developer meeting. I think a refactoring from a procedural approach to an object oriented approach has nothing to do with this decision.

ckaldemeyer:
> As there are always different ways to a good solution I am fine with the procedural approach as well as an object oriented approach. The new proposal should be at least as good as the existing functionality smile
> 
> @henhuy thought in the same direction maybe you could talk to him to share your ideas.
> 
> @ckaldemeyer In December we moved the plotting module to an extra repository (oemof-visio) and in January we removed the plotting functions from the graph module. Since v0.2.0 the matplotlib is no longer a requirement of oemof as decided on the developer meeting. I think a refactoring from a procedural approach to an object oriented approach has nothing to do with this decision.

:+1:

henhuy:
I really appreciate this!
In my private branch I'm already doing so.
For example - my first adaption:
My app calculates multiple results and therefore calls the `node`-function for several nodes multiple times - to avoid this, my `Results`-object stores already made node results and only returns them, instead of calling the function again. 
To finish: I would definitely participate in creating an oemof results object!

henhuy:
[...]

Issue #463, opened 2018-03-12
Outputlib: Param Results 


simnh:
Hey, 

I just stumbled upon the `param_results` function in the outputlib (I kind of missed the PR). I really like the idea of this feature and its a great enhancement I think! However, I would like to point out a few things: 

* I guess the Docstring is not correct, at least I get back only tuples as keys and if its about the node parameters i get `(node, None)`. The docu states that flows are found under key `flows` and nodes under `nodes`. But maybe I just don't understand the doc string right? 
[see code here](https://github.com/oemof/oemof/blob/ba321fd38fcc7d94932684fbdaae631af1b1eff4/oemof/outputlib/processing.py#L319-L321)
* The outputlib may be the right place for now, but I think the name is kind of confusing, as their are not results processed but input parameter (if I understand it correct?). Therefore I would suggest renaming it. 
* In addition, I would like to think about if None values should be converted to strings when `key_as_str `is set to True. I think we could rather leave it as None

I think this all relates to handling of input/output (exo-endogenous parameter handling, viewing, processing), the scenario definition and energy system methods. But the things above can be addressed anyway...

uvchik:
This was a project of @ckaldemeyer and @henhuy and I did not use it so far, so I removed my name from the assignees.

henhuy:
Hi @simnh,
thank you for your feedback.
Indeed, I agree to all of your recommendations!
As Uwe pointed out, we shouldn't change the API. So I think, I will cover point 1 and 3, but renaming of the function has to be postponed, right?

simnh:
Yes, but we should keep it in mind before the next major release v0.3...

We can make a PR for this and discuss further things there...

uvchik:
> Yes, but we should keep it in mind before the next major release v0.3...

Indeed. We can already start the roadmap for the v0.3 release and add everything we would like to change for the next mayor release.  We can discuss the list on the dev-meeting in May and complete the release on the dev-meeting in December... :smile:

uvchik:
The None/'None' issue is still open. Everything else is fixed.

busiing:
Self-assignment due to oemof dev meeting - issue auction

uvchik:
I think this is fixed with a compromise. By default None will be 'None' if the keys have been converted to strings but one can set the parameter `keep_none_type` to True to avoid 'None' strings and keep the None Type.

```python
convert_keys_to_strings(result, keep_none_type=True)
```

PR #485, opened 2018-05-07
Revise param_results/results (e.g. #463)

henhuy:
This PR fixes points 1 & 3 from #463 

 * [x] ~~fix documentation~~
 * [x] None is not String but NoneType after to_string operation
 * [x] ~~Rename parm_results to `parameter_as_dict`~~
 * [x] ~~Add to outputlib documentation~~ -> see #500

Open questions?

 * [x] ~~Should parm_results work with views as well (s. #475, #494)?~~
 * [x] ~~Should the result dictionary rather be an object (s #470)?~~ (nobody is interested by now)


jnnr:
I just discussed with Uwe about the API of the outputlib, and we would like to ask you about your experience and opinion.

@henhuy @gnn @simnh : In which cases is the results dictionary used in such a way that the keys are really the objects? The cases I've seen always convert to keys_as_strings. The oemof_examples do that. Can you give any examples, i.e. use cases, in which the actual objects are used?

uvchik:
I have two ideas how we could merge the results and param_results into one structure.

`all_results = merge_results(results, param_resutls)`

`all_results = resutls(model, parameter=True)`

uvchik:
> I have two ideas how we could merge the results and param_results into one structure.
> 
> all_results = merge_results(results, param_resutls)
> 
> all_results = results(model, parameter=True)

I changed my mind. Let us merge it.

uvchik:
If nobody objects we can merge.

uvchik:
Sorry, I accidentally closed this PR.

The new work-around does not work with tuples as label.

uvchik:
I changed my mind a little. I am fine with both variants None (none-type) and 'None' (string). It just have to be clear. We could make it optional with something like `keep_none_type=False`. This would leave the API untouched.

henhuy:
> We could make it optional with something like `keep_none_type=False`
I am okay with that solution!
If #508 is accepted we can merge it and then tests should pass...

henhuy:
Added `keep_none_type=False` parameter to `processing.convert_keys_to_string` **and** `views.node` function (as this function uses `processing.convert_keys_to_strings` internally).
Now, this PR waits for acceptation of PR #508 - after that, tests should pass...

uvchik:
Hey, great :-)

I agree to set keep_none_type=False as default, because the default call of `convert_keys_to_string` should convert all(!) keys to strings, that is what people expect.

We have to wait for @gnn to approve #508 but then we can merge both. Thanks again.

uvchik:
I accidentality merged this PR a little too early. 

We just waited for PR #508 which is merged now. 

I check the dev branch after the merge and all tests are fine :-)

Thank you @henhuy for the improvements.

PR #489, opened 2018-05-14
Features/solph existing investment

simnh:
**Description**

Allows to set existing capacity for Investment objects

Has been asked here:
https://forum.openmod-initiative.org/t/transformer-investment-mode-while-considering-for-existing-capacity/821

**Related issues**
#480 

**Other comments and questions**

We need to agree on the naming and the location. One could argue, that the better place would be the nominal value of the flow objects. But I think using it in the investment object is also fine. But feel free to adapt accordingly. 
I just added the adaption inside the constraints, because it was quite easy for me atm...

**API**
```python
solph.Investment(ep_costs=[float], minimum=[float], maximum=[float], existing=[float])
```

**TODO'S**

* [x] Agree on location
* [x] Agree on naming 
* [x] Fix one test
* [x] Write new test 
* [x] Add information to What's new and documentation


uvchik:
I added the API to make it easier for @gplssm.

uvchik:
I added some test but I do not really understand the behaviour.

uvchik:
Commit 1ee2218 fixes a really odd behaviour. I should not be necessary to set the registry to None.

...but it works now :-)

simnh:
Nice!

gplssm:
Can I help somehow without digging too deep into oemof? Maybe I could to have one hour this evening...

uvchik:
> Can I help somehow without digging too deep into oemof? Maybe I could to have one hour this evening...

You could have a look a [this script-test](https://github.com/oemof/oemof/blob/features/solph-existing-investment/tests/test_scripts/test_solph/test_simple_model/test_simple_invest.py). It might be too big. Maybe you can shorten it and try out if it behaves like it should behave if `existing > 0`.

If I add one 1MW existing capacity to an investment wind turbine the investment of the wind turbine decreases by 5MW and the investment of the pv plant also decreases by some MW. For me this does not make sense but maybe I just do not understand it. It might be easier to check if we remove some components.

gplssm:
>     Can I help somehow without digging too deep into oemof? Maybe I could to have one hour this evening...
> 
> You could have a look a this script-test. It might be too big. Maybe you can shorten it and try out if it behaves like it should behave if existing > 0.
> 
> If I add one 1MW existing capacity to an investment wind turbine the investment of the wind turbine decreases by 5MW and the investment of the pv plant also decreases by some MW. For me this does not make sense but maybe I just do not understand it. It might be easier to check if we remove some components.

I can try to reduce to less technologies being involved. I'll do it this evening....

oemof/solph/blocks.py
```diff
@@ -367,7 +367,8 @@ def _investflow_fixed_rule(block, i, o, t):
             """Rule definition of constraint to fix flow variable
             of investment flow to (normed) actual value
             """
-            return (m.flow[i, o, t] == (self.invest[i, o] *
+            return (m.flow[i, o, t] == (m.flows[i, o].investment.existing +
```
simnh: missing brackets

uvchik:
Thanks @simnh. I will have a look a let you, @gplssm know if it is still necessary.

simnh:
I think some brackets have been messed up..could you cheack again?

simnh:
The storage should be fine though...

uvchik:
Everything seems to work now. Adding an existing capacity of 100 to the investment wind turbine decreases the invest variable by 100. This does not have to be like this in larger models but indicates that everything works fine now :-)

uvchik:
We could add the new option to the documentation.

http://oemof.readthedocs.io/en/latest/oemof_solph.html?highlight=investment#using-the-investment-mode

Just a sentence. Could you do that @gplssm. I think the code works fine now but you might want to check this anyway.

simnh:
Just out of curiosity: What is the application you are building and for what purpose?

uvchik:
I added this branch to readthedocs. Now it is easier to check out your changes. You have to commit and push your changes to start the build process:

http://readthedocs.org/projects/oemof/builds/

The results can be found here:

http://oemof.readthedocs.io/en/features-solph-existing-investment/oemof_solph.html#using-the-investment-mode

gplssm:
> Just out of curiosity: What is the application you are building and for what purpose?

I'm creating an OS implementation of the model I used for my PhD. 

> BTW: oemof support is always this fast: less-than-24h-new-feature guarantee (only for premium users)

I highly appreciate this support! It's nice to be a premium user :smile:

gplssm:
I polished the "What's new" section and added a small part about the existing capacity in the investment mode in the documentation. I have just some quick questions @simnh / @uvchik 

- do you think it's worth to mention the changed tests in "what's new"?
- "If existing capacity is considered for a component with investment mode enabled,
the *ep_costs* apply to both, the newly built capacity and existing capacity." is this true?
- "All parameters that usually refer to the *nominal_value/capacity* will
now refer to the investment variables and existing capacity." is this true?

gplssm:
I don't have the access rights to push to this branch. Is it complicated granting them or shall I create a separate PR?

uvchik:
> do you think it's worth to mention the changed tests in "what's new"?

No, I don't.



> "If existing capacity is considered for a component with investment mode enabled,
> the ep_costs apply to both, the newly built capacity and existing capacity." is this true?

No, the `ep_costs` apply to the newly build capacity but not to the existing capacity. This is the main point of this feature, otherwise you could have used the `minimum` attribute for the existing capacity. In that case the `ep_costs` will apply to the existing and the newly build capacity.


> "All parameters that usually refer to the nominal_value/capacity will
> now refer to the investment variables and existing capacity." is this true?

Yes, if you want to set the simultaneity of power plants within a region you could use the `max` attribute (e.g. `max=0.9`). The `flow_var` will get an upper bound like this:
 `flow_var <= (invest_var + existing) * max`

gplssm:
Thanks for answering. I'll update the docs accordingly...

PR #492, opened 2018-05-17
Changes of documentation of ExtractionTurbineCHP

FranziPl:
This PR continues the PR #486 from @gnn, which improves the documentation of the component "Extraction Turbine".

I changed the the formulas of the constraints in the docstring again:
 - Now, common symbols (common for the CHP) are used (for example P_el for elektric power). A table below the formulas translates these symols into the attributes used in the code.
 - The main-flow-loss-index is used in the constrain and is explained in a third formula
 - One of the equal-signs is changed to a greater or equal sign

In the documentation, little part of the text is changed. Furthermore, an image is added, which explains the constrains.


uvchik:
Looks really cool :tada:  Thank you.

PR #491, opened 2018-05-16
Decouple the investment of storage flows and capacity - Part 2

uvchik:
* Describe your pull request as transparent as possible
  * Fix pep8
  * Add tests
  * Rename attributes with deprecated warning
  ...

* Follows PR #480 

* [x] Add tests
* [x] Adapt tests
* ~~Non-cycled storage~~

PR #493, opened 2018-05-24
Revise GenericStorage and add option to decouple last and first timestep of the storage capacity

uvchik:
An option should be added to decouple the last and the first time step of the
storage capacity. At the moment the last time step has always the same
value as the first.

The idea is to add an optional parameter `cycled=True/False` (naming?) to decide whether the last time step is connectd to the first or not.

see #491

The problem is the actual storage balance. We may have to add an extra constraint for the first OR last time step and a new variable for the `initial capacity`.

* [x] implement constraints
* [x] adapt documentation
* [ ] add tests 
* [x] rename attributes according to #538 
* [x] remove deprecated attributes
* [x] provide an example


uvchik:
I will try to integrate this feature until the dev-meeting.

uvchik:
https://oemof.readthedocs.io/en/features-add_option_to_decouple_last_first_timestep/api/oemof.solph.html#oemof.solph.components.GenericStorageBlock

p-snft:
> The idea is to add an optional parameter `cycled=True/False` (naming?) to decide whether the last time step is connectd to the first or not.

Do you mean `cyclic`? (For me, _cycled_ sounds like _rode a bicycle_.)

PS: I am looking forward to seeing this branch being merged.

uvchik:
It came from a previous discussion and I just started to implement it without thinking about the name. But you are right `cylced` does not sound very good. I would suggest `balanced`.

uvchik:
> PS: I am looking forward to seeing this branch being merged.

It breaks the API so it will be part of v0.3.0. If you need it you can already merge it in your own local branch. The code is adapted but we need to adapt the documentation and to add tests.

p-snft:
`balanced` is a good choice.

ckaldemeyer:
> It came from a previous discussion and I just started to implement it without thinking about the name. But you are right `cylced` does not sound very good. I would suggest `balanced`.

:+1: I like it, too.

p-snft:
I like the description. Reading it, even revealed a possible problem that should be tackled in the implementation (at some point):

When `balanced=False` and `initial_capacity=None`, the initial capacity will typically be optimized to 1, the final capacity to 0: This allows free energy consumption. (I assume, energy typically has a cost.) A solution might be to add a cost for the initial and the final capacity. (The latter might be negative costs to recompensate the choice to not consume/ sell that energy, yet.) This way, both capacities could be optimized more realistically.

Probably, you do not need to make this explicit in the documentation. You clearly warn the user, already.

joroeder:
Thanks for you feedback!

joroeder:
> I would mark code phrase as balanced=False as code. I think it works like this ``code``.

I wonder if this has to be adapted in the whole documentation? Or just code phrases and not single words like ``True``?

uvchik:
> I wonder if this has to be adapted in the whole documentation? Or just code phrases and not single words like `True`?

I think it depends on the context but you are right it is not often used in the documentation. In my opinion it improves readability but it might be a matter of taste.

p-snft:
I adopted the documentation using an updated version of the commit I previously named in #544. In the process, I also simplified the notation. (When defining the `initial_capacity` to be the capacity at t<0, no additional case is needed for t=0.)

uvchik:
There was a rst-bug, but now you can find the results [here](https://oemof.readthedocs.io/en/features-add_option_to_decouple_last_first_timestep/api/oemof.solph.html#oemof.solph.components.GenericStorageBlock) (scroll down).

In my opinion it looks better than before.

D = Delta ( I don't know how to get greek symbols here)

First I would say that t is not the time but the number of the time step. The time increment does not have to be equal. Therefore, t is the actual time step and t-1 the previous time step. We could say that `E[-1]` is the `initial_capacity`. But if t is the time step we cannot name the time increment `Dt`.
We could use `i` instead of `t` for the time step and use `Dt` for the time increment. Or we use `t` for the time step and find something else for the time increment e.g. Tau.

p-snft:
Thanks for the corrections. Is there a good way to check this offline?

Symbols: I see. I assumed equally spaced time steps. In that case, my notation would yield the same results. So, it is probably best to replace `Δt` by `τ(t)` to emphasize that it does not have to be constant. I will rework the documentation accordingly.

PS: I use UTF-8 for Greek letters.

uvchik:
> Thanks for the corrections. Is there a good way to check this offline?

```
sphinx-build /your/path/oemof/doc /your/path/oemof/doc/_build/
```
In firefox you watch the results:
```
file:///your/path/oemof/doc/_build/api/oemof.solph.html#oemof.solph.components.GenericStorageBlock
```

p-snft:
While revising the documentation, I found that there is `initial_cap`, which is an energy (`0 < init\_cap <  nominal\_capacity`). `initial_capacity`, on the other hand is not an energy, but more a state of charge (or *storage level*).

As this will be fixed using the new naming scheme (#538), I will now address this. For a short period of time, however, this might lead to new inconsistencies.

p-snft:
Somehow, I have managed to introduce changes that run with Python 3.5 but that do not run with Python 3.6 or Python 3.6. I do not see, where the problem comes from. Could somebody please help?

uvchik:
Some IDEs have refactoring tools which make such things much easier.

Thank you for your work. Looks fine now :tada:

p-snft:
I tried using the refactoring functionality of PyCharm, but it did not rename kwargs and strings that just have (and should have) the same name as the variable. Also, backslashes are used in the docstrings. Those prevented the IDE from identifying the names also in these places. In the end it took longer then expected ("It's just refactoring, so ...") but it was OK.

Note that I did not touch occurrences written in ALL_CAPS . What are their meanings?

uvchik:
#### Until now the docstring was very close to the code/mathematics.

### Sets
STORAGES - ALL storages
STORAGES_BALANCED - A set of all Storage objects, with ‘balanced’ attribute set to True.)
STORAGES_SOMETHING
### Constraints
E = ... for all storages in set STORAGES
E[t_last] = E[-1] for all storages in set STORAGES_BALANCED
something for all storages in set STORAGES_SOMETHING


#### The sets and the constraints are not writen next to each othere. Therefore, I would rather like to remove the sets and write something like this:
### For all storages:
E = ...

### For all storages with attribute `balanced=True`
E[t_last] = E[-1]

### For all storages with something
something

uvchik:
We should not forget the documentation for the InvestmentStorage :exclamation:

uvchik:
> "one reason for the solver not to find a solution is “there might be a storage somewhere not able to reach its initial capacity at the end of the period”

This is from [here](https://forum.openmod-initiative.org/t/infeasible-solution-with-a-lower-bound-on-a-flow-otherwise-bigger-than-the-limit/1353/5). We could think of adding this to the documentation if balanced is True.

p-snft:
@uvchik: Would you please check the adjusted unit test I added in the last commit?

uvchik:
@p-snft Thank you :smile: 

As I mentioned above, I do not like the SETS in the docstrings at all, because people have to scroll up and down. At the dev-meeting in Magdeburg we decided to detach the docstrings from the code to make it better readable. In my opinion the removal of the sets would help, but we have to decide what we want according to #544. The worst thing is a different docstring in each class.

The test looks fine if it is a balanced storage. In the commit message you called it (un)balanced storage, so I do not know for sure.

In my opinion the equations are a little bit to small. A line break would help.

-> https://oemof.readthedocs.io/en/features-add_option_to_decouple_last_first_timestep/api/oemof.solph.html#oemof.solph.components.GenericInvestmentStorageBlock

p-snft:
The test is for the balanced case of a storage that can also be unbalanced. I only updated the the existing test, so a separate test for an unbalanced storage is missing.

p-snft:
I am sorry, the test is not missing.

* @joroeder already contributed [test_storage_decouple_first_last_timestep.py](https://github.com/oemof/oemof/blob/fd963eec3326009e564a9c630d396f42deb0092d/tests/test_scripts/test_solph/test_storage_decouple_first_last_timestep/test_storage_decouple_first_last_timestep.py).
*An [example](https://github.com/oemof/oemof/blob/fd963eec3326009e564a9c630d396f42deb0092d/tests/test_scripts/test_solph/test_storage_decouple_first_last_timestep/example_storage_decouple_first_last_timestep.py) was added by @uvchik.
* I adapted the documentation (content-wise, the style is not completely updated but I would consider this a separate task).

All that is missing from the checklist is to "remove deprecated attributes". Can you elaborate?

PS: I would consider to merge this rather soon to have a consistent code base for documentation and style improvements. At the moments, there are too many diverged branches to do this effectively.

uvchik:
I rebased this PR because we do have some API changes and we decided to use the release branch in this case.

1. I added constraint tests for an unbalanced storage (invest/no invest)
2. I remove the test script but kept the example script because the example script is just the clean variant of the test script. I also deactivated it because it creates plots which is not good for a test.

I think we could keep it, but we have to remove the plots and add some test (comparing the results) but I would prefer to move it to the example repository because we do not really need it if everything works fine.

@p-snft , @joroeder Do you agree? If you agree, could you move the example to the oemof-examples? If you do not agree could you remove the plots and add some test? Anyway I would move it to the oemof-examples because it is a good example.

If this is done, I agree that we should merge to avoid conflicts.

p-snft:
I agree and moved the example.

joroeder:
I agree, thanks!

PR #494, opened 2018-05-28
Fixed #485

henhuy:
`scalars` and `sequences` of `param_results` are now pandas objects

Function `outputlib.views.node` now also works with `param_results`.

Related issue: #485


henhuy:
This fix changes the API as output changes

uvchik:
I changed the base to dev03, where we collect the API-changes.

uvchik:
I fixed the tests caused by another PR but there is a Python-3.6-issue. It seems to me that the order of the keys is different in Python 3.6 and therefore the assert test fails.

uvchik:
Thank you for your fixes. Shall we merge this one into dev03 and go on with PR #485.

gnn:
Yeah, sure.

Also, something totally unrelated: here's a [short][0] and a [slightly longer][1] guide to commit messages. The two points why I bring that up are these:

  * `Capitalized, short (50 chars or less) summary`: the first letter should be capitalized because the first line of a commit might be used as a subject line in an email and those are usually capitalized to make them easily distinguishable from a leading `Re: ` or a `[mailing list] ` prefix.
  * `Write your commit message in the imperative...`: There are two reasons for this. First: it lines up with how Git phrases its automatically generated commit messages. Second: a subject line written in imperative completes a sentence starting with: "If you apply this commit, it will ...".

It's nothing important really, but I thought now's as good a time to point this out as any. And yeah. I should probably put those links in "How to contribute" or somewhere similar.

[0]: https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html
[1]: https://chris.beams.io/posts/git-commit/

uvchik:
> And yeah. I should probably put those links in "How to contribute" or somewhere similar.

We already have the second link in our dev-rules:
http://oemof.readthedocs.io/en/stable/developing_oemof.html#commit-message

PR #499, opened 2018-05-31
Add test to keep the API of dump/restore within a mayor release

uvchik:
By now we have not tested if a dump created with a lower version can be restored in higher version. This should be the case within a mayor release.

With PR #466 we broke the backward compatibility within a mayor release even though we are not allowed to do so following our own rules.

I was effected by this bug but I do not know if there is a way to fix it. If not I will adapt the test to avoid such problems in the future.


uvchik:
I think it does not make sense to repair it. It was annoying but we will not turn the wheel back.

Workaround:

* Downgrade oemof to v0.2.0
* restore you old dump and create a results dictionary using outputlib.processing
* dump the results dictionary
* Upgrade oemof and restore your results dictionary.

Now you will be able to use the actual outputlib functions.

@gnn If you agree I would merge this PR and take no further action.

PR #507, opened 2018-07-03
Fix/tuple as label causes problems

uvchik:
If we use tuples as labels as it was originally intended but never used some errors occur.

I fixed some of these bugs and adapted some tests. 

- ExtractionTurbineCHP now uses objects instead of labels
- The graph module uses the string of the label instead of the label (`str(n.label)` instead of `n.label`)
- in the `detect_scalars_and_sequences` function the label is filtered to be not detected as sequence if the label is a tuple.

I want to merge them soon.

### DIFF
--- a/doc/whatsnew/v0-2-3.rst
+++ b/doc/whatsnew/v0-2-3.rst
@@ -26,6 +26,7 @@ Known issues
 
 Bug fixes
 #########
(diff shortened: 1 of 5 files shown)


uvchik:
Any objections here. It effects my app and it would help to merge it soon.

uvchik:
No objections, tests fine, coverall fine... I will merge :smile:

PR #508, opened 2018-07-03
Change existing test to make it fail

uvchik:
It should be possible to use tuple and(!) string labels in one EnergySystem. At the moment it is only possible to use either string or tuple labels.

If you mix them you get the following error:
```python
  File ".virtualenv/oemof_cosmos/lib/python3.5/site-packages/pyomo/core/base/block.py", line 1314, in component_data_objects sort=sort):
  File ".virtualenv/oemof_cosmos/lib/python3.5/site-packages/pyomo/core/base/block.py", line 1244, in _component_data_iter _items = sorted(_items, key=itemgetter(0))
  File "oemof/oemof/network.py", line 228, in __lt__
    return self.label < other.label
```

### DIFF
--- a/oemof/network.py
+++ b/oemof/network.py
@@ -225,7 +225,7 @@ class Node:
     def __lt__(self, other):
         if other is None:
             return False
-        return self.label < other.label
+        return str(self.label) < str(other.label)
 
     def __hash__(self):
         return hash(self.label)
--- a/oemof/network.py
+++ b/oemof/network.py
@@ -223,10 +223,7 @@ class Node:
         return id(self) == id(other)
 
     def __lt__(self, other):
-        try:
-            return str(self.label) < str(other.label)
-        except AttributeError:
-            return False
+        return str(self) < str(other)
 
     def __hash__(self):
         return hash(self.label)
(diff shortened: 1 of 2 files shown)


henhuy:
I suggest a very simple solution:
Comparison of the str-representation of labels, instead of labels themselves...
More complex comparisons are possible, but I would avoid over-engineering it!

uvchik:
Sounds reasonable. 

@gnn Do you agree with this solution?

henhuy:
Added try-except AttributeError-Block to avoid error, if `other` is not a `Node` and therefor has no attribute `label` (this can happen when pandas indices, containing mixed indices of `Node` and `str` are sorted).

gnn:
> I suggest a very simple solution:
> Comparison of the str-representation of labels, instead of labels themselves...
> More complex comparisons are possible, but I would avoid over-engineering it!

That's the solution I would've been going for too, so it gets a :+1: from me. The only thing is, that it can be simplified a bit. I pushed the respective changes and explained it in the commit message.
Now you guys can have a look and if you're OK with it, it's good to get merged.

henhuy:
Simple and nice! Thank you!

uvchik:
:+1:  You can squash and merge or simply merge 

We do not have a merge-strategy so far, but we could think about to have one in the future. Feel free to add an issue or a topic for the developer meeting.

PR #509, opened 2018-07-08
Add operation cost independent from output to NONCONVEX_FLOW

mloenneberga:
This PR implements the functionality of adding operation cost independently from the output at each time interval. As a result, opr_cost can be now attributed the NONCONVEX_FLOW objects which occur whenever the status of nonconvex flow is 1, otherwise they are 0. 

For me, this functionality was practically useful when modelling e.g. DieselGenerator (DG), as it allowed me to add fix operation cost in case the DG-component is running.

Changes are made to oemof/solph/blocks.py and oemof/sopf/options.py

What do you _think?

### DIFF
--- a/oemof/solph/blocks.py
+++ b/oemof/solph/blocks.py
@@ -863,6 +863,6 @@ class NonConvexFlow(SimpleBlock):
                         m.flows[i, o].nonconvex.activity_costs[t]
                         for t in m.TIMESTEPS)
 
-            self.operationcosts = Expression(expr=activitycosts)
+            self.activitycosts = Expression(expr=activitycosts)
 
         return startcosts + shutdowncosts + activitycosts
--- a/oemof/solph/blocks.py
+++ b/oemof/solph/blocks.py
@@ -582,6 +582,9 @@ class NonConvexFlow(SimpleBlock):
     SHUTDOWN_FLOWS
         A subset of set NONCONVEX_FLOWS with the attribute
         :attr:`shutdown_costs` being not None.
+    OPR_FLOWS
+        A subset of set NONCONVEX_FLOWS with the attribute
+        :attr:`opr_costs` being not None.
 
     MINUPTIMEFLOWS
         A subset of set NONCONVEX_FLOWS with the attribute
@@ -678,6 +681,11 @@ class NonConvexFlow(SimpleBlock):
             \sum_{i, o \in SHUTDOWN\_FLOWS} \sum_t shutdown(i, o, t) \
                 \cdot shutdown\_costs(i, o)
 
+    If :attr:`nonconvex.opr_costs` is set by the user:
+        .. math::
+            \sum_{i, o \in OPR\_FLOWS} \sum_t status(i, o, t) \
+                \cdot opr\_costs(i, o)
+
     """
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
@@ -717,6 +725,10 @@ class NonConvexFlow(SimpleBlock):
         self.MINDOWNTIMEFLOWS = Set(initialize=[(g[0], g[1]) for g in group
                                     if g[2].nonconvex.minimum_downtime
                                     is not None])
+
+        self.OPR_FLOWS = Set(initialize=[(g[0], g[1]) for g in group
+                                 if g[2].nonconvex.opr_costs[0]
+                                 is not None])
 
         # ################### VARIABLES AND CONSTRAINTS #######################
         self.status = Var(self.NONCONVEX_FLOWS, m.TIMESTEPS, within=Binary)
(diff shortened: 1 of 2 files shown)


simnh:
I think the overall implementation looks fine. I am just a little concerned with the naming at this point. 

Generally, I would suggest moving the operational cost attribute to the Flow class on the solph level. We would need to change the API to get the naming straight (variable cost, fixed operation cost etc). 

On the other hand this seems like a major conceptual effort to get everything straight. Therefore I suggest adding this as it is and create another issue / PR for refactoring all the cost stuff...

uvchik:
@simnh Do you think of implementing this in a general way? I have never used it, but as I understand @mloenneberga these costs are fixed once the flow is greater than `0`. I wonder if it is possible to implement it generally? You could have something like this in the objective function but I do not know how the objective handles `NaN` or if the difference is already treated as non-linear.:

`flow_var(i, o, t)/flow_var(i, o, t) * operational_cost(i, o)`

I guess the code above does not work so I do not see a way to implement it without integer variables (I may oversee something :smile:)  and if it only works for non-convex flows I would not define them in the 'normal' flows and keep it in the non-convex flows.

I would name them rather `operating_cost` than `opr_costs`. But I am unsure with the name at all. I would name them like `operation_time_costs` to make clear that they are not related to the operation output but the operation time. As you work in this subject you may know the "right" term better than I do.

We aim to increase the test coverage but your PR decreases it, so please write a test :exclamation: :smile:

mloenneberga:
I am totally with you regarding the naming. I like the operation_time_cost suggestion, though it is also rather long name for a variable. Does anyone have other (shorter) naming suggestions?
As i understood, the more general implementation, as @simnh suggested it, would be to simply add the opr_cost-attribute to the FLOW-objects instead of setting it within NONCONVEX_FLOWS. Correct? 
Restructuring it in another PR sounds good to me, since the moment startup/shutdown costs are attributed to NONCONVEX_FLOWS, too.

Finally, having a test would be great. Though I have no clue how to write one myself. Is there anyone who may help out?

uvchik:
> Restructuring it in another PR sounds good to me, since the moment startup/shutdown costs are attributed to NONCONVEX_FLOWS, too.

You are right. Let us finish this one and take the next step afterwards.

> Finally, having a test would be great. Though I have no clue how to write one myself. Is there anyone who may help out?

Please post a component definition here and I will show you how to write a constraint test.

uvchik:
Keep the definition as reduced as possible:

Here is an example of a "constraint test":

```python
    def test_min_max_runtime(self):
        """One sentence what the test is about."""

        bus_t = solph.Bus(label='Bus_T')

        solph.Source(
            label='cheap_plant_min_down_constraints',
            outputs={bus_t: solph.Flow(
                nominal_value=10, variable_costs=10,
                nonconvex=solph.NonConvex(
                    minimum_downtime=4, minimum_uptime=2, initial_status=2,
                    startup_costs=5, shutdown_costs=7))})
        self.compare_lp_files('min_max_runtime.lp')
```

simnh:
> As i understood, the more general implementation, as @simnh suggested it, would be to simply add the opr_cost-attribute to the FLOW-objects instead of setting it within NONCONVEX_FLOWS. Correct?

Yes! 


Ok the problem here should not be discussed in text form I assume. Though I am not really happy how its implemented right now (I do not see a solph-consitent concept for the naming) you can move on like this.

oemof/solph/blocks.py
```diff
@@ -582,6 +582,9 @@ class NonConvexFlow(SimpleBlock):
     SHUTDOWN_FLOWS
         A subset of set NONCONVEX_FLOWS with the attribute
         :attr:`shutdown_costs` being not None.
+    OPR_FLOWS
```
simnh: Could we just use ACTIVE_FLOWS ? If not, nevertheless using OPERATIONAL_COST

oemof/solph/blocks.py
```diff
@@ -582,6 +582,9 @@ class NonConvexFlow(SimpleBlock):
...
+        A subset of set NONCONVEX_FLOWS with the attribute
+        :attr:`opr_costs` being not None.
```
simnh: Sam as above.... voting for explicit naming here, i.e. `operating_costs`

uvchik:
> I am totally with you regarding the naming. I like the operation_time_cost suggestion

Long names are no problem and better than cryptic names. In oemof's [naming rules](http://oemof.readthedocs.io/en/stable/developing_oemof.html#naming-conventions) it says: "Use talking names".

p-snft:
@uvchik: A test analogue to the constraint test you suggested above was easy to implement. Still, I cannot push it. (This pull request is filed against mloenneberga:features_operationcosts. So he has to add the test. Alternatively, I can close this request and pull a different branch.)

p-snft:
I did not realize that I could change the base branch. Thanks for your help.

Issue #511, opened 2018-07-09
python 3.7 release


uvchik:
Python 3.7 is released and we should start to test it.

How long shall we support python 3.4?

I would suggest to replace python 3.4 with python 3.7 in the travis test to save resources! 

At the moment we test 3.4, 3.5, 3.6.

I think one of the "slowest" Linux distributions (debian stable) released debian 9 (stretch) with python 3.5 a year ago. As python 3.5 is backwards compatible it should be no problem to update from 3.4.

@oemof/oemof-developer-group Do you agree?

uvchik:
No opinions, comments...?

I would like to **remove python 3.4** from the test and **add python 3.7**. That means that oemof officially supports **Python >= 3.5**. It may work for other version but as it is not tested it is not guaranteed.

@oemof/oemof-developer-group Do you agree?

c-moeller:
Sounds good.

henhuy:
I agree!

p-snft:
Done with the release of [v0.2.3](https://github.com/oemof/oemof/releases/tag/v0.2.3).

PR #516, opened 2018-08-27
Terminate python 3.4 support and start python 3.7 support

uvchik:
As discussed in #511 we want to stop the support of python 3.4 and start the support of 3.7. The most important step to do so is to change travis test.

Python 3.4 may still work but it is no longer tested anymore. If incompatible code is use for a new feature, a fix etc. we may not be aware. Anyway it should be easy to update python to version 3.5 or higher because the version is backward compatible to 3.4.

@oemof/oemof-developer-group : Last chance for objections!


uvchik:
I removed the workarounds because they do not work either and causes new problems.

There is an open issue on Travis-Ci python 3.7 support.

I will put this on on hold until 3.7 is officially supported by travis.

uvchik:
- pyomo will support Python 3.7 with the next release, but there is no release date published. On github a minor release should be published ['soon'](https://github.com/Pyomo/pyomo/issues/604).

- networkx is planning to support Python 3.7 with the [next release](https://github.com/networkx/networkx/milestone/10), planned on September 07, 2018.

- Our datapackage reader relies on [cchardet](https://github.com/PyYoshi/cChardet) which does not support Python 3.7 at the moment. No further information found.

uvchik:
Pyomo published a Python 3.7 compatible version and all test work fine now.

Issue #306, opened 2017-04-22
Simulation with one timestep not possible


simnh:
Cannot simulate an energy system with only one timestep...

ckaldemeyer:
> Cannot simulate an energy system with only one timestep...

This is probably due to the set "previous_timesteps", right?

simnh:
maybe...

uvchik:
An error message would be helpful.

If it is easy to fix, we should fix it. Otherwise we could catch it and do not allow less than two steps.

uvchik:
Some people were thinking about looping time step by time step around a simple model. This will not be possible if it is not possible to simulate one time step. I do not have the time to care about now but I think it's good to keep it in mind. So I will postpone it.

uvchik:
Closed by #519

PR #519, opened 2018-09-06
Add test and/or example with one time step

uvchik:
Due to a pyomo error it was not possible to model one time step but this issue seems to be fixed recently. We should add tests to make sure that it keeps working.

* Related to #306 

* Share your knowledge:
In the past there was a pyomo error message which we could not be solve but this message is gone and a test with a simple example was successful.

* [x] Add to whatsnew, it was announced as known issue, so we should announce if it is solved
* [x] Add a tests
* [ ] ~Add an example~
* [x] Check documentation


uvchik:
I added a test with one time step, I checked the documentation and could not find anything about the minimum number of time steps. There is an old whatsnew entry that says that it is not possible but the actual whatsnew will make clear that it is now possible. I do not think that I will add an example at the moment. This can be done by users/developers who use this feature in time step model or any kind of outer loop.

uvchik:
Sorry the tests has interfered with the previous tests I had to explicit switch off the auto-registration.

PR #522, opened 2018-09-18
Features/improve offset transformer

uvchik:
* Docstring was wrong. Dictionary was described but list/tuple is used. Is tuple a good idea or would it be better to use to parameters?
* Sequences was not used, so it was not possible to pass a scalar

* [x] Improve the documentation. Describe the coefficients was an image(?).
* [ ] Create an example for the example repository?
* [x] Add equations to docstring
* [ ] ~Add diagram with three slopes into documentation (no intersect, intersect, with/without minrange)~
* [x] Add constraint test (@uvchik gets combinations)
* [ ] ~Create a warning if the y-intersect is positive and no min range is defined~
* [x] Add test (@ckaldemeyer could deliver an application for a heat pump and @mloenneberga for a diesel generator?)


uvchik:
@mloenneberga Did you use the OffsetTransformer? Do you have an example?

ckaldemeyer:
Hi guys! Thanks for this feature but I am not able to contribute at the moment. But keep on going ;-)

uvchik:
Both errors in the init function of the `OffsetTransformer` class are untested. The tests need to be added.

How to test exceptions: https://stackoverflow.com/a/40638805

Contact me if you have questions.

```python
def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)

        self.coefficients = [
            solph_sequence(i) for i in kwargs.get('coefficients')]

        for k, v in self.inputs.items():
            if not v.nonconvex:
                raise TypeError('Input flows must be of type NonConvexFlow!')

        if len(self.inputs) > 1 or len(self.outputs) > 1:
            raise ValueError("Component `OffsetTransformer` must not have" +
                             "more than 1 input and 1 output!")
```

uvchik:
I think the component looks pretty good now. The tests are fine the docstring is good and the documentation readable. One can always do more but this should be enough for the release.

Thank you @joroeder for your contribution :exclamation:

uvchik:
I would wait for one approval.

PR #525, opened 2018-09-26
Make the docu on Generic CHP more transparent

jnnr:
The documentation of the Generic CHP component has been extended in the following way (see rendered version at https://oemof.readthedocs.io/en/features-docu-improvements-jnnr/oemof_solph.html#genericchp-component):

* A figure shows the operation ranges for the 3 types of CHP 
* A table lists all attributes, their corresponding mathematical symbol  and their meaning
* We have corrected the explanation of `H_L_FG_max` in the docstring of the component to "flue gas losses at maximum heat extraction".
* The constraints are listed as formulas using the mathematical symbols
* The text has been extended. It explains how to implement each of the types and what this means in terms of constraints implied.

The changes affect both the doc as well as the docstring of the component.

Some remarks:

* It is counterintuitive to give a backpressure turbine a parameter wo_DH. This fits with CCET, however, and is probably simply due to having one component for all 3 types. An suggestion: We could change P_max_woDH to P_max.
* We believe that it is most transparent to set Beta=0 for backpressure_turbine. This should be adapted in the examples. See [oemof_examples PR 29](https://github.com/oemof/oemof-examples/pull/29)

Finally, a question: Does constraint 4 have any effect? P_el_wo_DH is not used in any other constraint or output.


FranziPl:
I'm fine with it.

ckaldemeyer:
> * It is counterintuitive to give a backpressure turbine a parameter wo_DH. This fits with CCET, however, and is probably simply due to having one component for all 3 types. An suggestion: We could change P_max_woDH to P_max.
> 
> * We believe that it is most transparent to set Beta=0 for backpressure_turbine. This should be adapted in the examples.

:+1:

jnnr:
> It is counterintuitive to give a backpressure turbine a parameter wo_DH. This fits with CCET, however, and is probably simply due to having one component for all 3 types. An suggestion: We could change P_max_woDH to P_max.

Any objections to this?

ckaldemeyer:
> > It is counterintuitive to give a backpressure turbine a parameter wo_DH. This fits with CCET, however, and is probably simply due to having one component for all 3 types. An suggestion: We could change P_max_woDH to P_max.
> 
> Any objections to this?

As I said, I would leave it as it is described like this in the formulation paper and does not change a thing.

jnnr:
Ok, I will leave that one like it is and merge. Thanks for your comments and feedback!

PR #529, opened 2018-11-06
Squelch deprecation warning when testing processing

gnn:
It's really simple. The tests weren't fully updated to take a renamed function into account and the resulting deprecation warning messed up the test output. The PR really only renames the old function call to the new one.

### DIFF
--- a/tests/test_processing.py
+++ b/tests/test_processing.py
@@ -116,9 +116,7 @@ class Parameter_Result_Tests:
         """
         b_el2 = self.es.groups['b_el2']
         demand = self.es.groups['demand_el']
-        param_results = processing.parameter_as_dict(
-                self.es,
-                exclude_none=False)
+        param_results = processing.param_results(self.es, exclude_none=False)
 
         scalar_attributes = {
             'fixed': True,
--- a/tests/test_processing.py
+++ b/tests/test_processing.py
@@ -116,7 +116,9 @@ class Parameter_Result_Tests:
         """
         b_el2 = self.es.groups['b_el2']
         demand = self.es.groups['demand_el']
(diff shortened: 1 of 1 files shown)


gnn:
Ok, the PR no longer renames the old function name to the new one. Instead, it now catches the `DeprecationWarning` and checks its correctness. Have a look at b7340d8 to see how it's done, @uvchik. Of course you're welcome to check the other changes too, but they're purely cosmetic. Since this test will be removed soon with the next major release, it was a bit like doing the sand mandala version of coding. :)

uvchik:
Thank you. Looks better now.

I merged it to avoid conflicts with the upcoming changes. We should conserve it as a "How to" in the Wiki pages because this will not be the last `DeprecatedWarning`.

uvchik:
https://github.com/oemof/oemof/wiki/How-to-test

gnn:
Thanks. I'll add testing warnings to the wiki.
One question: why did we start to do squash merges?

uvchik:
> But why did we start to do squash merges? Did we decide on this somehow and I didn't notice?

No, it seems to be the default github option now. Sorry, I accidentally pressed the default button.

I actually do not like it but you could just ask the others and make a quick decision. I would prefer to use merge commits (again).

gnn:
No biggie. Good to know that this is github's default way (or that it's even an option at all), as it reminds me of paying attention. ;)
Yeah, squash merges are the most horrible (because they have actual non-cosmetic disadvantages) way of doing it, so it's a no-brainer to go back to doing regular merges (with `--no-ff`, but github does this anyway for PRs).
Man, it's always weird when tools start acting up. :)

PR #532, opened 2018-11-07
Improve `groupings` and `network` coverage

gnn:
The test coverage of these modules is a bit on the low side, so this branch collects some additional tests improving it.

### DIFF
--- a/oemof/groupings.py
+++ b/oemof/groupings.py
@@ -109,18 +109,18 @@ class Grouping:
             if kw in kwargs:
                 setattr(self, kw, kwargs[kw])
 
-    def key(self, e):
+    def key(self, node):
         """ Obtain a key under which to store the group.
 
         You have to supply this method yourself using the :obj:`key` parameter
         when creating :class:`Grouping` instances.
 
-        Called for every :class:`entity <oemof.core.network.Entity>` :obj:`e`
-        of the energy system. Expected to return the key (i.e. a valid
-        :class:`hashable`) under which the group :meth:`value(e)
-        <Grouping.value>` will be stored. If it should be added to more than
-        one group, return a :class:`list` (or any other :class:`non-hashable
-        <Hashable>`, :class:`iterable`) containing the group keys.
+        Called for every :class:`node <oemof.core.network.Node>` of the energy
+        system. Expected to return the key (i.e. a valid :class:`hashable`)
+        under which the group :meth:`value(node) <Grouping.value>` will be
+        stored. If it should be added to more than one group, return a
+        :class:`list` (or any other non-:class:`hashable <Hashable>`,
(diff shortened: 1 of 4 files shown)


oemof/groupings.py
```diff
@@ -113,26 +115,26 @@ def __init__(self, key=None, constant_key=None, filter=None, **kwargs):
...
 
         Return :obj:`None` if you don't want to store :obj:`e` in a group.
         """
-        raise NotImplementedError(
-            "There is no default implementation for `Groupings.key`.\n" +
-            "Congratulations, you managed to execute supposedly " +
-            "unreachable code.\n" +
-            "Please let us know by filing a bug at:\n\n    " +
+        raise NotImplementedError("\n\n"
+            "There is no default implementation for `Groupings.key`.\n"
```
p-snft: I am not sure whether explicit concatenation was better.

oemof/network.py
```diff
@@ -257,12 +259,12 @@ def __init__(self, input=None, output=None, flow=None, values=None,
             **kwargs):
         if flow is not None and values is not None:
             raise ValueError(
-                    "`Edge`'s `flow` and `values` keyword arguments are " +
-                    "aliases of each other, so they're mutually exclusive.\n" +
-                    "You supplied:\n"+
-                    "    `flow`  : {}".format(flow) +
-                    "    `values`: {}".format(values) +
-                    "\nChoose one.")
+                    "\n\n`Edge`'s `flow` and `values` keyword arguments are "
```
p-snft: Above, you delete double newlines, here you add some.

oemof/groupings.py
```diff
@@ -9,15 +9,17 @@
...
-    from collections.abc import (Hashable, Iterable, Mapping,
-                                 MutableMapping as MuMa)
-except ImportError:
-    from collections import (Hashable, Iterable, Mapping,
+from collections.abc import (Hashable, Iterable, Mapping,
                              MutableMapping as MuMa)
 from itertools import chain, filterfalse
 
 
+# TODO: Update docstrings.
```
p-snft: Is this done? They seem to use nodes, now.

oemof/groupings.py
p-snft: I have seen this is consistent with other parts of the code.

oemof/network.py
p-snft: Seems to be consistent with other parts of the code.

oemof/groupings.py
gnn: Eight lines later, `Entity` is still used. So that `TODO` can unfortunately not be ticked off yet.

gnn:
This can be merged to get the new tests into v0.3.0.
I'll continue to improve coverage, but it's nice to already get some of these tests into the release cycle.

Issue #533, opened 2018-11-08
Create system emission constraint for solph constraits module


simnh:
Many people like to add emission constraints. Currently there is an example for this. I would like to add  this functionality to the constraints.py module to increase usability. 

There will be a pull request I guess...

PR #534, opened 2018-11-08
Features/solph system emissions

simnh:
The constraint has already been there but this one is an improvement for usability: 

It will now be possible to access the emissons with `m.total_emissions()`, as emissions are now an own pyomo expression object. 

  * Emissions constraints for an energy system model
  * omoef.solph.constraints.py
  

* Related issues?
#533 

* TODOS:

* [x] Test
* [ ] Example

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,13 @@
 language: python
-python:
-  - "3.4"
-  - "3.5"
-  - "3.6"
+
+matrix:
+  include:
+    - python: 3.5
+    - python: 3.6
+    - python: 3.7
+      dist: xenial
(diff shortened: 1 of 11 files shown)


simnh:
@uvchik you are right this happened kind of accidentally....

I would prefer explaining instead of moving (though I know it's the proper way)....

uvchik:
> I would prefer explaining instead of moving (though I know it's the proper way)....

:smirk: Go ahead! Just let me know when you want your final review.

simnh:
Ok I added three tests, I hope this  is fine?

Issue #431, opened 2018-01-10
Add economic calculations to oemof


simnh:
I think it would be great to just have some economic calculations based on a standard approach (e.g. NREL or VDI-standard). Maybe, if somebody has some money, we could just pay a HiWi for that...? 

(we already have the annuity, yeah!), but LCOE etc. would be nice to, and I guess if you start thinking about it, there is much more...

henhuy:
I just made calculations for LCOE, OPEX, Investment, Fuel Costs.
They depend on the new param_results from PR #416 
Once this is approved, I will start a PR for my implementations

henhuy:
Hey @simnh - I just added PR #433 to cover this issue.
Would be cool if you could review too!

uvchik:
> Would be cool if you could review too!

If you want somebody to be a reviewer you can select him within the PR (top right). The person will be noticed.

p-snft:
As #543 just covers improved annuity calculation and I do not see e.g. LCOE coming in the near future, I will postpone this feature request.

PR #543, opened 2018-11-23
Improve annuity calculation in economics tool

c-moeller:
* Describe your pull request as transparent as possible:
The annuity function in the economics tool has been supplemented by possible multiple investment and cost decreases. Both parameters  ('u', 'cost_decrease') can be set optionally. If not set, the function works as before.

* Related issues:
The function was suggested by a user on the last user meeting: https://github.com/baumconsult/gridcon/blob/master/economics_BAUM_171221.py
This is now the output of the working group from the last developer meeting.
This issue #431 might also be related.

### DIFF
--- a/oemof/tools/economics.py
+++ b/oemof/tools/economics.py
@@ -9,9 +9,6 @@ available from its original location oemof/oemof/tools/economics.py
 SPDX-License-Identifier: GPL-3.0-or-later
 """
 
-import logging
-
-
 def annuity(capex, n, wacc, u=None, cost_decrease=0):
     """Calculates the annuity of an initial investment 'capex', considering the 
     cost of capital 'wacc' during a project horizon 'n'
@@ -62,8 +59,7 @@ def annuity(capex, n, wacc, u=None, cost_decrease=0):
         
     if ((n < 1) or (wacc < 0 or wacc > 1) or (u < 1) or
             (cost_decrease < 0 or cost_decrease > 1)):
-        msg = 'Input arguments for annuity function out of bounds!'
-        logging.warning(msg)
+        raise ValueError('Input arguments for "annuity" out of bounds!')
 
     return (
         capex * (wacc*(1+wacc)**n) / ((1 + wacc)**n - 1) *
--- a/oemof/tools/economics.py
+++ b/oemof/tools/economics.py
@@ -13,16 +13,34 @@ SPDX-License-Identifier: GPL-3.0-or-later
 def annuity(capex, n, wacc, **kwargs):
     """Calculate the annuity.
 
+    In case of a single initial investment:
+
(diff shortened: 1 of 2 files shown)


uvchik:
Thank you for improving this calculations.

1. Could you write a short test similar to [this one](https://github.com/oemof/oemof/blob/dev/tests/tool_tests.py#L36).
2. I would prefer optional parameters instead of `**kwargs`. It is much more straightforward.
```python
def annuity(capex, n, wacc, n=None, cost_decrease=0)
    """...."""
    if u is None:
        u = n
```

lluismillet:
within these last commits I have tried to: improve docstrings for clarity, add tests to cover additional inputs, adequate or get closer to pep8 (don't have much experience, is it now correct? how can it still be improved?), modify kwargs as suggested by @uvchik 
I think it is good to merge :)

lluismillet:
please, could you review logging block? I have seen different criteria for logging along different files. Otherwise, any suggestion on how to improve test coverage by adding an example that checks those logging lines?

uvchik:
Thank you for the changes. Tests looks good. I just fixed a few pep8 Issues but in general everything was fine.

Sometimes we use logging warnings but the attention level is quite low. So I would rather raise a python warning or even an error if something is wrong.

Warning: The results might be unrealistic due to a strange combination of parameters but in some (rare) cases it might be okay. In that case it has to be possible to switch off the warning because a warning is annoying, if the results are okay. 

Error: No chance of getting good results with this combination of input parameters.

In most cases I prefer raising an error or doing nothing.

@lluismillet Does it help?

lluismillet:
Thank you for the corrections and the tips.
Regarding the logging, if the condition 
```
if ((n < 1) or (wacc < 0 or wacc > 1) or (u < 1) or
          (cost_decrease < 0 or cost_decrease > 1)):
```
is True, a python error will raise by itself as a division by zero will be encountered within the return clause. 
The logging I added was just an additional way to track what went wrong, in case the source of this error might not trivially be understood by the user. 
Anyway, in this case the error raising will probably be clear enough, as you suggested @uvchik.
For me, we could delete the logging block and is ready to merge.

uvchik:
It might help the user so you can keep the if statement but raise an error instead of a logging warning:

```python
if ....:
    raise ValueError("Input arguments for annuity function out of bounds!")
```

lluismillet:
Trying to adopt this error raising I figured out I do not have rights to commit on this branch anymore..

uvchik:
> Trying to adopt this error raising I figured out I do not have rights to commit on this branch anymore..

Sure? According to the team lists you should have writing access.

uvchik:
Thanks. I added a test to test the `ValueError`. I think we can merge now.

p-snft:
Looks good for me.

PR #546, opened 2018-12-04
Features/activity costs

uvchik:
Follows PR #509.


uvchik:
> I did not realize that I could change the base branch. Thanks for your help.
 
I think it is a more a hack than a typical workflow :smile: 

Thank you for adding this test. I just changed the name because it was duplicate.

For the constraint test it is very important to double check the lp-file but if everything works and @simnh is fine he could merge it from my point of view.

simnh:
Great new MILP feature, thanks for implementing!

PR #548, opened 2018-12-11
Update electrical components to work with Bus-Flow-Bus logic

simnh:
The new oemof core allows to connect Bus-Flow-Bus which is much easier for the lopf formulation,
Therefore this code has been adapted to the new core. 

Still missing Investment grouping for Flow connected between two buses @gnn should fix this for the core


ckaldemeyer:
@gnn I need this feature also quite soon! Can we continue to work on this?

gnn:
Finally managed to thoroughly investigate this and fix all corner cases. It did work before though, it's just that `add`ing the `Node`s connected by the `Flow` to the energy system BEFORE creating the `Flow` was a corner case that wasn't supported/failed. It all works now and discovering this even made me simplify the code a bit and make it more flexible.
Could you two check that things are working as expected now for you, @ckaldemeyer and @simnh?

gnn:
And oh yeah: don't you dare merge the PR yourself. If you guys approve, I'd like to update the changelog with the new features before merging. ;)

simnh:
I run the example in oemof-example and it looks fine. At least it produces the same results as the pypsa minimal lop example on which it is based. Also equations etc. look fine. Maybe we could add this example as a test including the results?

ckaldemeyer:
I just talked to @simnh . Thanks to both of you for implementing this!!

p-snft:
This new feature breaks with two former paradigms of oemof. According to https://oemof.org/libraries/, oemof.network creates "energy systems which are internally represented as a bipartite directed graph". With these changes, the graphs won't be neither directed nor bipartite.

However, I did not find any specific reasons why these design decisions have been taken. The bipartite graph was named in [doi:10.1016/j.esr.2018.07.001](dx.doi.org/10.1016/j.esr.2018.07.001)), but it hints to Hilpert et al. 2006 for more details (and I cannot find that proceedings paper). For the directed graph I would see an advantage if it was loop-free, as this would allow to find a topological sorting that could then be used to speed up the minimization problem.

simnh:
Actually, the bipartite graph concept will allow for non-negative weight  on this directed graph edge as far as I understood @gnn. However, the +/- convention will be given by the directed edge. So from this perspective it does not break the concept. 

If it does break, we should adapt the concept. Because there is no reason to keep a concept that does not suit our needs....

p-snft:
As far as I understand, the bipartite graph concept was implemented by the "bus-flow-component" layout. Now, that you allow buses to be directly connected, it is no longer bipartite. However, if there is no advantage in having it bipartite, we should relax that concept.

simnh:
To be honest, the concept has no always have one common understanding among the developers. Actually, I can not explain the different views on that here, its implications and thus also implementation. All I remember is, that this was the reason, why Flow/Edge is also a Node in the graph. Which in it self is kind of confusing...

For me the only thing that should be clear is that the description in documentation matches the implementation. For the rest, lets just relax / adapt the concept as far as it is necessary...

simnh:
@gnn What about this one? Is there anything left to be done (except finishing the discussion on the core concept maybe?)

p-snft:
To be precise: As the electrical components utilizing the bus-flow-bus logic are still marked as custom, I do not see any strong reason why this should not be merged, whatsoever will be decided on the core concept.

gnn:
I changed everything I wanted to change and from my POV this PR can (and should) be merged prior to the v0.3.0 release.

p-snft:
@gnn: Do you understand why the push test is failing?

PR #549, opened 2018-12-13
Increase test coverage of solph.models module

uvchik:
There are some special conditions in this module and we should remove them or test them.

Furthermore `custom.py` is removed from the test coverage check, because we announced that it is not necessary to write tests for the custom components but we should warn users about this.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,4 +1,4 @@
 [run]
 omit =
     *datapackage*
-    custom.py
+    *custom*
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,2 +1,4 @@
 [run]
-omit = *datapackage*
+omit =
+    *datapackage*
+    custom.py
--- a/doc/whatsnew/v0-3-0.rst
+++ b/doc/whatsnew/v0-3-0.rst
@@ -40,7 +40,11 @@ Bug fixes
 Testing
 #######
 
-* something
+* Automatic test coverage control was implemented. Now a PR will not be
+    accepted if it decreases the test coverage.
+* Test coverage was increased to over 90%. A badge was added to the
+    `oemof github page <https://github.com/oemof/oemof>`_ that shows the
+    actual test coverage.
 
 Other changes
 #############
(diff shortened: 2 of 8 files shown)


uvchik:
First thing to discuss in this context:

At the moment a `timeincrement` of one is set if the time increment wasn't defined correctly. If I have a time increment of 15 minutes but made a mistake the time increment would be set to one in the background and all I get is a logging warning.

To leave the opportunity to create your own time increment I would add the possibility to pass it directly to the model, otherwise solph tries to get it from the `timeindex` and if this is not possible an error is raised. If you want we could add our own error message but an error should be raised in my opinion.

ACTUAL VERSION:
```python
    def __init__(self, energysystem, **kwargs):
        super().__init__()
        self.name = kwargs.get('name', type(self).__name__)
        self.es = energysystem
        try:
            self.timeincrement = sequence(self.es.timeindex.freq.nanos / 3.6e12)
        except AttributeError:
            logging.warning(
                'Could not get timeincrement from pd.DateTimeIndex! ' +
                'To avoid this warning, make sure the `freq` attribute of ' +
                'your timeindex is not None. Setting timeincrement to 1...')
            self.timeincrement = sequence(1)
```

MY PROPOSAL:
```python
    def __init__(self, energysystem, **kwargs):
        super().__init__()
        self.name = kwargs.get('name', type(self).__name__)
        self.es = energysystem
        self.timeincrement = kwargs.get(timeincrement, None)

        If self.timeincrement is None:
            self.timeincrement = sequence(self.es.timeindex.freq.nanos / 3.6e12)
```

The if block could also look like this to make it more clear for the user.
```python
        If self.timeincrement is None:
            try:
                self.timeincrement = sequence(self.es.timeindex.freq.nanos / 3.6e12)
            except AttributeError:
                raise AttributeError(our_message)
```

uvchik:
As nobody commented the idea above I pushed a concrete proposal. If nobody objects I will merge.

uvchik:
@oemof/oemof-solph See commit 2f6430 to review the changes on the time increment.

uvchik:
@simnh Do have any objections or may just approve this?

PR #551, opened 2019-01-12
Remove __setitem__ because this is not intended

uvchik:
Setting single values of a sequence leads to unwanted behaviour (e.g. never ending loop). Therefore, I think it is better to allow only the full replacement of a sequence but not the change of single values.

@oemof/oemof-developer-group : Does anybody needs this method???


henhuy:
I think it is okay to delete it, as the use case is very small or even not existing!
It was painfully to integrate it and, as you detected, it caused a bug - so maybe it is the best to delete it right away. 

Whole refactoring of Sequence from UserList to numpy.array is still pending - no time, sorry.

PR #552, opened 2019-01-14
Revise results handling of the model

uvchik:
At the moment the results handling is confusing:

1. Is it still necessary to call  `self.solutions.load_from(results)`?
2. I think we should distinguish just two or three termination conditions.
      1. Everything is fine. Results can be written
      2. Something not optimal, but there are some results that can be processed (Does this case exist?)
      3. No solution -> no results
3. What shall we do with the results?
    1. Leave them where they are and let the user call `results = outputlib.processing.results(model)`
    2. Write them to the EnergySystem results attribute.
    3. Return them (`results = model.solve(...)`) - see (4).
4. What shall we do with the meta results return by pyomos `solve` method?
    1. Leave them where they are an let the user call `results = outputlib.processing.meta_results(model)`
    2. Write them to an EnergySystem attribute?
    3. Return them (`meta_results = model.solve(...)`) - see (3).
5. I can create small test models for 2.I. and 2.III. but I have no experience with 2.II.. Can anybody help.

@oemof/oemof-developer-group I need a feedback:exclamation:

### DIFF
--- a/oemof/solph/models.py
+++ b/oemof/solph/models.py
@@ -184,8 +184,8 @@ class BaseModel(po.ConcreteModel):
         results = opt.solve(self, **solve_kwargs)
 
         status = results["Solver"][0]["Status"].key
-        termination_condition = \
-            results["Solver"][0]["Termination condition"].key
+        termination_condition = (
+            results["Solver"][0]["Termination condition"].key)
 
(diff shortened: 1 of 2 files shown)


uvchik:
I will not be able to prepare this PR for the v0.3.0 release so I shifted it to v0.3.1.

PR #558, opened 2019-01-24
Allow conversion factor of zero for transformer

henhuy:
Adapted constraint for comparison of input and output flows with
conversion factors to allow zero values.

By now, setting a conversion factor of "0" to a transformer leads to an zero-division-error due to following part in constraints (`oemof.solph.blocks.Transformer._create()._input_output_relation()`:
```
lhs = (m.flow[i, n, t] / n.conversion_factors[i][t])
rhs = (m.flow[n, o, t] / n.conversion_factors[o][t])
...
block.relation.add((n, i, o, t), (lhs == rhs))
```
I simply switched conversion factors and avoid the zero-division-error!
```
lhs = (m.flow[i, n, t] * n.conversion_factors[o][t])
rhs = (m.flow[n, o, t] * n.conversion_factors[i][t])
...
block.relation.add((n, i, o, t), (lhs == rhs))
```
Results stay the same, only lp-file tests are failing.
I wanted to discuss this before I adapt those tests.
Any drawbacks?

### DIFF
--- a/oemof/solph/blocks.py
+++ b/oemof/solph/blocks.py
@@ -553,10 +553,10 @@ class Transformer(SimpleBlock):
                     for o in out_flows[n]:
                         for i in in_flows[n]:
                             try:
-                                lhs = (m.flow[i, n, t] /
-                                       n.conversion_factors[i][t])
-                                rhs = (m.flow[n, o, t] /
+                                lhs = (m.flow[i, n, t] *
                                        n.conversion_factors[o][t])
+                                rhs = (m.flow[n, o, t] *
+                                       n.conversion_factors[i][t])
                             except ValueError:
                                 raise ValueError(
                                     "Error in constraint creation",


uvchik:
Why do you want to set a `conversion_factor` to zero? What you get is an unlimited sink, so you better use a Sink object to get this behaviour. If you want to switch off the Transformer with parameters you should set the `nominal_value` to zero.

p-snft:
I see both arguments. We should either allow zero or explicitly reject non-positive conversion factors. (As we are talking about energy, they should not be negative.) I would consider it bad style to just use division by zero to validate input.

uvchik:
It was not a design decision to use the  'division by zero' error it just came out like this.

If you want to change it you have to change the docstring, too and provide a test that would fail now and do not fail with the changes.

I think for most users this is a disservice because they will not receive an error if they accidentality defined a wrong conversion factor. We could add this case to our planed checking function.

p-snft:
Of course, it was not designed to use the division by zero error. However, if we now leave it untouched because of that "implicit check", we would now use it that way.

As @uvchik suggest, this is something for the checking function. Until then, I would consider @henhuy's suggestion better than the "implicit check by different failure".

uvchik:
I am not totally convinced but I will not object.

Shall we set the milestone to v0.3.1 because @gnn will present the checker function in June and we could directly add this check.

uvchik:
We do have a `SuspiciousUserWarning`  now. So you can change the equations and the docs and add a warning. After that you can ask for a review again. If a zero is allowed, it will not break the API so we can integrate it with the next minor release. I changed the milestone accordingly.

henhuy:
Okay - I will look into it. thx!

p-snft:
I desperately needed this...

uvchik:
I think at least the docstring has to be adapted. I would also add a test with a conversion factor of zero.

I am not sure if we decided to add a warning if there is just one input/output and a conversion_factor of zero.

PR #564, opened 2019-02-05
Fix/issue #563

### LINKED ISSUE #563: Time increment not needed in Bus balance
The Bus balance looks like this:

bildschirmfoto_2019-02-05_17-58-22

While τ is the time increment. Within one time step the time increment is constant and therefore it can be eliminated from the equation. Or did I forget a special case?

It does not harm anybody. It is just a useless constant on both sides .

uvchik:
Fixes #563

### DIFF
--- a/oemof/solph/blocks.py
+++ b/oemof/solph/blocks.py
@@ -448,8 +448,8 @@ class Bus(SimpleBlock):
 
     Bus balance  :attr:`om.Bus.balance[i, o, t]`
       .. math::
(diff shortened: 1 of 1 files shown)


uvchik:
@simnh Any objections?

uvchik:
@simnh Any objections?

uvchik:
As @simnh did not raise objections in the last four month I guess he is fine with the changes.

I would like to keep the so called "four-eyes-principal". Could one of you, @FranziPl , @p-snft approve this PR if you agree (see #563). I will merge afterwards.

Issue #566, opened 2019-02-12
Error in docstring: Equation on summed_max


jnnr:
The docstring does not represent the equation that is implemented a few lines later in https://github.com/oemof/oemof/blob/dev/oemof/solph/blocks.py
Docstring:
>
    Flow max sum :attr:`om.Flow.summed_max[i, o]`
      .. math::
        \sum_t flow(i, o, t) \cdot \tau \leq summed\_max(i, o), \\
        \forall (i, o) \in \textrm{SUMMED\_MAX\_FLOWS}.

This is the [rendered version](https://oemof.readthedocs.io/en/latest/api/oemof.solph.html#oemof.solph.blocks.Flow)
Equation:
>
        def _flow_summed_max_rule(model):
            """Rule definition for build action of max. sum flow constraint.
            """
            for inp, out in self.SUMMED_MAX_FLOWS:
                lhs = sum(m.flow[inp, out, ts] * m.timeincrement[ts]
                          for ts in m.TIMESTEPS)
                rhs = (m.flows[inp, out].summed_max *
                       m.flows[inp, out].nominal_value)
                self.summed_max.add((inp, out), lhs <= rhs)
        self.summed_max = Constraint(self.SUMMED_MAX_FLOWS, noruleinit=True)
        self.summed_max_build = BuildAction(rule=_flow_summed_max_rule)

Same for `summed_min`. Can be easily fixed by adding the factor `nominal_value` in the docstrings.

uvchik:
@oemof/oemof-assistants : Could be fixed in one PR together with #565.

ajimenezUCLA:
I would like to be assigned to this, I have a question about the equation. nominal_value should be multiplied to tau? or summed_max(i,o)
I'm trying to understand the equation.

p-snft:
I read "summed max" as "maximum full-load hours": The total integrated energy flow (lhs) has to be lower or equal the maximum full load hours times the installed power ("nominal value").

PS: This specific issue has been fixed by my commit above. It is not merged, yet, so the ticket is still open.

Issue #565, opened 2019-02-07
Wrong equation in documentation


uvchik:
In the second equation the P_el should be `greater equal` not `lower equal` than the right hand side.

https://oemof.readthedocs.io/en/latest/api/oemof.solph.html#oemof.solph.components.ExtractionTurbineCHPBlock

In this context we could check the rest :-)

@oemof/oemof-assistants Does anybody wants to help?

ajimenezUCLA:
I would like to be assigned this, I need to figure out how to use docstring though.

p-snft:
As nobody joins the search for errors in equations: It would be useful to perform the check based on the updated docstring template (see #565). When the equations are rewritten to a more readable form, it is easy to check them right away.

Thus, I am going to perform the pull request #568.

Issue #544, opened 2018-11-27
Docstring template


jnnr:
@FranziPl and @jnnr have extended the docu on GenericCHP and ExtractionTurbineCHP. 
The aim was the following:
* show the constraints of the components in the [oemof-solph](https://oemof.readthedocs.io/en/stable/oemof_solph.html#) part of the docu 
* use mathematical symbols that can be understood by the energy community
* show a table that translates between mathematical symbol, meaning and variable name in the code.

These parts have been written to the docstring of the respective Block class of the component 

* [ExtractionTurbineCHPBlock](https://oemof.readthedocs.io/en/latest/_modules/oemof/solph/components.html#ExtractionTurbineCHPBlock)
* [GenericCHPBlock](https://oemof.readthedocs.io/en/latest/_modules/oemof/solph/components.html#GenericCHPBlock)

and are included in the oemof-solph docu. 

The purpose of this issue: To keep things consistent, we want to write down a template for this docstring to make it possible to have the same structure for all components.

jnnr:
class <Some>Block(SimpleBlock):
r"""
    :class:

    **The following sets are created:** 

    **The following two constraints are created:**

    .. _<SomeBlock>-equations:

        .. math::
            <Equations>
    

    ========================= ======================== =========
    symbol                    explanation              attribute
    ========================= ======================== =========

   
    ========================= ======================== =========

"""

uvchik:
The main point to discuss is the structure of the equations.

The old equations were (more or less) mathematically exact. In the equations of the ExtractionTurbine the explicit time dependency was removed and the sets.

1. I like the renaming because it is easier to understand the equations.

2. I would suggest to keep the explicite time dependency: `a(t) + b = c(t)` instead of `a+b = c
(see [ExtractionTurbine](https://oemof.readthedocs.io/en/latest/api/oemof.solph.html#oemof.solph.components.ExtractionTurbineCHPBlock) for a "negative" example and [GenericCHP](https://oemof.readthedocs.io/en/latest/api/oemof.solph.html#oemof.solph.components.GenericCHPBlock) for a better example.

3. I would suggest to remove the sets because they are not interesting. I would rather list the equations that will be created every time and then add a line: The following equations will be build if the `examp` attribute is set: _equations_ and so on.

uvchik:
Would be great to have consistent docstrings for v0.3.0.

@oemof/oemof-assistants @oemof/documentation This is not to difficult and good issue for new developer.

[...]

PR #569, opened 2019-02-18
Fix/equations in documentation

p-snft:
Fix equations in documentation listed in issues #565 and #566. As #565 also asks to search further incorrect equations, the merge should not be performed immediately.

### DIFF
--- a/oemof/solph/blocks.py
+++ b/oemof/solph/blocks.py
@@ -49,12 +49,14 @@ class Flow(SimpleBlock):
 
     Flow max sum :attr:`om.Flow.summed_max[i, o]`
       .. math::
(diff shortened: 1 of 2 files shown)


uvchik:
Thanks, @p-snft for bringing everything in line.

I added the branch to the documentation, to make it easier to check.

https://oemof.readthedocs.io/en/fix-equations_in_documentation/

In my opinion it is also related to #544, because the docstrings of the solph classes are quite different (Flow, GenericStorage, GenericCHP, ExtractionTurbineCHP....). I think we should discuss our favoured variant in that issue and adapt all docstrings (step by step) according to the outcome.

Nevertheless, it might be better to open a new PR for this kind of adaptions, because the fixes in this PR are bug fixes and not layout improvements.

I did not find anything else, so i would merge.

PR #581, opened 2019-04-24
Fix/refine_datapackage

p-snft:
Just followed the advice of some lint tools (including [LGTM](https://lgtm.com/projects/g/oemof/oemof/snapshot/821cc53a63d19b86662ae3017cd66065c5fdcd80/files/oemof/tools/datapackage.py?sort=name&dir=ASC&mode=heatmap)).

### DIFF
--- a/oemof/tools/datapackage.py
+++ b/oemof/tools/datapackage.py
@@ -52,7 +52,8 @@ def remap(attributes, translations, target_class):
     for c in mro():
         if c in translations:
             break
-    return {translations.get(c, {}).get(k, k): v for k, v in attributes.items()}
+    return {
+        translations.get(c, {}).get(k, k): v for k, v in attributes.items()}
 
 
 def sequences(r, timeindices=None):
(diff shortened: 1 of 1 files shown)


p-snft:
@gnn: Were there a specific reason not to have `HSN.__eq__` defined or to have an unreachable return in `datapackage.raisestatement`? (The return even stated that it is unreachable...)

p-snft:
Except for the explicit definition of `__eq__`, this is purely cosmetic.

uvchik:
I just fixed some pep issues to be in line with our developer rules.

I think we can merge if @gnn is fine with the `__eq__` change.

uvchik:
I would avoid name-shadowing because I had bad experience with that.

p-snft:
@uvchik As stated in the commit, `__eq__` is defined because there is a `__hash__` function. By convention, one should not be defined without the other.

gnn:
Thanks for your suggestions and contributions. As the purpose of this module was pretty specific, we (@simnh and me) actually decided a while ago to move the code to a [dedicated package](https://github.com/oemof/oemof-tabular) dealing with tabular data sources in general and tabular datapackages specifically as the first use case. That is why this PR will not be merged, because the file in question no longer exists in the current development branch.
Your contributions where helpful though as some of them have been addressed in the new package, albeit sometimes differently than what you suggested.
Concerning the question of the missing `HSN.__eq__` implementation: yes, not implementing that method was intentional. For once, Python already supplies a default `__eq__` implementation on all classes, doing exactly what got implemented in this PR: defining equality by `id`. User defined classes usually also get a matching `__hash__` implementation. The reason I have to explicitly provide one for `HSN` is the fact that `HSN` subclasses a built-in Python class which doesn't have one by default. But it has the `__eq__` implementation. In short: `HSN`'s sole purpose is to add a trivial `__hash__` method and nothing else to `types.SimpleNamespace`, which is why it's the only thing written in the class body.
I should probably make that equally explicit in the documentation.

p-snft:
Obsolete, since datapackage was removed.

Issue #573, opened 2019-03-06
Timedependent summed_max for flows


simnh:
Hey I would like to change the summed_max from a scalar to a kind of dictionary:

Concated subperiods values would be all timesteps of the model
```

subperiods = {p1: [0, 1, 2, ...23], p2: [24, 26, ..., 47]}

summed_max = {'p1': 2000, 'p2': 5000}
```

What do you think of this? Is this still possible for version 0.0.3?

p-snft:
I like the idea. How long would it take you to implement it?

simnh:
Well, it depends on how we implement it, because we need to provide the information of subperiods (I would suggest in the energy system object). I could make a pull request with a first drafted idea and we can discuss there. 

Probably it makes sense to just make the 0.0.3 release and have shorter release cycles afterwards.

p-snft:
I have had a closer look into this and I now think it is a good idea to plan this for v0.4. I now see the various possible implementation variants. 


PS: Does it make sense to have a pull request at this early point in time?

simnh:
I think using a pull request for the discussion is better than the issue or both ...

busiing:
Self-assignment due to oemof dev meeting - issue auction

PR #582, opened 2019-04-26
Merge features/rework_startups

p-snft:
* Adds a possibility for maximum startups/shutdowns for non-convex flows
  * API: parameters `maximum_startups` and `maximum_shutdowns` of NonConvex
  * Unit test (and LP file) are also added.
* Naming of sets has been should now been consistent between docstrings and implementation. (Underscores in variable names are a pain if used in LaTeX math strings. We should avoid this for the future.)


p-snft:
The maximum number of startups feature can be used for various things, e.g. to fulfil the guidelines of typical maintenance contracts for CHPs. I will provide an example as soon as the feature is merged.

p-snft:
This also effects #573, which parallelly applies changes to the same fractions of code.

Issue #428, opened 2018-01-09
Add mathematical description to docstrings


ckaldemeyer:
The following (custom) components do not have their mathematical description listed in their docstrings yet:

* [x] GenericCHP: https://github.com/oemof/oemof/blob/dev/oemof/solph/components.py#L498
* [x] GenericCAES: custom.py (to be added when merged)
* OffsetTransformer: https://github.com/oemof/oemof/blob/dev/oemof/solph/custom.py#L361  (#522 already deals with the `OffsetTransformer` - including the math descricption)

It should/could be completed easily until the next release by looking into the referenced papers and transferring the equations.

jnnr:
The mathematical documentation of the named components is still missing. Please give us some formulas! :) I added checkboxes to Cords list.

ckaldemeyer:
> The mathematical documentation of the named components is still missing. Please give us some formulas! :) I added checkboxes to Cords list.

I do not have time for this until September and the references are already contained. If anyone has spare time, feel free to transfer the equations (which is only work and not much thinking) ;)

uvchik:
> I do not have time for this until September and the references are already contained.

@ckaldemeyer The next regular release is scheduled for October :smile:

jakob-wo:
I marked the GenericCHP as "done". As far as I can tell, PR #525 (already merged into dev) contributed a detailed explanation of the equations.

uvchik:
You could talk to @joroeder about the docstring template for all block classes and check if the docstring of the `GenericCHPBlock` is in line with the new conventions.

jakob-wo:
jip! I will compare status quo of the `GenericCHPBlock` docstring with the recommondations made in Issue #544 in the next days - Anything regarding the docstring conventions shall be commented and added there, so that I can take it into account.

jakob-wo:
In the Branch "feature/docu_add_math_desriction_caes" I am going to implement the caes equations from the paper mentioned in `GenericCAES.` I would like to see if the formatting works as expected - How can I make the Branch a selectable docu in readthedocs?

joroeder:
For checking the formatting, you can create a local build - I just learned at dev meeting :)

```pip install sphinx```

```sphinx-build -b html "...\oemof\doc" ".\your\local\path\```

For checking your branch after committing, your branch needs to be added the readthedocs and build there.

jakob-wo:
[...]

Issue #588, opened 2019-05-22
Restructure components


p-snft:
The name "custom components" implies that also third party custom components should go there. However, it is perfectly possible to define custom components outside the oemof directory in the project using oemof.

Now, I think of, "custom components" somewhat as of the "staging" part of the Linux kernel. The code is already part of it, but it is not fully tested or documented. At some point in time, it will eventually be become core functionality. What do you think?

Update: Also, as the number of components grows, it might be useful to have some more structure, eventually even a document comparing different models for one (real world) component.

jakob-wo:
I quote myself from another Issue because it belongs here:

> I see solph.custon as a place not only for "not-100%-tested-and-documented-Components" but also for highly specified Components that have no generic formulation or application.

That means, I do not see solph.custom as something like "staging" even though in some cases it can be used for staging.
I prefere to stay with "solph.custom".

p-snft:
I see. Probably this is just one aspect of a broader topic. I will tag it for the v0.4 release and bring it to the agenda at the 2019 spring dev meeting.

jakob-wo:
@uvchik suggested a web-conference in Issue #603 that is  related to this topic:
> Shall we make a web-conference about "Where to place advanced components?

Would you like to join us?

uvchik:
I agree in both. Changing the name and dividing the modules. But we will not be able to discuss it before the v0.4.0 release. But we could change it with v0.4.1 and add a deprecate warning. We could remove the old name with v0.5.0.

p-snft:
As a follow up: I am currently reworking the Link, which I see as a simple component in staging. Could we sort the others? (Probably, ElectricalLine is more of a complex component rather than staging. Have we a name for this kind?)

p-snft:
Taking a broader view (chatting with @uvchik), I was convinced that this structure is caused by the fact that oemof.solph is historically grown. Not knowing the exact background, it can be confusing. Thus, I propose the following:

1. There should be a submodule (directory) `components` containing all components.
2. Inside that directory, _all_ components are placed in individual modules (python files). E.g., there will be `generic_storage.py`, but also `transformer.py`. (The latter is a component but currently placed in `solph.network`).
3. These modules contain both, the solph  class and the Pyomo block class. (This, i.a. means that the `TransformerBlock` will move from `solph.blocks` to `solph.components.transformer`.)
[...]

PR #589, opened 2019-05-23
Features/docu add math description caes

jakob-wo:
* This PR adds a mathematical description to the docstring of `custom.GenericCAESBlock`
  * It Includes a list of created variables, of created constraints and a table

* Related issues: This PR fulfils a task of #428 and follows the docstring template recommedation made in #544

* It bases on the paper:
    Kaldemeyer, C.; Boysen, C.; Tuschy, I.
    A Generic Formulation of Compressed Air Energy Storage as
    Mixed Integer Linear Program – Unit Commitment of Specific
    Technical Concepts in Arbitrary Market Environments
    Materials Today: Proceedings 00 (2018)


oemof/solph/custom.py
```diff
@@ -412,7 +412,225 @@ def constraint_group(self):
 
 
 class GenericCAESBlock(SimpleBlock):
-    """Block for nodes of class:`.GenericCAES`."""
+    r"""Block for nodes of class:`.GenericCAES`.
+
+    **The following variables are created:**
```
p-snft: This is already part of the table. I think, it is not needed here.

p-snft: This is already part of the table. I think, it is not needed here.

jakob-wo:
Thanks for your feedback!

I will try to reply to every aspect you mentioned:

1) List of Variables: I agree on @p-snft comment. The list is a partial repetition of the table below and therfore we may leave it out in the docstring. But in that case I suggest that the Table, containing variables (V) and parameters (P), shall always be ordered according to the the "type", i.e. V or P. So one can get a quick overview of the created variables. Here, in case of teh CAES this is already done and I will remove the List of variables.

2) Moving the code to components: In my opinion we shout leave it in custom. For two reasons: a) The component and its code as it is now is not well enough documented and an example is missing.  I could go on improving it but would need help from @ckaldemeyer as a few aspects are still not clear to me. b) I see solph.custon as a place not only for "not-100%-tested-and-documented-Components" but also for highly specified Components that have no generic formulation or application. And I see the CAES as a highly specified component that fits very well in custom. Parts of this topic are related to #588 and we shall continue the discussion there, I suggest.

3) Size of variables in the table: Yes, that is due to the size of the indices. I tried to pre-define the column-width but it did not work (I did not spend much time on this). If anyone knows a solution or work-around... let me know, please.

4) "intersection" instead of "intersept": Yes, you are right! I will change that. That was copy/paste from the paper, so @ckaldemeyer you might want to correct that typo in your paper as well ;-)

5) paper is necessary to use the component: Yes, I agree on that, too. For me it was just adding the constraints and variables and I would prefere to finish and merge that part first.

jakob-wo:
> 
> There are so many off-set linear equations. Is it possible to have a short summary alread in the table at the parameters of all the linear equations?
>

I did not find a nice and consistent way to avoid naming all paramters in the table. I left it as it was before but I am open to suggestions.

PR #592, opened 2019-05-27
Custom component: Piecewise linear transformer

jnnr:
This PR implements a custom component called PiecewiseLinearTransformer.
  * The new component is a 1x1-Transformer whose conversion factor depends on the input flow. It allows to approximate nonlinear conversion relations as piecewise functions.
  * Location: As of now, it is in custom.py. With tests, docu and enough generality, it could go to components.py later.
Related: [#403](https://github.com/oemof/oemof/pull/403) on piecewise linear costs functions.

It can be used like this, passing breakpoints, a conversion function and the piecewise representation that should be used internally:

```python
conv_func = lambda x: 0.01 * x**2
in_breakpoints = np.arange(0, 150, 30)

pwltf = solph.custom.PiecewiseLinearTransformer(
    label='pwltf',
    inputs={b_gas: solph.Flow(
    nominal_value=220,
    variable_costs=1)},
    outputs={b_el: solph.Flow()},
    in_breakpoints=in_breakpoints,
    conversion_function=conv_func,
    pw_repn='CC')
```

Internally, the new component uses pyomos own Piecewise() constraints. To feed it domain and range variables, new variables are introduced and equated with the incoming and outgoing flows. Maybe there is a more clever way to do it. In the future, piecewise linear nxn-Transformers could be a nice feature.

TODO:
* [x] Check compatibility with outputlib.
  * [x] Solve problem with extra indices: The auxiliary variables are ignored in the postprocessing
         Still, there are the extra variables inflow and outflow. It would be even cleaner if they would not appear. With a different implementation of `PiecewiseLinearTransformerBlock`, these variables could be possibly avoided altogether.
* [x] Write test.
  * [x] rewrite as constraint test.
* [x] make test_script an example: https://github.com/oemof/oemof-examples/pull/67

Note: Keep in mind that some solvers in combination with some representations can lead to wrong results: https://github.com/Pyomo/pyomo/issues/216


oemof/solph/custom.py
```diff
@@ -797,3 +798,121 @@ def _relation_rule(block, n, t):
...
         self.relation = Constraint(self.OFFSETTRANSFORMERS, m.TIMESTEPS,
                                    rule=_relation_rule)
+
+class PiecewiseLinearTransformer(Transformer):
+    """An object with one inputs and one output.
+
+    Parameters
+    ----------
+    x : list
+    y : list
```
p-snft: The parameters are not documented properly:
* `x` might be `in_breakpoints`
* `y` might be `out_breakpoints`
* `conversion_function` remains fully unclear
* `pw_repn` remains fully unclear

oemof/solph/custom.py
```diff
@@ -797,3 +798,121 @@ def _relation_rule(block, n, t):
...
+            return None
+
+        m = self.parent_block()
+
+        self.PWLINEARTRANSFORMERS = Set(initialize=[n for n in group])
+
+        # TODO:
+        # pw_repn ?
+
+        pw_repn = 'CC'
```
p-snft: You defined `self.pw_repn` above but do not use it, here.

jnnr:
Thanks for your comments, @p-snft! I adapted the code and added a test_script. Maybe some more review can help to decide if this can be merged?

oemof/outputlib/processing.py
```diff
@@ -86,6 +86,7 @@ def create_dataframe(om):
     # on which dimension the variable/parameter has (scalar/sequence).
     # columns for the oemof tuple and timestep are created
     df['oemof_tuple'] = df['pyomo_tuple'].map(get_tuple)
+    df = df[df['oemof_tuple'].map(lambda x: x!=None)]
```
jnnr: One thing I am not completely sure about is this little hack that I used to avoid an error in the outputlib. The constraints produced by pyomo's Piecewise are labelled differently in the LP-file than oemof normally expects. This line excludes those, but I am not sure if this is the best way and if it does not interfere somewhere.

uvchik:
Is this urgent? I think I may not be able to review it this week.

I think we should focus on constraint test and other test besides script tests. Script-tests are slow. If the constraints are build correctly the components will behave correctly and it is not necessary to prove that by solving the problem.

It would be better to provide an example with plots to show how the new component works or to show the differences to existing components.

jnnr:
Thanks for asking, @uvchik! There is nothing speaking against putting this to v0.3.1, as it is not extremely urgent and as there is no API-change involved.

This gives me more time to think about the test. Thanks for your feedback. An example with plots will be provided as well.

jnnr:
As a reference, this Thesis describes the different approaches to piecewise linearisation offered by pyomo ('CC', 'DCC', 'INC'): https://kops.uni-konstanz.de/bitstream/handle/123456789/40493/Bernreuther_2-1gxw6ja5feelr3.pdf?sequence=3&isAllowed=y

jnnr:
The following problem arises in the outputlib when using pyomo's Piecewise:

Where the blockvars for all other constraints have this form:

```
flow[gas,pwltf,t]
flow[electricity,demand,t]
PiecewiseLinearTransformerBlock.inflow[pwltf,t]
```

the blockvars produced by pyomo's Piecewise look like this:

```
PiecewiseLinearTransformerBlock.piecewise[pwltf,t].CC_lambda[i]
PiecewiseLinearTransformerBlock.piecewise[pwltf,t].CC_bin_y[i]
```

where i is the index for the different segments in the piecewise linearised function. Apparently, Piecewise creates child variables
that carry this new piecewise index.

This gives wrong results in `outputlib.processing.create_dataframe() `when creating the pyomo tuple/oemof tuple.
The index i is written instead of` [pwlft,t]`, which it should be according to what usually happens.

TODO: Make the outputlib understand how to process the index relevant for the results dataframe, i.e. the one describing the component and time.

uvchik:
> TODO: Make the outputlib understand how to process the index relevant for the results dataframe, i.e. the one describing the component and time.

See issue #630.

jnnr:
I had another look at the PiecewiseLinearTransformer and the problem with the outputlib. I think I could solve it, but it maybe not the nicest way to handle it. 

The thing is that pyomo's Piecewise is a block created within `_create` of the PiecewiseLinearTransformerBlock. Therefore, it is a child block of the PiecewiseLinearTransformerBlock. When constructing the `var_dict` in the outputlib (https://github.com/oemof/oemof/blob/dev/oemof/outputlib/processing.py#L77), the block name and variable name are thus the ones of the child block of PiecewiseLinearTransformer. 

This made me think about why not to create the Piecewise block in `_create`, but inherit from it. However, this did not work out, because Piecewise expects to be initialized with pyomo variables and other keywords that are not defined when calling __init__(), but only later, when calling `_create()`. 

I did not find a solution for that, so I stayed with the previous implementation of the component and adapted the outputlib (the last two commits). 

I read through #630 and tested the adaption in #623. The change of `get_tuple()` in the latter seems to be compatible with the changes here.

What do you think, @oemof/oemof-solph?

uvchik:
We had to fix the conflicts against the new `dev` and to fix errors of the new tests.

Okay this seems to work more or less now, so that we can return to the content :smile: 

There is no need to hurry because we could add this component later as well, because there is no API break.

uvchik:
Did you test your code on big energy systems? I think it could be slower than the original code.

Therefore I would prefer a way, that is more or less the same in the "normal" case and extended in case of a PicewiseLinearTransformer.

Please see commit 1b80bd6. The code is faster in case of a present PicewiseLinearTransformer and not slower than the original code if no PicewiseLinearTransformer is present. At least I could not find any difference on very large system (processing time 44 sec).

jnnr:
Thanks for updating to the new oemof-solph, @uvchik! As you say, this PR does not break the API, so it is ok to merge it later. As I remember, the problem with the output of the results is not solved completely. At least, another look at it would be necessary. As I will not have time to work on it for the next 2-3 weeks, I propose to change the milestone.

jnnr:
Together with @stefansc1 we updated this PR which has been resting for a while. We would like to bring this to a merge soon as the feature is needed and actually already used in some models. What did we do?

* We could keep the auxiliary variables from appearing in the results by a0cd1f6.
* We added a PR for an example in oemof-examples https://github.com/oemof/oemof-examples/pull/67
* We resolved merge conflicts
* We tested locally: Pytest passes, however I am not sure how to interpret the amount of feedback I get from tox or from CI tools like appveyor. If there is more to do, let us know!

The internal variables `inflow` and `outflow` in the PiecewiseLinearBlock are still there and appear in the results. In a future, more clever implementation, they could become obsolete.

@oemof/oemof-solph: Thanks for your help here so far, especially @uvchik and @p-snft. Please have another review, if you want to help with this feature!

jnnr:
Hi @oemof/oemof-solph and happy new year! Do you have any feedback on this one? We hope to be able to merge this soon, as it is a long-awaited feature for some of us.

p-snft (commented):
Despite two little things, this looks good. However, I do not block this by demanding the changes as I believe that you know what to do.

1. You import a private class from Pyomo (see code comment). This should be avoided.
2. There is an issue with the CI. (I can't help you with that, sorry.)

src/oemof/solph/processing.py
```diff
@@ -19,6 +19,7 @@
 
 import pandas as pd
 from oemof.network.network import Node
+from pyomo.core.base.piecewise import _PiecewiseData
```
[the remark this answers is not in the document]
jnnr: Thanks for the feedback! We will try to adapt this.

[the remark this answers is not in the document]
jnnr: Resolved with #741

jnnr:
Thanks everybody for your help and contribution! Especially, @p-snft, @uvchik and @stefansc1. Great that we can finally merge!

jnnr:
Realized just after merging that the entry in the what's new is missing. Will have to do that separately.

Issue #560, opened 2019-01-29
Improve handling of kwargs


p-snft:
Update: There seem to be four degrees of possible kwarg handling:

1. Use wild card **kwargs for all key word arguments. This is currently done, e.g. in [Flow](https://github.com/oemof/oemof/blob/12a30e28f19373033c01fb9297a7de0a36568f91/oemof/solph/blocks.py#L82). Currently, this is the most common way in oemof.

2. Define known kwargs but set empty defaults. Technically, this is the same API as in 1. but it improves readability and comfort. This way was suggested [later in this ticket](https://github.com/oemof/oemof/issues/560#issuecomment-463573255). It also reduces the likeliness of typos.

3. Enforce values for obligatory quantities. This would also improve readability, and would also solve copy and paste errors, typos, etc. It would require programmers to make explicit when obligatory parameters should be left uninitialized (e.g. `Sink(inputs={})`). It is already implemented in [ExtractionTurbineCHP](https://github.com/oemof/oemof/blob/12a30e28f19373033c01fb9297a7de0a36568f91/oemof/solph/components.py#L1070) but quite uncommon in oemof.

4. Disallow faulty kwargs (like `outputs` for a `Sink`, or those including typos). This option would either cost flexibility, require some "expert mode" where strict enforcement is turned off, or need explicit checks. (Option 4 is what I originally proposed here.)

**Original text:**
I would suggest to raise errors if unknown kwargs are passed to functions. I believe this would make it greatly easier to debug code using oemof and improve the usability.

uvchik:
Could you provide an example? 

Sometimes it is wanted to give the user the option to pass own arguments (e.g. emissions to a Flow object).

But we could add a warning for such cases which is suppressed by default but can be activated for debugging.

henhuy:
Personally, I think we should put more stuff out of kwargs into named parameters.
But I also use them to provide additional parameters for flows. Unfortunately, this does not work for Node objects. This would be cool too!

p-snft:
Consider copy&paste errors where a `Source` has `inputs` instead of `outputs`. The resulting model will probably still be computed, but with different results. In fact, even typos (e.g. `imputs`) may result in a system with a valid result, just with disconnected nodes.

uvchik:
> Personally, I think we should put more stuff out of kwargs into named parameters.

[...]

PR #601, opened 2019-06-05
Raise an AttributeError if deprecated attribute is used

uvchik:
The deprecated attributes will be silently ignored. This may lead to an unwanted behavior if users have updated from v0.2.3 to v0.3.0. The error message will help to avoid problems.

The following Storage will work with v0.2.3 as well as with v0.3.0 but some attributes will be ignored, so that the capacity of the storage at time step 0  will be `None` and not `0` as intended because the parameter `initial_capacity` does not exists in v0.3.0 and therefore will be ignored. The parameter `initial_storage_level` (new name of `initial_capacity`) will be set to the default value, which is `None`.

```python
storage = solph.components.GenericStorage(
    label='storage',
    inputs={bel: solph.Flow()},
    outputs={bel: solph.Flow()},
    capacity_loss=0.5,
    initial_capacity=0,
    invest_relation_input_capacity=1/6,
    invest_relation_output_capacity=1/6,
    inflow_conversion_factor=1,
    outflow_conversion_factor=0.8,
    investment=solph.Investment(ep_costs=epc_storage),
)
```

We could also raise a warning instead of an error but I think this is clearer.

We may release an Hotfix v0.3.1 soon to help people to avoid problems with there models.

### DIFF
--- a/oemof/solph/components.py
+++ b/oemof/solph/components.py
@@ -152,6 +152,19 @@ class GenericStorage(network.Transformer):
         if self._invest_group is True:
             self._check_invest_attributes()
 
+        # Check for old parameter names.
+        rename_parameters = [
+            ('nominal_capacity', 'nominal_storage_capacity'),
+            ('initial_capacity', 'initial_storage_level'),
+            ('capacity_loss', 'loss_rate'),
+            ('capacity_min', 'min_storage_level'),
+            ('capacity_max', 'max_storage_level')]
+        for parameter in rename_parameters:
+            if kwargs.get(parameter[0]) is not None:
+                msg = "The attribute '{0}' has been renamed to '{1}'.".format(
+                    parameter[0], parameter[1])
+                raise AttributeError(msg)
+
     def _set_flows(self):
         for flow in self.inputs.values():
             if (self.invest_relation_input_capacity is not None and
--- a/oemof/solph/components.py
+++ b/oemof/solph/components.py
@@ -153,7 +153,7 @@ class GenericStorage(network.Transformer):
             self._check_invest_attributes()
 
         # Check for old parameter names. This is a temporary fix and should
-        # be removed once a general solution is found
+        # be removed once a general solution is found.
         # Todo: https://github.com/oemof/oemof/issues/560
         rename_parameters = [
             ('nominal_capacity', 'nominal_storage_capacity'),
--- a/oemof/solph/components.py
+++ b/oemof/solph/components.py
@@ -155,13 +155,13 @@ class GenericStorage(network.Transformer):
         # Check for old parameter names. This is a temporary fix and should
         # be removed once a general solution is found.
         # TODO: https://github.com/oemof/oemof/issues/560
-        rename_parameters = [
+        renamed_parameters = [
             ('nominal_capacity', 'nominal_storage_capacity'),
             ('initial_capacity', 'initial_storage_level'),
             ('capacity_loss', 'loss_rate'),
             ('capacity_min', 'min_storage_level'),
             ('capacity_max', 'max_storage_level')]
-        for parameter in rename_parameters:
+        for parameter in renamed_parameters:
             if kwargs.get(parameter[0]) is not None:
                 msg = "The attribute '{0}' has been renamed to '{1}'.".format(
                     parameter[0], parameter[1])
--- a/oemof/solph/components.py
+++ b/oemof/solph/components.py
@@ -161,11 +161,23 @@ class GenericStorage(network.Transformer):
             ('capacity_loss', 'loss_rate'),
             ('capacity_min', 'min_storage_level'),
             ('capacity_max', 'max_storage_level')]
-        for parameter in renamed_parameters:
-            if kwargs.get(parameter[0]) is not None:
-                msg = "The attribute '{0}' has been renamed to '{1}'.".format(
-                    parameter[0], parameter[1])
-                raise AttributeError(msg)
+        messages = [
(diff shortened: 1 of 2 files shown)


uvchik:
I could add a test if everybody is fine with my solution.

p-snft:
I like it, as its probably the current best solution.

(On the long run, I would consider it a workaround. I tend to reopen the kwargs discussion for 0.4.)

oemof/solph/components.py
```diff
@@ -152,6 +152,19 @@ def __init__(self, *args,
         if self._invest_group is True:
             self._check_invest_attributes()
 
+        # Check for old parameter names.
```
p-snft: I thought about putting this to a central location, e.g. `oemof.network.Node.__init__`. However, as `max_storage_level` and `min_storage_level` are not passed to the parent constructors, this location is optimal. (Maybe, you could add a comment to explain this.)

uvchik: I do not understand why you want to move this to the Node class. For me it is only related to the GenericStorage and only related to the update process from 0.2.x to 0.3.x. Therefore I think it is a temporal code block and we should keep on discussing a general solution.

BTW: I have no idea why `max_storage_level` and `min_storage_level` are optional parameters and not covered by `kwargs` as all the other parameters.

p-snft: Parameters in kwargs: That's inconsistent anyway. See `ExtractionTurbineCHP`, for another example.

Locating check in node: Similar things could happen with other releases. So it might help. Still, a general solution is of course preferable. #560 really should be resolved before the next major release...

uvchik: As this is a temporal hack I would let it where it is and add a comment that it is temporal.

uvchik:
I would like to release this fix as v0.3.1. within a week. The earlier the better.

 @oemof-developer Any objections?

gnn:
I'd vote for a deprecation warning instead of an error. This is exactly what warnings are for because those can be suppressed. If somebody really wants to use that parameter, because i.e. he defined a subclass and this is just the most reasonable name to use as a parameter in his subclass constructor, he can just use it and suppress the warning because he knows what he's doing. Errors can be caught, but they still jump out of the constructor and even if you catch the error, you can never get any code in the constructor which would be executed after raising the error executed. This might not seem like a big deal here because the check is the last code in the constructor. But even this innocuous example already contains a problem: if somebody uses multiple renamed parameters, he only gets an error for the first one checked. If he then proceeds to catch that error and repair his mistake by setting the attribute manually, he will never be notified of the other misused names. Using warnings would mean the user gets notified of all misused names and can act appropriately.

uvchik:
> I'm not familiar enough with the `GenericStorage` class, but at least to me, this example also raises the question why an `initial_storage_level` of `None` doesn't lead to an error later on. Maybe this is another instance where `initial_storage_level` should have been explicitly set prior to running the model, which means this is another use case for a `validate` function.

`initial_storage_level=None` is the default value and therefore the default use case. It does make sense with `balanced=True` and less sense with `balanced=False`.

p-snft:
> I'm not familiar enough with the `GenericStorage` class, but at least to me, this example also raises the question why an `initial_storage_level` of `None` doesn't lead to an error later on. Maybe this is another instance where `initial_storage_level` should have been explicitly set prior to running the model, which means this is another use case for a `validate` function.

Setting it to `None` is a way to explicitly not set an `initial_storage_level`. IMHO, this is perfectly fine.


> I'd vote for a deprecation warning instead of an error.

This would be another option. Currently, this PR allows arbitrary parameters except those that were handled in the old release, which is not fully aligned with the highly flexible concept of oemof. I think, it is Ok to be strict here, as the meta-programming part is more of an expert users thing.

gnn:
Uwe convinced me that in this case, errors are fine, as it is a temporary solution. While talking we also agreed that in case of such API changes, both versions should be kept around for a deprecation period of at least the majority of one major release's lifetime, with the old version emitting deprecation warnings for that duration. This is something which we'll open a discussion about in another issue though as it's  also something that should eventually be placed in the development rules.

> Setting it to `None` is a way to explicitly not set an `initial_storage_level`. IMHO, this is perfectly fine.
Well yes, but with a slight twist. Setting the attribute to `None` means: "I want the attribute to be present, but with an explicitly undefined value." Not setting it at all means that attribute won't be present, which is different, because in that case, solving the model will fail with an error if it's necessary. So not having a default value here would obviate the need for us to throw an error for this specific attribute.

oemof/solph/components.py
```diff
@@ -152,6 +152,19 @@ def __init__(self, *args,
...
+        rename_parameters = [
+            ('nominal_capacity', 'nominal_storage_capacity'),
+            ('initial_capacity', 'initial_storage_level'),
+            ('capacity_loss', 'loss_rate'),
+            ('capacity_min', 'min_storage_level'),
+            ('capacity_max', 'max_storage_level')]
+        for parameter in rename_parameters:
+            if kwargs.get(parameter[0]) is not None:
```
gnn: This is probably not what you want, because if the user would set any of those parameters to `None`, your check would not get triggered in that case. There's a simpler way of checking for the presence of keyword arguments, which is `parameter[0] in kwargs` or `parameter[0] in kwargs.keys()` which is equivalent.

gnn: Just to be clear: I already changed that in the new patches, but I wanted to point that out anyway, as it might get lost in the diffs.

uvchik:
@gnn Thanks for clarification and you are right your way is more elegant :smile: 

It is on my Todo-List to open a new issue about the `initial_storage_level` attribute.

Issue #585, opened 2019-05-20
Generalize constraints.emission_limit


p-snft:
The concept of additional constraints is rather powerful. Still, it is currently only transported by example. It might be a good idea to generalize `constraints.emission_limit(om, flows=None, limit=None)` to something like `constraints.shared_limit(om, key='emission_factor', shared_between=None, limit)`. This would then transport the idea that for example also limited shared space for PV/ST can also be modelled this way.

PS: The `constraints.investment_limit(model, limit)` would be `constraints.shared_limit(om,'investment_costs', limit).

simnh:
I like the idea!

joroeder:
Hey, in my opinion there are two options:
1) replacing the ``constraints.emission_limit``  by a more generous formulation ``constraints.shared_limit`` like Patrik described
2) keeping the ``constraints.emission_limit`` and adding a more generous formulation. In the end there could be many constraints (like transformer components) which already indicate a specific use

The advantage of 2) is that it might be more intuitive for a constraint many people use. What do you think?

p-snft:
If this is added e.g. in v0.3.1, API compatibility would anyway require to keep the old functionality. However, they should probably just wrap the generic `shared_limit` constraint .

uvchik:
I like the idea :+1: 

I wrap should be the right way if it is not part of the upcoming major release. But I would add a FuturWarning and remove it in the long run. We  could keep the emission constraint as an example in the docstring of the generalised function.

p-snft:
I would consider this done.

PR #602, opened 2019-06-06
Generalise shared limits

p-snft:
This PR implements a generalised version of "emission_limit", as suggested in #585. As this type of limit implements weighted integrals over the flows, I called it `generic_integral_limit`.

Note that `investment_limit` handles `Flow`s as well as `GenericInvestmentStorage`, so it needs a different treatment. It might be possible to incorporate all of these into one creator function for global limit constraints, but that would then need to work on `Node`s instead of specialised implementations.

### DIFF
--- a/oemof/solph/constraints.py
+++ b/oemof/solph/constraints.py
@@ -77,7 +77,7 @@ def generic_integral_limit(om, keyword, flows=None, limit=None):
     flows : dict
         Dictionary holding the flows that should be considered in constraint.
         Keys are (source, target) objects of the Flow. If no dictionary is
-        given all flows containing the 'emission_factor' attribute will be
+        given all flows containing the keyword attribute will be
         used.
     keyword : attribute to consider
     limit : numeric
@@ -99,8 +99,8 @@ def generic_integral_limit(om, keyword, flows=None, limit=None):
             if not hasattr(flows[i, o], keyword):
                 raise AttributeError(
                     ('Flow with source: {0} and target: {1} '
-                     'has no attribute emission_factor.').format(i.label,
-                                                                 o.label))
+                     'has no attribute {2}.').format(
+                        i.label, o.label, keyword))
 
     limit_name = "integral_limit_"+keyword
 
--- a/oemof/solph/constraints.py
+++ b/oemof/solph/constraints.py
@@ -88,6 +88,22 @@ def generic_integral_limit(om, keyword, flows=None, limit=None):
     ----
     Flow objects required an attribute named like keyword!
 
+    **Constraint:**
+
+    .. math::
+        \sum_n \sum_t w_n(t) \cdot P_n(t) \cdot \tau(t) \le L \\
+
+    The symbols used are defined as follows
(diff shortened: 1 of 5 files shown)


oemof/solph/constraints.py
```diff
@@ -60,38 +78,41 @@ def emission_limit(om, flows=None, limit=None):
...
+                     'has no attribute {2}.').format(
+                        i.label, o.label, keyword))
+
+    limit_name = "integral_limit_"+keyword
 
-    om.total_emissions =  po.Expression(
+    setattr(om, limit_name, po.Expression(
         expr=sum(om.flow[inflow, outflow, t] * om.timeincrement[t] *
-                 flows[inflow, outflow].emission_factor
+                 sequence(getattr(flows[inflow, outflow], keyword))[t]
```
p-snft: When we allow time dependent weights, this should also be covered in the unit tests.

joroeder: Yeah sure. Do you think it is sufficient to adapt source1 in the constraint test?
```suggestion
                 solph.Source(label='source1', outputs={bel: solph.Flow(
                     nominal_value=100, emission_factor=[0.5, 0.6, 0.7])})
```
If you want, I can adapt the test and lp file...?

p-snft: Done.

p-snft: Done.

joroeder:
Thanks!

uvchik:
@p-snft , @joroeder Thanks for your changes. I really like it. Just two remarks.

1. We should not use `om` anymore. It comes from the old `OperationalModel` class which does not exist anymore. You could use `model`, `solph_model` or something like this instead. Furthermore, this parameter cannot be found in the docstring. e.g. `model : oemof.solph.Model`.

2. Do you want to keep the `emission_limit` function? Otherwise we should add a deprecated warning. I think we could keep it like it is, because `emission_limit` might be a very common use of this function.

p-snft:
1. Interesting background information. I translated it to OemofModel...
2. I would keep it, as it also helps to learn by example how the limits.

PS: Probably, I was a bit fast with this one.

uvchik:
2. Fine.

1a. Good translation, but it is more a solph model than an oemof model, so I would like to change it. What about just `model`. Even though it is not urgent, but I would change it in the long run.

1b. We should add the parameter to the docstring. We could open a new PR for that or just a fix-commit.

p-snft:
1b. Actually, I do not see what is missing. the docstring says:
```
    Parameters
    ----------
    om : oemof.solph.Model
        Model to which constraints are added.
    flows : dict
        Dictionary holding the flows that should be considered in constraint.
        Keys are (source, target) objects of the Flow. If no dictionary is
        given all flows containing the keyword attribute will be
        used.
    keyword : attribute to consider
    limit : numeric
        Absolute limit of keyword attribute for the energy system.
```

uvchik:
Ups, sorry. Everything is fine.

PR #605, opened 2019-06-11
Revise results handling in the BaseModel

uvchik:
In my opinion the results handling in the BaseModel is confusing, so I would like to change it.

1. The solver-results (see below) are written to the `results` attribute of the EnergySystem object.

I could not find it in the documentation that the solver-results can be found there and I wonder why the main results (results of the variables) cannot be found there. We should discuss if we want the results written back in the EnergySystem or not. If we want to keep it as it is, we should at least rename the attribute from `results` to `solver_results`.

Normally I go this way:

```python
[...]
model = solph.Model(energysystem)
model.solve()
energysystem.results['main'] = model.results()
energysystem.results['meta'] = outputlib.processing.meta_results(model)
energysystem.dump(dpath=something, file=something)  # to store the results
```

1.1 If we do have `results = model.results()` we should also have `solver_results = model.solver_results()`

1.2 Executing the code above I overwrite the solver results in `energysystem.results`. This happend because I did not know about them. Therefore I vote for either having two attributes/keys or stop writing the solver-results to the EnergySystem. That means that we do have an empty dictionary to add the results with the keys you want as I do it normally or we already add them directly. I do not care if we use keys (as above) or attributes like `energysystem.results` and `energysystem.solver_results`.

2. There are different combinations of status messages and termination conditions given back by the solver. Typically we do have `status=ok` and `termination condition=optimal` or  `status=ok` and `termination condition=infeasible`.

I reduced the number of possible outputs to these two combinations. There might be other solutions but I do need help to reproduce them in a test.

3. It was not possible to create a useful error message if the problem was infeasible.

Now you can try to access the solver-results if the problem is infeasible. You get an error message with the constraint the causes the problem. This may help to find the problem.

4. The model is not well documented in the documentation.

Please help.

Thanks for reading :smile: 

solver-results:
```python
Problem: 
- Name: unknown
  Lower bound: 25.0
  Upper bound: 25.0
  Number of objectives: 1
  Number of constraints: 2
  Number of variables: 2
  Number of nonzeros: 0
  Sense: minimize
Solver: 
- Status: ok
  User time: -1.0
  System time: 0.0
  Wallclock time: 0.0
  Termination condition: optimal
  Termination message: Model was solved to optimality (subject to tolerances), and an optimal solution is available.
  Statistics: 
    Branch and bound: 
      Number of bounded subproblems: None
      Number of created subproblems: None
    Black box: 
      Number of iterations: 0
  Error rc: 0
  Time: 0.01261448860168457
Solution: 
- number of solutions: 0
  number of solutions displayed: 0
```

### DIFF
--- a/VERSION
+++ b/VERSION
@@ -1,2 +1,2 @@
-__version__ = "0.3.1"
+__version__ = "0.3.2dev"
 
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -8,6 +8,7 @@ These are new features and improvements of note in each release
     :local:
     :backlinks: top
 
+.. include::  whatsnew/v0-3-2.rst
 .. include::  whatsnew/v0-3-1.rst
 .. include::  whatsnew/v0-3-0.rst
 .. include::  whatsnew/v0-2-3.rst
--- a/doc/whatsnew/v0-3-1.rst
+++ b/doc/whatsnew/v0-3-1.rst
@@ -1,48 +1,19 @@
-v0.3.1 (September ??, 2019)
+v0.3.1 (June 11, 2019)
 ++++++++++++++++++++++++++
 
 
-API changes
+Other changes
 ###########
 
-* something
-
-New features
-############
-
-* something
-
-New components
-##############
-
-* something
-
-Documentation
-#############
-
-* something
-
-Known issues
-############
-
-* something
-
-Bug fixes
-#########
-
-* something
-
-Testing
-#######
-
-* something
-
-Other changes
-#############
-
-* something
+* The API of the GenericStorage changed. Due to the open structure of solph
+  the old parameters are still accepted. Therefore users may not notice that
+  the default value is used instead of the given value after an update from
+  v0.2.x to v0.3.x. With this version an error is raised. We work on a
+  structure to avoid such problems in the future.
 
 Contributors
 ############
 
-* something
+* Patrik Schönfeldt
+* Stephan Günther
+* Uwe Krien
--- a/doc/whatsnew/v0-3-1.rst
+++ b/doc/whatsnew/v0-3-1.rst
@@ -3,7 +3,7 @@ v0.3.1 (June 11, 2019)
 
 
 Other changes
-###########
+#############
 
 * The API of the GenericStorage changed. Due to the open structure of solph
   the old parameters are still accepted. Therefore users may not notice that
--- a/doc/whatsnew/v0-3-2.rst
+++ b/doc/whatsnew/v0-3-2.rst
@@ -1,0 +1,48 @@
+v0.3.2 (September ??, 2019)
++++++++++++++++++++++++++++
+
+
+API changes
+###########
+
+* something
+
+New features
+############
+
+* something
+
+New components
+##############
+
+* something
+
+Documentation
+#############
+
+* something
+
+Known issues
+############
+
+* something
+
+Bug fixes
+#########
+
+* something
+
+Testing
+#######
+
+* something
+
+Other changes
+#############
+
+* something
+
+Contributors
+############
+
+* something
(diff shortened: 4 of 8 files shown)


simnh:
I really like that this PR exists. Separating solver results from optimization values sound really sound. In addition we might think about the dual variables / reduced costs. Right now we only take the dual variables of the balance constraint, as this is the most important one for energy system analysis. However, others might also be important. Therefore two options exists: 

1) either provide all duals as results (we could simply transfer the pyomo values for the duals to a dataframe and drop it somewhere)
2) Be more specific of the output for the balance duals (which would be something like 'bus_duals' and not only duals as duals does not contain all opimization duals but only the ones from the bus....

uvchik:
@simnh What if we keep the name of the attribute but change the name of the method e.g. to `receive_bus_duals` and `receive_all_duals`.

uvchik:
What is the expression `self.solutions.load_from(solver_results)` for. I do not see any difference if I remove it. I think we do not need it anymore. I found on a pyomo site:

> The solve method on solver plugins now automatically loads results into Pyomo models. This means that there is no need to call load with the results object after the solve completes inside of a Pyomo script. To recover the old behavior, simply add load_solutions=False to the call to solve and use solutions.load_from in place of load to store a solution in a Pyomo model. Example:

> Old Coopr Code:
 results = opt.solve(instance)
 instance.load(results)

> Pyomo 4.x Code:
 results = opt.solve(instance, load_solutions=False)
 instance.solutions.load_from(results) 

The actual version of pyomo is 5.6.

simnh:
Regarding the 'load_from..) method, I also think this is obsolete for some time of the last pyomo versions already.

simnh:
> @simnh What if we keep the name of the attribute but change the name of the method e.g. to `receive_bus_duals()` and `receive_all_duals()`.

Yes this might be a quick and easy solution (Maybei `get_bus_balance_duals` would be the most specific one. Maybe 'receive_duals('all') or 'receive_duals('bus_balance'), etc. could be the best way because with every constraint in the model their exist a dual variable. If we just let the user enter the constraint name (and if everything should be received) this would be most generic and maybe also easy to implement?

uvchik:
I try to structure this discussion so I opened #610 to discuss the duals-API.

I will remove the `load_from...` expression.

@oemof-developer: Still open: 

* Shall we add the results to the EnergySystem object, the Model or just return them (as it is now).

* Shall we distinguish between different solver termination conditions?

uvchik:
I talked to @simnh and we decided to move the changes of the results processing to an extra PR.

uvchik:
There are still open questions, but as everybody seems to be busy at the moment I would merge this PR by the end of the week if no one objects. Afterwards I will open an issue with the open topics.

@oemof-developer: Still open: 

* Shall we add the results to the EnergySystem object, the Model or just return them (as it is now).

* Shall we distinguish between different solver termination conditions?

p-snft:
> Shall we add the results to the EnergySystem object, the Model or just return them (as it is now).

I do not see any advantage in adding them to one of the objects. When the user likes to have them at that location, this can as well be done after returning.


> Shall we distinguish between different solver termination conditions?

Would this be this possible? To my knowledge, different solvers do not even have the same set of termination conditions.

uvchik:
> I do not see any advantage in adding them to one of the objects. When the user likes to have them at that location, this can as well be done after returning.

Most people may want to analyse the results after the solver finished. Therefore processing the results could be the default behaviour. At the moment it seems to be something special which can be done afterwards. But I do not have a strong opinion on that.

> Would this be this possible? To my knowledge, different solvers do not even have the same set of termination conditions.

I don't know. It is just a question. At the moment there are four different types but I think they do not work properly. So I reduced them to two (optimal or not optimal). Anybody might miss a option and so I asked. But we can just merge and add an option later on if anything is missing.

I will think of opening an issue with the first question and ignore the second until anybody objects. Anyhow I would like to merge the improvements I already did.

PR #613, opened 2019-06-18
Test docs in CI chain

p-snft:


uvchik:
It looks good. Why did you close it?

p-snft:
As there are several errors also in existing docs, I decided to replace this by #614.

PR #614, opened 2019-06-18
Fixes/latex in docstrings

p-snft:
* [X] Add test of docstrings to CI toolchain
* [ ] Fix docstrings (should happen before merge)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,24 +1,32 @@
 language: python
 
+dist: xenial
+
 matrix:
   include:
     - python: 3.5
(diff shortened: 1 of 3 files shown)


p-snft:
The test is pretty pedantic. Could someone help?

uvchik:
I changed the base because `v0.3` will no longer exist.

This PR is about a year old. @p-snft and @FranziPl, do you still care?

p-snft:
Docstrings are now tested anyway, so this is not needed any more.

Issue #610, opened 2019-06-14
API to get results of dual variables from the model


uvchik:
This is a spin-off of PR #605 to discuss the API of dual variables.

@simnh 

Shall we call it `fetch_`, `get_` or `receive_` `duals`?

I think `receive_duals` is fine because it is executed before the solver run.

Furthermore, I like the idea to use attributes but I do not know how it works. This would be comfortable API.

```python
fetch_duals(category=None, constraint=None)
```

`category` could be `bus`, `all` etc. and `constraint` could be more explicit to one constraint. Is it possible to get the dual variable for just one constraint? Is it possible to get duals for all buses or all components or some components?

If it is not possible we could just use `bus` or `all`.

simnh:
Ok I added a few lines. Now the duals will be available by its constraint: 

```
results = m.results()
results[m.Transformer.relation, None]
```

I do not know if not an additional result dict would be good. We can think about the keys as well. However I think we will always fetch all duals or None and then provide an restriction in the result dict I would suggest.

PR #616, opened 2019-06-18
Add all duals to result dict

uvchik:
* Add the possibility to add all dual variables to the results dictionary. By now it is only possible to get the dual variables of the buses.

* The API stays the same, but we could add a parameter to filter the dual variables to keep the results smaller e.g. only the dual variables of the electricity bus

* #605, #610 

* [ ] Add tests
* [ ] Adapt documentation
* [ ] Adapt whatsnew


p-snft:
Today, we agreed to completely revise the result handling. Thus, I think this is outdated.

PR #620, opened 2019-06-24
Feature/generic invest limit

joroeder:
re-opening of PR due to rebase to v0.3 instead of dev (branches with no API changes)

+++

Analog to the general formulation of a constraint function for flows (generalization of co2 limit #602 ), there was the idea of providing an option of putting attributes to investments, which are limited. (e.g. different investments need some further resources which are limited). Maybe #585 was ment for discussing that as well, at least there was a discussion at the dev meeting about that functionality.
This features requires a modification of the Investment class in option.py - is there something we need to consider at this point?

What do you think about that generally? And could you give me a feedback about the implementation? It is very close to generic_integral_limit.

- [x] tests
- [x] documentation
- [x] example

[readthedocs - generic_investment_limit](https://oemof.readthedocs.io/en/feature-generic_invest_limit_2/api/oemof.solph.html#oemof.solph.constraints.generic_investment_limit)


joroeder:
I "manually" merged the dev into this feature (by replacing all files of my local feature by copying the dev repo, and then, adding the lines of the feature again), but github says, 156 changed files ?! hm ..

uvchik:
Seems to work as expected. Test are still missing though.

Issue #621, opened 2019-06-25
New component: Stratified Thermal Storage


jnnr:
We want to represent stratified thermal storages more in detail. Therefore we want to implement the following component. Any feedback or experiences are warm welcome!

A simplified 2-zone-model of a stratified thermal energy storage can be formulated like this:

- We assume a cylindrical storage with of diameter d and height h, with two temperature regions that are perfectly separated. 
- The temperatures are assumed to be constant and correspond to the feedin/return temperature of the heating system. 
- Heat conductivity of the storage has to be passed as well as a timeseries for outside temperatures for the calculation of heat losses. 
- There is no distinction between outside temperature and ground temperature.
- Material properties are constant.

In the case of investment, the diameter d is given and the height can be adapted to adapt the nominal capacity of the storage. With this assumption, all relations stay linear.

jnnr:
![heat_storage](https://user-images.githubusercontent.com/32454596/60095750-95073980-974f-11e9-855b-4ad2ce4152d7.png)

jnnr:
![2DG_OEM_ESP_01_1-0](https://user-images.githubusercontent.com/32454596/60095782-afd9ae00-974f-11e9-889c-28e7861ce6a3.jpg)

jnnr:
The loss term looks a bit different than in the GenericStorage. In case of investment, there are a two more equations. This suggests that a new component would be necessary.

simnh:
Nice idea, definitely a new component. Maybe someday we eventually also get a battery model, which is more complex than the generic storage! 

Contentwise: I am not a thermal engineer ;-), but sounds good anyway. Go for it!

joroeder:
Hey, nice, very interesting!

in the end, this component has an additional constant loss, which does not depend on the state of charge. Did I get it right? The term -U*A*(Q_t-1/Q_N)*delta(T_HC) could be modeled with the existing loss_rate, but a constant loss -U*A**delta(T_C0) independent of the state of charge does not exists at the moment (like for the transformer and flows ( #618 )).

For the investment, the diameter needs to fixed, and all the other stuff results from it with a given Q_N - which is a variable... am I right? Because an optimal diameter - high relation would results in a NLP..

How about adapting the existing general storage by a constant loss, and doing a face class out it with all the precalculations you described? I am happy to help :)

p-snft:
I support @joroeder's idea to allow a constant loss in the GenericStorage and to add a facade.

jnnr:
[...]

Issue #538, opened 2018-11-13
Word "Capacity" in GenericStorage can be misleading


p-snft:
Current renaming due to this ticket:

    capacity_loss -> loss_rate
    initial_capacity -> initial_storage_level (between 0 and 1)
    capacity_max -> max_storage_level (between 0 and 1)
    capacity_min -> min_storage_level (between 0 and 1)
    nominal_capacity -> nominal_storage_capacity (an Energy)

**Original comment:**
In the context of the _GenericStorage_, _Capacity_ can refer to different things. From my perspective, the word is used inconsistently:

* nominal_capacity (maximum energy that can be stored, the _storage capacity_)
* capacity_loss (loss of currently stored loss per time step, might be confused with parameter for general degradation of the storage, cf. [Wikipedia](https://en.wikipedia.org/wiki/Capacity_loss))
* initial_capacity (initially stored energy, might be confused with parameter for general degradation of the storage)
* "Invest into capacity" (in documentation, cf. [doc/oemof_solph.rst#genericstorage-component](https://github.com/oemof/oemof/blob/60b30d9676c6b05ffa487570e4d3e2222d3ec37b/doc/oemof_solph.rst#genericstorage-component)), again refers to the storage capacity

The problem is known, and somewhat addressed in the documentation: "Investment in this context refers to the value of the variable for the 'nominal_value' (installed capacity) in the investment mode."

Capacity is often used synonym to storage capacity. This holds for batteries [J], and even water basins (in litres, when not used as pumped hydroelectric energy storage). For thermal energy storage, I found capacity used the same way. (Note also _heat capacity_ is a fixed term defined in units of J/K. For thermal reservoirs the temperature is typically assumed to be constant.) Concluding, I would opt to rename capacity_loss and initial_capacity.

Suggestions:
* loss rate (instead of capacity loss)
* initial storage level (instead of initial capacity, used e.g. in [doi:10.1109/TIA.2007.895768](https://doi.org/10.1109/TIA.2007.895768))

simnh:
I like your suggestions!!

One additional thought: We actually decided to use capacity for power and storage_capacity for energy in our models. Typically capacity is often used in energy system analysis context for power related values like the capacity of the powerplant...we found this very helpful. This makes also investment more precise in terms of naming, as capacity investment is then also for power... 

Therefore maybe `nominal_capacity` might also be changed to `nominal_storage_capacity`

[...]

PR #625, opened 2019-07-24
Features/storage constant loss

jnnr:
To implement the functionality of #621, a constant loss term is added to GenericStorage. New attributes `fixed_losses_relative` and `fixed_losses_absolute` are introduced. They are set to zero by default. A scalar or a sequence can be passed.

TODO:
* [x] Docstring formulas,
* [x] an example: https://github.com/oemof/oemof-thermal/blob/features/stratified_thermal_storage/examples/stratified_thermal_storage/example_stratified_thermal_storage.py
* [x] Updated tests.


oemof/solph/components.py
```diff
@@ -273,6 +277,8 @@ class GenericStorageBlock(SimpleBlock):
     :math:`c_{max}(t)`          maximum allowed storage :py:obj:`max_storage_level[t]`
     :math:`\delta(t)`           fraction of lost energy :py:obj:`loss_rate[t]`
                                 (e.g. leakage) per time
```
joroeder: maybe add here `(dependent on E(t)) `

oemof/solph/components.py
```diff
@@ -273,6 +277,8 @@ class GenericStorageBlock(SimpleBlock):
...
+    :math:`\gamma(t)`           constant loss of energy :py:obj:`loss_constant[t]`
+                                (e.g. leakage) per time
```
joroeder: maybe add here `(independent of E(t))`

tests/test_processing.py
```diff
@@ -169,6 +169,7 @@ def test_nodes_with_none_exclusion(self):
                 'investment_minimum': 0,
                 'label': 'storage',
                 'inflow_conversion_factor': 1,
+                'loss_constant': 0,
```
joroeder: I have no idea what test_processing is testing - so probably thats fine, but I don't now.

p-snft:
>  Last dev meeting we were discussing the branch structure and @p-snft made a nice suggestion, I think everybody agreed on... Or is adding a new optional parameter an API change? The old models will still work.

You are right. I rebased the PR to the v0.3 branch.

oemof/solph/components.py
```diff
@@ -65,6 +65,9 @@ class GenericStorage(network.Transformer):
     loss_rate : numeric (sequence or scalar)
         The relative loss of the storage capacity from between two consecutive
         timesteps.
+    loss_constant : numeric (sequence or scalar)
+        A constant loss of the storage level between two consecutive
+        timesteps.
```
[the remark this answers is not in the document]
p-snft: As it can be time-dependent, I would call it "fixed loss" instead of "constant loss".

oemof/solph/components.py
```diff
@@ -377,6 +384,7 @@ def _storage_balance_rule(block, n, t):
             expr += block.capacity[n, t]
             expr += - block.capacity[n, t-1] * (
                 1 - n.loss_rate[t])
+            expr += n.loss_constant[t]
```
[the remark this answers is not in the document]
p-snft: For the loss_rate, relative losses are inherently needed. I think, it is easier here, to have absolute values.

For invest storages, it would even make sense to have to kinds of "fixed" losses: One type growing with the nominal capacity, one that is really fixed.

oemof/solph/components.py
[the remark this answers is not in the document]
joroeder: I like to call it "fixed losses"

oemof/solph/components.py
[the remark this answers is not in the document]
joroeder: Hm... especially in case of invest storages, I think a capacity dependent fix-loss makes sense. In case of a water tank, whose volume is growing and with the volume the surface, then the fix-losses are growing as well. So, in case if no investment in the storage is done, there will be no fix-losses as well.

jnnr:
Thank you for your helpful comments! 

I agree that the losses are better referred to as fixed rather than constant losses. I share @joroeder's view, that the new losses should be in reference to the nominal_storage_capacity. To avoid confusion, I would parallel this in storages and investstorages. So we would have:

* Energy content related loss (`loss_rate`)
* Capacity related losses

As @p-snft suggested, we could have a third type, truly fixed losses. There might be situation where this could be useful. For now, the suggested thermal storages requires capacity related losses (depending on the surface of the storage).

In the meantime, I wonder why the energy content in the storage at each time is called `capacity.` In my view, capacity is what is now called `nominal_storage_capacity`, which tells us the maximum amount of energy that can be stored. What is now called capacity would be better called `energy_content`, no?

p-snft:
> In the meantime, I wonder why the energy content in the storage at each time is called capacity. In my view, capacity is what is now called nominal_storage_capacity, which tells us the maximum amount of energy that can be stored. What is now called capacity would be better called energy_content, no?

We have had some discussion about that naming in #538.

jnnr:
Thanks for pointing to that. I wrote a short comment there.

tests/test_processing.py
jnnr: It tests if the return of the outputlib.processing.param_as_dict() function returns the expected values.

jnnr:
Thanks again for your helpful comments, @joroeder and @p-snft. I read your feedback and improved the docstrings. The new attribute is now called `fixed_losses`, as you suggested. It is defined as share of the nominal storage capacity. If you find the time, please have another look and tell me if you got other remarks.

jnnr:
Hi @FranziPl, I invited you to review as this PR is connected to https://github.com/oemof/oemof-thermal/pull/19

joroeder:
sorry, this week is already lot of stuff, hopefully next week I find some time to do a review

oemof/solph/components.py
```diff
@@ -63,8 +63,11 @@ class GenericStorage(network.Transformer):
         Couple storage level of first and last time step.
         (Total inflow and total outflow are balanced.)
     loss_rate : numeric (sequence or scalar)
-        The relative loss of the storage capacity from between two consecutive
+        The relative loss of the storage capacity between two consecutive
         timesteps.
+    fixed_losses : numeric (sequence or scalar)
+        Losses as share of nominal storage capacity between two consecutive
+        timesteps independent of storage capacity.
```
FranziPl: Isn't it missleading to call it independent of storage capacity (naming discussion #538). I would say independent of storage_content.

tests/constraint_tests.py
```diff
@@ -225,6 +225,7 @@ def test_storage(self):
             outputs={bel: solph.Flow(nominal_value=16667, variable_costs=24)},
             nominal_storage_capacity=10e4,
             loss_rate=0.13,
+            fixed_losses=0.01,
```
uvchik: Please add a new test instead of changing an existing one.

oemof/solph/components.py
jnnr: >  SoC is well know I think. "storage_content" is misleading to me as well, some could interpret this as water, oil, ...

For discussing this, please go to #538.

jnnr:
> But I have a question: What happens, if the storage is empty, but you still have the fixed losses? How is avoided, that there is a constant input, just to serve this losses?

As you say it. The storage will never be empty as the solver will find a way to avoid that.

tests/constraint_tests.py
jnnr: I reverted the commit and will add a new test.

oemof/solph/components.py
jnnr: I changed to 'state of charge' to avoid confusion.

jnnr:
Thanks for all your comments, very helpful. We noticed that we actually need a third attribute, fixed absolute losses that are not related to nominal storage capacity. I will work on that soon.

jnnr:
Thanks again for your helpful comments. I tried to follow your advice and changed the tests as you asked for, @uvchik. In addition, I implemented a further distinction that we discussed before:

> For invest storages, it would even make sense to have to kinds of "fixed" losses: One type growing with the nominal capacity, one that is really fixed.

I did exactly that. Now there are `fixed_losses_relative` and `fixed_losses_absolute`. Some more explanation can be found in the docstring.

Happy to read your feedback!

uvchik:
> But I have a question: What happens, if the storage is empty, but you still have the fixed losses? How is avoided, that there is a constant input, just to serve this losses?

The only way to avoid this is a binary variable, which would lead to an MILP but I would open an extra PR if you need it. This will reduce the complexity of each PR.

> I did exactly that. Now there are `fixed_losses_relative` and `fixed_losses_absolute`. Some more explanation can be found in the docstring.

I am interested in a use case for constant absolute fixed losses for an investment storage.
Do you have an example that shows that it works as expected?

I am fine with the tests now :smile:

jnnr:
> The only way to avoid this is a binary variable, which would lead to an MILP but I would open an extra PR if you need it. This will reduce the complexity of each PR.

I agree. This can be added if needed, but we prefer not to introduce an integer variable at that point.

> I am interested in a use case for constant absolute fixed losses for an investment storage.
> 
> Do you have an example that shows that your changes work as expected?

There is an example in here: https://github.com/oemof/oemof-thermal/pull/19
If needed, I can add one generic example to the solph-examples in oemof-examples.

joroeder:
> > For invest storages, it would even make sense to have to kinds of "fixed" losses: One type growing with the nominal capacity, one that is really fixed.
> 
> I did exactly that. Now there are `fixed_losses_relative` and `fixed_losses_absolute`. Some more explanation can be found in the docstring.

For now, I don't think thats a good idea. In case of an investment storage and the solver is not investing in the storage, there will still be a constant loss. So, there will be a energy loss of a storage without having a storage? E.g. if an invest_relation_input_capacity is given, the solver is forced to build a storage, which is exactly as big as the inflow can cover the losses.
As Uwe said, if we want to consider fix losses, which do not dependent on the nominal_capacity, we need an MILP formulation. Let's skip that for now and think about that in #636 - there we can consider that in a good way.
In case of fix nominal_capacity ("non-invest" case), you can directly calculate the constant losses, I don't understand why additional fixed_losses are necessary in that case - this could be part of the precalculation.

uvchik:
> There is an example in here: [oemof/oemof-thermal#19](https://github.com/oemof/oemof-thermal/pull/19)
> If needed, I can add one generic example to the solph-examples in oemof-examples.

Wherever it is convenient for you. 

A link to an example will make the review easier. Otherwise it is time-consuming to check the outcome of the changes. The reviewers have to create their own examples.

jnnr:
> For now, I don't think thats a good idea. In case of an investment storage and the solver is not investing in the storage, there will still be a constant loss. So, there will be a energy loss of a storage without having a storage? E.g. if an invest_relation_input_capacity is given, the solver is forced to build a storage, which is exactly as big as the inflow can cover the losses.

Yes, that is a problematic edge case. Good that you point to it! 

> As Uwe said, if we want to consider fix losses, which do not dependent on the nominal_capacity, we need an MILP formulation. Let's skip that for now and think about that in #636 - there we can consider that in a good way.

#636 implements another functionality not directly related to this problem here, but you are right, it is similar in the way that it uses a MILP approach.

> In case of fix nominal_capacity ("non-invest" case), you can directly calculate the constant losses, I don't understand why additional fixed_losses are necessary in that case - this could be part of the precalculation.

Yes, thats right. We could restrict the attribute fixed_losses_absolute to the investment case. But then we had different APIs in the non-invest and invest case, which potentially would cause more confusion.

> One use case could be, that a minimum investment needs to be set as well (in case of investment storage). For example, you want to know how big your storage should be and you already know, that you want to build a storage with a minimum capacity. Then, an additional absolute fixed_losses can make sense. What do you think about that?

This would mean to set another constraint to the already existing one. The existing one is the invest_relation_input_capacity together with the absolute losses. But the new constraint still excludes the option of not building a storage.

uvchik:
> Yes, thats right. We could restrict the attribute fixed_losses_absolute to the investment case. But then we had different APIs in the non-invest and invest case, which potentially would cause more confusion.

I would not want different APIs for the invest or non-invest case even though absolute losses and relative losses are the same in the non-invest case.





> There is an example in here: [oemof/oemof-thermal#19](https://github.com/oemof/oemof-thermal/pull/19)
> If needed, I can add one generic example to the solph-examples in oemof-examples.

I get it. Your model of a storage is a tube with a fixed diameter and a variable height. And the losses at the top and the bottom will be the same regardless of the height. In that case you have to define a minimum and maximum invest. Because such an approach needs boarders. In that case the considers of @joroeder are not problematic any more.

jnnr:
> I get it. Your model of a storage is a tube with a fixed diameter and a variable height. And the losses at the top and the bottom will be the same regardless of the height. 

Exactly. The losses at the top and bottom are the independent of the invested storage capacity. The `fixed_losses_absolute` could become useful in other storage or battery models as well.

> In that case you have to define a minimum and maximum invest. Because such an approach needs boarders. In that case the considers of @joroeder are not problematic any more.

 Shall I add some check that warns the user when a non-zero `fixed_losses_absolute` is set but no invest_minimum is given?

joroeder:
> Shall I add some check that warns the user when a non-zero `fixed_losses_absolute` is set but no invest_minimum is given?

I think thats a nice solution. So people who are using the investment without minimum storage capacity, they are informed that they migth build a constant sink. And, the pre-calculations for investment- and non-investment storage can stay similar. And finally, we already have the ```fixed_absolute_losses``` in the storage API, which we can use as well for develop the milp version where the constant part of the losses is switched off in case of non-investment.

jnnr:
Great! I included an error message that checks investment.existing and investment.minimum. 

Thanks again for all your feedback. It there are no more open points, are we ready to merge?

PR #627, opened 2019-08-26
Fix error message if object is a component not a Flow

uvchik:
According to [this issue in the forum](https://forum.openmod-initiative.org/t/genericstorage-results-non-invest-error-label-bus-tuple-nonetype/1676) the solph error message fails if the object is a component. This is rare because the error message is raised if a sequence has a different size than the time index. For example storages do have time series.

* [ ] Add a test.


uvchik:
I drew a wrong conclusion. I will close this PR.

PR #632, opened 2019-09-25
Features/#622 dsm custom component

### LINKED ISSUE #622: New custom component for Demand Side Management
#### Dear oemof dev(_elopment_)-com(_unity_),

Within the Windnode-Project, we are about to develop a new custom component addressing the implementation of "Demand-Side-Management (DSM)" in _oemof_. 

Due to the rising shares of renewables in the grid and the german market design, production flexibility has high costs. Therefore solving the issue on the counterside by demand side flexibility comes to the fore. Shifting load capacities for a certain amount of time will be used for peak load shaving or to smoothen load curves.

W. Schill & A. Zerrahn (DIW) already proposed an alternative formulation in 

> Alexander Zerrahn, Wolf-Peter Schill: 
On the Representation of Demand-Side Management in Power System Models 
[Energy 84 (2015), S. 840-845] [see here: DIW ](https://www.diw.de/sixcms/detail.php?id=diw_01.c.503717.de)

which they implemented in the GAMS-tool "_DIETER_" 

As the formulation is promising, we will follow the path and seek for further development. Any kind of contribution is warmly welcomed.

Regards
Julian
Student assistant at RLI 
Windnode Project @gplssm

nailend:
Fix #622 

This is a new custom component  addressing demand-side-management. 

It is a special sink component which modifies the input demand series within boundaries. These boundaries are set by certain constraints.
For more information have a look in the corresponding issue #622 .

There is a choice of 2 different methods:

-  delay: implementation after Zerrahn&Schill (DIW)

> Alexander Zerrahn, Wolf-Peter Schill:
On the Representation of Demand-Side Management in Power System Models
[Energy 84 (2015), S. 840-845

- interval: simple implementation in which DSM has to be compensated during a certain interval. 

 Parameters
    ----------
   - demand: int or array
        demand defines the original demand
  -  c_up: int or array
        DSM capacity that may be increased at maximum
 -   c_do: int or array
        DSM capacity that may be reduced at maximum
  -  **method: 'interval' , 'delay'
  -  **shift_interval: int
     only used in "interval" :
        interval in between which total DSM  must be fully compensated (default = 24h )
  -  **delay_time: int
        only used in "delay" :
        the time after which the load shift must be fully compensated(default = 3h )

There is a third method combining both 'interval' and ' delay' method. 
But as there are a lot of fringe effects provoked which haven't been examined it is not implemented.
You can find the code here:
 https://gist.github.com/nailend/fd365b541114976a4dfa17f290bcb216


nailend:
@gplssm

gplssm:
@nailend could you please tell about the new component in the [What's new](https://github.com/oemof/oemof/blob/dev/doc/whatsnew/v0-3-2.rst) file? 
Afterwards, please remove the "draft" label and ask an experienced oemof developer for a review (i.e. @p-snft ).

nailend:
@p-snft as you have already  been very engaged in the issue I would like to ask you to review the new custom component.  Would you mind?
This will be very much appreciated.

oemof/solph/custom.py
```diff
@@ -11,15 +11,20 @@
...
-                           Expression, BuildAction)
+                           Expression, BuildAction, Reals)
 import logging
 
-from oemof.solph.network import Bus, Transformer, Flow
+from oemof.solph.network import Bus, Transformer, Flow, Sink
 from oemof.solph.plumbing import sequence
 
 
+
```
p-snft: According to PEP8, there should be exactly two blank lines before class definitions.

oemof/solph/custom.py
```diff
@@ -839,3 +844,491 @@ def tes_ub_constr_rule(block, n, t):
...
+
+
+    **Table: Symbols and attribute names of variables and parameters**
+
+        .. csv-table:: Variables (V) and Parameters (P)
+            :header: "symbol", "attribute", "type", "explanation"
+            :widths: 1, 1, 1, 1
+
+            ":math:`DSM_{t}^{updown}` ",":py:obj:`DSMupdown[g, tt]` ", "V", "DSM up/down shift"
```
p-snft: It is not immediately clear, what the variables/parameters mean. https://oemof.readthedocs.io/en/latest/api/oemof.solph.html#oemof.solph.components.GenericStorageBlock can be used as a best-practice example:

* For the storage, we use E for energies are call them capacity.
* Normalized units are "levels" and have the symbol "c".
* Flows are \dot{E}.

oemof/solph/custom.py
```diff
@@ -839,3 +844,491 @@ def tes_ub_constr_rule(block, n, t):
...
+   **Table: Symbols and attribute names of variables and parameters**
+
+
+        .. csv-table:: Variables (V) and Parameters (P)
+            :header: "symbol", "attribute", "type", "explanation"
+            :widths: 1, 1, 1, 1
+
+
+
+            ":math:`DSM_{t}^{up}` ", ":py:obj:`DSMdo[g,t,tt]`",  "V",    "DSM up shift (additional load)"
```
p-snft: (Same as above.)

nailend:
Corrected most of the PEP8 recommendations as well as the parameter/variable description in the docstring.
I also changed deleted  the default setting as there is no "normal" way to use the component. It is always depends on the situation.

@uvchik 
I tried an alternative formulation without if-statements. The code is on  [gist](https://gist.github.com/nailend/609480d924c0548109540a416e204235).
The comparison of the creation of  a test model:

- 100 dsm components 

- 100 timesteps

 resulted in a difference of 0.07 seconds in favor of the if-statement formulation.
delay-method:
if-statements: 16,33
for-loop:16,39

So a change of the formulation doesn't seem reasonable


I also prepared a model for the constraint test. [Gist](https://gist.github.com/nailend/fd1e62aa3d97ea60b315abadc17686c2)

Does this work for you? Contact me if you need anything else.

Thanks a lot for the testing!

oemof/solph/custom.py
```diff
@@ -839,3 +844,582 @@ def tes_ub_constr_rule(block, n, t):
...
+    Based on the paper by Zerrahn, Alexander and Schill, Wolf-Peter (2015):
+    On the representation of demand-side management in power system models,
+    in: Energy (84), pp. 840-845, 10.1016/j.energy.2015.03.037,
+    accessed 17.09.2019, pp. 842-843.
+
+    A special sink component which modifies the input demand series.
+
+    Parameters
+    ----------
+    demand: int or array
```
p-snft: I assume, floats are allowed. We use to write "numeric (sequence or scalar)" in cases like these.

nailend:
@p-snft is there already a solution for the travis integration error ? It seems to be non-specif for this component, but rather a general problem with the cbc-coiner solver under python 3.7

p-snft:
It seems to be a problem with the dependency management inside TravisCI. It should not even be specific to Python3.7. I improved that part and hope, the problem is now solved.

uvchik:
> I tried an alternative formulation without if-statements. The code is on [gist](https://gist.github.com/nailend/609480d924c0548109540a416e204235).
> The comparison of the creation of a test model:
> 
>     * 100 dsm components
> 
>     * 100 timesteps
> 
> 
> resulted in a difference of 0.07 seconds in favor of the if-statement formulation.
> delay-method:
> if-statements: 16,33
> for-loop:16,39
> 
> So a change of the formulation doesn't seem reasonable

I agree.



> I also prepared a model for the constraint test. [Gist](https://gist.github.com/nailend/fd1e62aa3d97ea60b315abadc17686c2)

You do not have to create constraint tests because we already have them in solph. You just have to provide different use cases to cover possible variants of constraints (different method, constraints that will be build/not build if a parameter is set/not set).

nailend:
> > I tried an alternative formulation without if-statements. The code is on [gist](https://gist.github.com/nailend/609480d924c0548109540a416e204235).
> > The comparison of the creation of a test model:
> > ```
> > * 100 dsm components
> > 
> > * 100 timesteps
> > ```
> > 
> > 
> > resulted in a difference of 0.07 seconds in favor of the if-statement formulation.
> > delay-method:
> > if-statements: 16,33
> > for-loop:16,39
> > So a change of the formulation doesn't seem reasonable
> 
> I agree.
> 
> > I also prepared a model for the constraint test. [Gist](https://gist.github.com/nailend/fd1e62aa3d97ea60b315abadc17686c2)
> 
> You do not have to create constraint tests because we already have them in solph. You just have to provide different use cases to cover possible variants of constraints (different method, constraints that will be build/not build if a parameter is set/not set).
> 
> Please have a look at [4dbc17c](https://github.com/oemof/oemof/commit/4dbc17c010f97088c189ee1bee1c38c7aeaea5bb) and check the constraints. The example shows how to build a constraint test. At the first run you have to create an empty file, after that you can copy the temp file there.

uvchik:
If you want to add more constraint tests yourself, you can rename my test to something like `dsm_basic_delay_test()` or whatever you like. You could also add a docstring with one sentence to explain which parameter is set/unset.

uvchik:
Did you test you changes in 53d4536 with sphinx. As I know it you have to break the line within quotes but I am not totally sure.

nailend:
> Did you test you changes in [53d4536](https://github.com/oemof/oemof/commit/53d45369e45eb7f88f549b398cc379134a3c3f67) with sphinx? As I know it you have to break the line within quotes but I am not totally sure.

you are totally right! thanks

uvchik:
You can ignore the lgtm alert. It is not your fault.

nailend:
> If you want to add more constraint tests yourself, you can rename my test to something like `dsm_basic_delay_test()` or whatever you like. You could also add a docstring with one sentence to explain which parameter is set/unset.

I renamed your test and added one for the second method option, ran nosetests, checked and then added the lp files. anything missing?

oemof/solph/custom.py
```diff
@@ -839,3 +844,593 @@ def tes_ub_constr_rule(block, n, t):
...
+    accessed 17.09.2019, pp. 842-843.
+
+    * in 'delay' method: delay time might be extended artificially
+    if DSM-capacity isn't used to it's full extend.
+    see oemof/Issue: #622 for more information
+
+    """
+
+    def __init__(self, demand, capacity_up, capacity_down, method,
+                 *args, **kwargs):
```
uvchik: I do not know what *args is for. In my opinion you can safely remove it.

oemof/solph/custom.py
```diff
@@ -839,3 +844,593 @@ def tes_ub_constr_rule(block, n, t):
...
+        super().__init__(*args, **kwargs)
+
+        self.capacity_up = sequence(capacity_up)
+        self.capacity_down = sequence(capacity_down)
+        self.demand = sequence(demand)
+        self.method = method
+        self.shift_interval = kwargs.get('shift_interval', None)
+        self.delay_time = kwargs.get('delay_time', None)
+        self.recovery_time = kwargs.get('recovery_time', False)
```
uvchik: What is recovery time for? I do not find it anywhere else in the code or the docs.

oemof/solph/custom.py
```diff
@@ -839,3 +844,593 @@ def tes_ub_constr_rule(block, n, t):
...
+        delay :
+
+            Sophisticated model based on the formulation by
+            Zerrahn & Schill (DIW).T he load-shift of the component must be
+            compensated in a predefined delay-time (**delay_time is mandatory).
+            DSM capacity can either be a fixed value or an hourly time series.
+    **shift_interval: int
+
+        only mandatory in method = 'interval' :
+
```
uvchik: "Mandatory if method == 'interval'" and "Only valid if method == 'interval'. For me "only mandatory" sounds to me that you can use it for 'delay' but it is not mandatory.

oemof/solph/custom.py
```diff
@@ -839,3 +844,593 @@ def tes_ub_constr_rule(block, n, t):
...
+        interval in between which total DSM  must be fully compensated
+    **delay_time: int
```
uvchik: See above (two asterisks).

oemof/solph/custom.py
```diff
@@ -839,3 +844,593 @@ def tes_ub_constr_rule(block, n, t):
...
+
+        only mandatory in method =  'delay':
```
uvchik: see above

uvchik:
> I renamed your test and added one for the second method option, ran nosetests, checked and then added the lp files. anything missing?

If you check the resulting lp-files, this should be all for the tests.

See my small remarks above and thanks again.

oemof/solph/custom.py
nailend: both *args and **kwargs seem to be necessairy for "label". If I remove one it ens up in an error

oemof/solph/custom.py
nailend: a not used parameter, from and additional not working constraint I already removed. removed it.

uvchik:
Sorry, maybe I did not make myself clear. In my opinion the constructor should look like this:

```python
def __init__(self, demand, capacity_up, capacity_down, method,
             shift_interval=None, delay_time=None)
```

The parameters `shift_interval` and `delay_time` are optional and therefore `None` by default.

gplssm:
@uvchik told me on another channel that he's fine with this PR.

@uvchik  + @p-snft could you please click "approve". Then, we could merge it or you just do, once both of you approved.

If there still things to fix, don't hesitate to point that out!

uvchik:
I think it definitely has the quality to be used by others. We still need some API tests but I would recommend to open a new PR for that.

@nailend As you will have to describe it for you thesis you could also add a description, plot etc. [here](https://oemof.readthedocs.io/en/stable/oemof_solph.html#solph-components). But this is also a PR on its own.

If these two things are done we can move it from custom to components after a while.

Issue #597, opened 2019-06-04
Adapt docstrings to new template


uvchik:
In issue #544 we developed a new template for the doc string of block classes.

The following classes still have the old docstring design. 

 @oemof/documentation : Please assign yourself to one of the following components:

* [ ] block.Flow
* [x] block.InvestmentFlow
* [ ] block.Bus (@joroeder )
* [ ] block.Transformer (@uvchik) 
* [ ] block.NonConvexFlow
* [ ] component.GenericStorageBlock
* [ ] component.GenericInvestmentStorageBlock

Good examples are:

* components.ExtractionTurbineCHPBlock
* components.GenericCHPBlock

oakca:
I am checking the `components.GenericStorageBlock` and `components.GenericInvestmentStorageBlock `at the moment.

stianchris:
While trying to understand the issues, I just wanted to note down some annotations for those, that implement the template:
- if there are no variables created, the headline **The following variables are created:** is omitted (other possibility could be, to integrate also this headline and just write None)
- it is sometimes not clear (for people like me, that are not familiar with the code), when variables are classified, that they are created. Some are created to be able to write human readable code - to me it was not clear, if this is consistent for each of the components - is this the case?
- it also seems like the table of variables and parameters replaces the above mentioned headline. If this is the case - maybe an update of the template could be posted in this issue?

joroeder:
There are few good examples:
[OffsetTransformerBlock](https://oemof.readthedocs.io/en/stable/api/oemof.solph.html#oemof.solph.components.OffsetTransformerBlock)
[ExtractionTurbnineCHPBlock](https://oemof.readthedocs.io/en/stable/_modules/oemof/solph/components.html#ExtractionTurbineCHPBlock)
[GenericCHPBlock](https://oemof.readthedocs.io/en/stable/_modules/oemof/solph/components.html#GenericCHPBlock)
[GenericStorageBlock](https://oemof.readthedocs.io/en/stable/_modules/oemof/solph/components.html#GenericStorageBlock) ->maybe there needs to be still some small adaptions

There has not been written a template yet. This was the aim of #544 , but the status was that there was an agreement, but this has not been written down somewhere in a general form (e.g. at [https://oemof.readthedocs.io/en/stable/developing_oemof.html](https://oemof.readthedocs.io/en/stable/developing_oemof.html))

The aim of this issue was to change all docstrings which do not have the format yet. So we need some more people writing their names behind the blocks (see above)

[...]

PR #636, opened 2019-10-11
Features/add non convex investment flow

### LINKED ISSUE #628: Initial investment costs
I'm trying to implement initial costs for a PV component which holds an investment optimization.
For example, add 100€ to PV costs if (and only if) PV is built:
0 kW -> 0€
x kW -> 100€ + x kW * 2 €/kW

As far as I can see, there is not yet an option for that in oemof.
I tried to implement an additional `InitialInvestmentFlow` but as this leads to 
`RuntimeError: Cannot write legal LP file.  Objective 'objective' has nonlinear terms that are not quadratic.`
Is it even possible to implement it?
Some hints would be nice!

See my code example below.
Problematic line is:
```
initial_costs += (
    (sum(m.flow[i, o, t] for t in m.TIMESTEPS) > 0) *
    m.flows[i, o].investment.initial_costs
)
```
If I remove "greater than" part, some costs are added (not what I want, but only to check if custom flow works)

**Example script:**
```
"""
Test scratch to implement pv initial invest
"""

import pandas
from oemof.solph import EnergySystem, Bus, Flow, Sink, Source, Investment
from oemof.solph import Model
from oemof import outputlib

from pyomo.core import Set, Expression
from pyomo.core.base.block import SimpleBlock
import oemof.groupings as groupings


class InitialInvestmentFlow(SimpleBlock):
    def _create(self, group=None):
        if group is None:
            return None

        self.FLOWS = Set(initialize=[(g[0], g[1]) for g in group])

    def _objective_expression(self):
        if not hasattr(self, 'FLOWS'):
            return 0

        m = self.parent_block()
        initial_costs = 0

        for i, o in self.FLOWS:
            if m.flows[i, o].investment.initial_costs is not None:
                initial_costs += (
                    (sum(m.flow[i, o, t] for t in m.TIMESTEPS) > 0) *
                    m.flows[i, o].investment.initial_costs
                )
            else:
                raise ValueError("Missing value for fixed investment costs!")

        self.initial_costs = Expression(expr=initial_costs)
        return initial_costs

def _initial_investment_grouping(stf):
    if hasattr(stf[2], 'investment'):
        if stf[2].investment is not None:
            return True
    else:
        return False

initial_investment_flow_grouping = groupings.FlowsWithNodes(
    constant_key=InitialInvestmentFlow,
    filter=_initial_investment_grouping
)


energysystem = EnergySystem(
    timeindex=pandas.date_range('2019-01-01', periods=2, freq='H'),
    groupings=[initial_investment_flow_grouping]
)
bus = Bus()
net = Source(
    label="Net",
    outputs={
        bus: Flow(nominal_value=90)
    }
)
investment = Investment(ep_costs=2)
investment.initial_costs = 100
pv = Source(
    label="PV",
    outputs={
        bus: Flow(
            actual_value=[0.2, 0.2],
            fixed=True,
            investment=investment,
        )
    }
)
demand = Sink(
    label="demand",
    inputs={
        bus: Flow(
            nominal_value=1,
            actual_value=[100, 100],
            fixed=True
        )
    }
)
excess = Sink(
    label="excess",
    inputs={bus: Flow()}
)

energysystem.add(bus, net, pv, demand, excess)
om = Model(energysystem=energysystem, constraint_groups=[InitialInvestmentFlow])
om.solve(
    solver="cbc",
    solve_kwargs={'tee': True, 'keepfiles': True}
)

results = outputlib.processing.results(om)
# import pdb;pdb.set_trace()
print(results)
```

joroeder:
Fix #628 

This PR proposes the option of having fix investment costs additionally to capacity dependent investmentcost ...

C_invest = C_fix + P_install * C_var

... by using an additional binary variable. (Analogy to the OffsetTransformer for Investments)

The ```InvestmentFlow``` class in solph/blocks.py and the ```Investment``` solph/options.py are adapted.

The API could look like:

```python
trafo = solph.Transformer(
    label='transformer',
    inputs={bus_0: solph.Flow()},
    outputs={bus_1: solph.Flow(
        nominal_value=None,
        investment=solph.Investment(
            ep_costs=200,
            maximum=100,
            minimum=10,
            nonconvex=True,
            offset=1000),
        )},
    conversion_factors={bus_1: 0.8})
```

The idea behind the two attributes ```nonconvex``` and ```offset``` is, that a mimimum investment can be used without applying an offset. In this case the mimimum investment is not forced, it can be 0 as well. Maybe there is another option of implementing that ...

* [x] Adapt docstring
* [x] API
* [x] Tests
* [x] Documentation
* [x] Add to whatsnew

Example:
[example_offsetinvest.txt](https://github.com/oemof/oemof/files/3718302/example_offsetinvest.txt)


uvchik:
We may want to catch the case: `offset=0, minimum=0, nonconvex=True`

We still need some tests for `existing != 0` and for `ep_costs is None` in the `GenericStorage` and `Flow` class.

joroeder:
Hi Uwe, (@uvchik ), a missing 'maximum' attribute causes some problems when it comes to processing the results. I have no idea, whats the problem ...

I think it is a good idea to force a finite numeric value as maximum in case of non-convex investment. There is another numeric problem: If a very high value is set for maximum investment, no investment is done. Daniel demonstrated this in his example 'os_invest_test.py':
[https://github.com/oemof/oemof-examples/blob/non_convex_investment/oemof_examples/oemof.solph/v0.3.x/invest_non_convex/os_invest_test.py](https://github.com/oemof/oemof-examples/blob/non_convex_investment/oemof_examples/oemof.solph/v0.3.x/invest_non_convex/os_invest_test.py)

In the example, if a maximum value of ```maximum=60000000``` is applied, no investment is done, although it would be the optimal solution ...

chrklemm:
Hello everybody! I am very interested in this feature. Thank you very much for taking the trouble to develop it!

oemof/solph/blocks.py
```diff
@@ -245,62 +247,77 @@ class InvestmentFlow(SimpleBlock):
...
 
-    FLOWS
+    INVESTFLOWS
         A set of flows with the attribute :attr:`invest` of type
         :class:`.options.Investment`.
+    CONVEX_INVESTFLOWS
+        A subset of flows with the attribute :attr:`nonconvex` set to `False`.
+    NON_CONVEX_INVESTFLOWS
+        A subset of flows with the attribute :attr:`nonconvex` set to `True`.
     FIXED_FLOWS
```
joroeder: "FIXED_INVESTFLOWS" ?

oemof/solph/blocks.py
```diff
@@ -245,62 +247,77 @@ class InvestmentFlow(SimpleBlock):
     **The following sets are created:** (-> see basic sets at
```
joroeder: Do we want to have all these Sets in the docstring? We don't have the sets in the docstring of the more "applied" components, which serve as docstring templates (like genericCHP, extractionTurbine,..), but I am not sure about the more fundamental blocks like flow etc.

oemof/solph/blocks.py
```diff
@@ -245,62 +247,77 @@ class InvestmentFlow(SimpleBlock):
...
         A set of flows with the attribute :attr:`invest` of type
         :class:`.options.Investment`.
+    CONVEX_INVESTFLOWS
+        A subset of flows with the attribute :attr:`nonconvex` set to `False`.
+    NON_CONVEX_INVESTFLOWS
+        A subset of flows with the attribute :attr:`nonconvex` set to `True`.
     FIXED_FLOWS
-        A set of flow with the attribute :attr:`fixed` set to `True`
+        A set of flow with the attribute :attr:`fixed` set to `True`.
     SUMMED_MAX_FLOWS
```
joroeder: In this logic, it must be SUMMED_MAX_INVESTFLOWS
- actually, one basic question, because I get confused after a while, if there is the 'Investment' Attribute in the flow, there is no `Flow(SimpleBlock)` but an `InvestmentFlow(SimpleBlock)` build, so this is why e.g. the summed_min and summed_max rule need to be in the `InvestmentFlow(SimpleBlock)` class as well, right?

oemof/solph/blocks.py
```diff
@@ -245,62 +247,77 @@ class InvestmentFlow(SimpleBlock):
...
-        A subset of set FLOWS with flows with the attribute :attr:`summed_max`
-        being not None.
+        A subset of set INVESTFLOWS with flows with the attribute
+        :attr:`summed_max` being not None.
     SUMMED_MIN_FLOWS
-        A subset of set FLOWS with flows with the attribute
+        A subset of set INVESTFLOWS with flows with the attribute
         :attr:`summed_min` being not None.
     MIN_FLOWS
```
joroeder: MIN_INVESTFLOWS ?

oemof/solph/blocks.py
```diff
@@ -381,9 +428,9 @@ def _max_investflow_rule(block, i, o, t):
             """
             expr = (m.flow[i, o, t] <= (
                 (m.flows[i, o].investment.existing + self.invest[i, o]) *
-                 m.flows[i, o].max[t]))
+                m.flows[i, o].max[t]))
             return expr
-        self.max = Constraint(self.FLOWS, m.TIMESTEPS,
+        self.max = Constraint(self.INVESTFLOWS, m.TIMESTEPS,
```
joroeder: why not CONVEX_INVESTFLOWS ?

oemof/solph/blocks.py
```diff
@@ -355,23 +378,47 @@ def _create(self, group=None):
...
+                    self.invest_status[i, o] <= self.invest[i, o])
+            return expr
+        self.minimum_rule = Constraint(
+            self.NON_CONVEX_INVESTFLOWS, rule=_min_invest_rule)
+
+        def _max_invest_rule(block, i, o):
+            """Rule definition for applying a minimum investment
+            """
+            expr = self.invest[i, o] <= (
+                m.flows[i, o].investment.maximum * self.invest_status[i, o])
```
joroeder: here, we could also multiply with `m.flows[i, o].max[t]` ?

oemof/solph/blocks.py
```diff
@@ -355,23 +378,47 @@ def _create(self, group=None):
...
             """Rule definition for bounds of invest variable.
             """
-            return (m.flows[i, o].investment.minimum,
-                    m.flows[i, o].investment.maximum)
-        # create variable bounded for flows with investement attribute
-        self.invest = Var(self.FLOWS, within=NonNegativeReals,
+            if (i, o) in self.CONVEX_INVESTFLOWS:
+                return (m.flows[i, o].investment.minimum,
+                        m.flows[i, o].investment.maximum)
+            elif (i, o) in self.NON_CONVEX_INVESTFLOWS:
```
joroeder: Is this really necessary? Below, there is the _max_invest_rule which already sets the upper bound for the nonconvex case. Are do we need some bounds here?

oemof/solph/blocks.py
joroeder: replace in code

oemof/solph/blocks.py
joroeder: replace in code

oemof/solph/blocks.py
joroeder: no, INVESTFLOWS is right

oemof/solph/blocks.py
joroeder: for now, we need the definition of the bounds

joroeder:
> 
> We may want to catch the case: `offset=0, minimum=0, nonconvex=True`

Is this a really a problem? There is just an unnecessary binary variable created, the results will be the same. That could be a case for an user warning.

> 
> We still need some tests for `existing != 0` and for `ep_costs is None` in the `GenericStorage` and `Flow` class.

@uvchik For nonconvex investments? or do you mean constraint tests for the "regular" Investment?
In case of nonconvex, for now, we don't want to allow existing + nonconvex investment, right?

ep_costs are per default 0, or do you think about the case, if somebody writes: `ep_costs=None`?

Actually, I wonder, what tests we really need? We could write constraint tests for all combinations:
{Storage, Source, Sink, Transformer} x {nonconvex, convex} x {...} ...

joroeder:
hey, I completely revised the docstring of the InvestmentStorageBlock. From my perspective, everythings fine. ;)

Could please somebody review this PR :)

Since this feature has already been left undone for quite a while, and we would like to consider some functionalities of this branch as option in another oemof package ([https://github.com/oemof/DHNx/pull/29](https://github.com/oemof/DHNx/pull/29)), I would be really happy, if we could finish this one soon :)

oemof/solph/blocks.py
```diff
@@ -248,16 +248,19 @@ class InvestmentFlow(SimpleBlock):
 
     **The following variables are created:**
 
-    * :math:`F_{invest}`  :attr:`om.InvestmentFlow.invest[i, o]`
+    * :attr:`invest[i, o]` - :math:`F_{invest}`
```
p-snft: I personally like this, but it is not perfectly aligned with the typical docstring, which does not name the symbol here (see e.g. https://oemof.readthedocs.io/en/stable/api/oemof.solph.html#oemof.solph.components.GenericStorage).

joroeder: That's right, it is a new suggestion for docstring style, since most of the docstrings of the basic blocks need a revision anyway to adapt to the new style (#597).

In the meantime, I thought about putting the symbols to parameters at each class as well (e.g. for investment options:  [https://oemof.readthedocs.io/en/features-add_nonconvexinvestmentflow/api/oemof.solph.html#oemof.solph.options.Investment](https://oemof.readthedocs.io/en/features-add_nonconvexinvestmentflow/api/oemof.solph.html#oemof.solph.options.Investment)), but then, I noticed that these parameter symbols might be different in each block, where they are used. E.g. `ìnvestment.maximum` can be F_{invest,max} for flows or E_{invest,max} for storages. However, I still like the idea of having the symbols there as well. If you go through the API of readthedocs, then, it would be really easy for the reader to connect the content quickly. But for sure, this makes it a lot more prone to errors and inconsistencies. What do you think about that?

To reduce dependencies of PRs and issues and some decisions, do you think it is an option to leave the docstring for now, and make it consistent before v0.0.4? I could take care of that.

joroeder: *of course I meant  v0.4.0

joroeder:
@uvchik I am wondering, is there anything against merging besides docstring issues?

uvchik:
> @uvchik I am wondering, is there anything against merging besides docstring issues?

Docs are fixed. Two approvals. Go ahead.

joroeder:
@uvchik many thanks for your great support! @p-snft many thanks for your review!

PR #642, opened 2019-10-25
Correct storage investment example

jnnr:
When investing into storage capacity, no nominal_storage_capacity should be set.

### DIFF
--- a/doc/oemof_solph.rst
+++ b/doc/oemof_solph.rst
@@ -596,7 +596,6 @@ The following example describes a battery with flows coupled to the capacity of 
         inputs={b_el: solph.Flow()},
         outputs={b_el: solph.Flow()},
         loss_rate=0.001,
-        nominal_storage_capacity=50,
         inflow_conversion_factor=0.98,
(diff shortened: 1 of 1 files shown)


uvchik:
As this PR does not change the API you may want to rebase it against the `0.3` branch. the `0.3` branch is part of the `dev` branch but the `dev`-branch is not part of the `0.3` branch. As this is a fix it should be in both branches and therefore in `0.3`.

p-snft:
Was #641 a duplicate of this PR?

jnnr:
Hi there, and sorry for the confusion. I was a little confused by the new branching scheme and branched this from dev instead of the appropriate v0.3, as this is just a one-line docs correction and (without api change :) ). I wanted to clean this up but now you were quicker. Thanks! And indeed, #641 is a duplicate.

jnnr:
Should the base of the branch not be v0.3, @p-snft? It is just a docs correction, no API-change involved.

uvchik:
I moved the commit to the v0.3 branch.

PR #649, opened 2019-11-13
fix SPDX-License-Identifier entry

fwitte:
I found some entries using the GPL 3 License and replaced them with the MIT License, did I overlook any?


uvchik:
I did the same changes in the v0.3 branch.

PR #656, opened 2019-11-23
Fix/#655 cost assignment

### LINKED ISSUE #655: Variable cost for Flow at SinkDSM make no sense
As already mentioned in the re-opened #622, assigning `variable_cost` to a Flow that connects a SinkDSM, as for example done [here](https://github.com/oemof/oemof/blob/08bedbace421a624bbd9af5146f2cdfd2f61a591/tests/constraint_tests.py#L657), makes no sense. These cost would be assigned to every kWh that comes through this Flow to serve the load. It has no effect on triggering an DSM event or not.

In order to change this behaviour, 3 options are available

1. Assign cost only to DSM_up as this variable represents increased demand. So far, so easy. But, this would only work for `method` "delay", because in case of `method` "interval" on DSM_updown exists, which includes positive and negative demand shifts and sums to zero over the entire simulation time.
Could be achieved by adding a `_objective_expression()` method as for example [here](https://github.com/oemof/oemof/blob/08bedbace421a624bbd9af5146f2cdfd2f61a591/oemof/solph/blocks.py#L212-L239)
2. Assign no cost and either
    1. Warn about it
    2. Forbid it (if possible)

gplssm:
Fix #655 

Another problem of the initial implementation of method "interval" of SinkDSM was solved by this change. Early, the bounds of SinkDSMIntervalBlock_dsm_up_down were applied to all timesteps like

   -0.5 <= SinkDSMIntervalBlock_dsm_up_down(demand_dsm_0) <= 0.5
   -0.5 <= SinkDSMIntervalBlock_dsm_up_down(demand_dsm_1) <= 0.5
   -0.5 <= SinkDSMIntervalBlock_dsm_up_down(demand_dsm_2) <= 0.5
   -0.5 <= SinkDSMIntervalBlock_dsm_up_down(demand_dsm_3) <= 0.5
   ...

Actually, the bounds in the LP file should look like

   -0.5 <= SinkDSMIntervalBlock_dsm_up_down(demand_dsm_0) <= 0.5
   -0.4 <= SinkDSMIntervalBlock_dsm_up_down(demand_dsm_1) <= 0.4
   -0.5 <= SinkDSMIntervalBlock_dsm_up_down(demand_dsm_2) <= 0.5
   -0.4 <= SinkDSMIntervalBlock_dsm_up_down(demand_dsm_3) <= 0.4
   ...

While updating the test data "dsm_module_interval.lp", I recognized another bug. The constraint "SinkDSMIntervalBlock_dsm_sum_constraint" was build for each time step with the interval. This was changed to one constraint per interval.


gplssm:
@jnnr we can set the review request on hold. In #660 I created a PR to merge SinkDSM code into release v0.3.2. If this release is merged to dev anyways, we can simply close this PR.

gplssm:
The code was already integrated to `dev` by be206955e2400713cc025b44ef8c944f27a24913

PR #658, opened 2019-11-27
Replace travis CI tests of python 3.5 with 3.8

uvchik:
I would remove Python 3.5 support for the next major :exclamation: release. That does not mean that oemof will not work with python 3.5 any more but it will not be tested any more. In the long run code that is not compatible with 3.5 will not be detected. The official support for 3.5 will end in September 2020. Python 3.6 is backward compatible so it should be no problem for anybody to update from python 3.5 to 3.6.


uvchik:
The idea of opening an old dump should ensure that it is possible to open a dump from a previous minor version. I skipped the test because the problem is not caused by solph but by the python version. We can try to reactivate the test with the next major release.

Issue #646, opened 2019-11-04
Docs could use a review


jnnr:
This issue is related to tespy's https://github.com/oemof/tespy/issues/87

1. Check for a consistent structure
2. update old information, repair broken links and fix errors,
3. fill a glossary.

We aim to tackle this one at the next [oemof-dev meeting](https://github.com/oemof/organisation/wiki/Meeting-December-2019)

jnnr:
Related issues in oemof-solph on docstrings
#596 #479

jnnr:
Agenda for that issue (adapted from tespy docs review)

**Documentation in oemof/doc**

In the online documentation we need to check for

- typos,
- comprehensibility (would a new user understand the main ideas?) and
- structure (does the structure of the chapters help the user understand how to use oemof-solph?).

- [ ] Getting started
- [ ] About oemof
- [ ] Installation and setup
- [ ] Using oemof
- [ ] Developing oemof
- [ ] oemof-network
- [ ] oemof-solph
- [ ] oemof-outputlib
- [ ] oemof-tools

The same criteria should be checked in:

- [ ] README.rst

**Docstrings and inline comments**

In the doc-strings and the inline comments we need to check for the following:
- Are there any typos?
- Are the examples well-formatted, correct and comprehensible?
- Is information well placed (are the equations and parameters apparent)?
- Are the links in the doc-strings working?
- Are the inline comments clear?

- oemof/outputlib
  - [ ] processing.py
  - [ ] views.py

- oemof/solph
  - [ ] blocks.py
  - [ ] components.py
  - [ ] constraints.py
  - [ ] custom.py
  - [ ] groupings.py
  - [ ] models.py
  - [ ] network.py
  - [ ] options.py
  - [ ] plumbing.py

- oemof/tools/
  - [ ] economics.py
  - [ ] helpers.py
  - [ ] logger.py

- [ ] energy_system.py
- [ ] graph.py
- [ ] groupings.py
 - [ ] network.py

joroeder:
Do we want to have consistent AE or BE?

jnnr:
# Getting started
> Contribution is already possible on a low level
 Workflow could be described in more detail.

## Installing oemof 
Installing oemof is double: In getting started and in installation and setup.

## Structure of the oemof cosmos
Hard to read. Bullet points would be more readable. Could be include the figure showing the layers from Hilpert 2018? This section could be moved somewhere else if oemof and solph would be unbundled.

[...]

PR #661, opened 2019-12-02
Solph docs review

jnnr:
Addresses #646. We will use this branch at the upcoming winter dev meeting 2019 in the docs review hackathon session. Corrections can either be committed to the branch directly or alternatively be added as code comments as part of a review. If you are not yet a member of the oemof organisation, please

- fork oemof to your profile
- checkout the fix/review_docs branch
- push your local commits to your fork
- open a PR to merge your changes into fix/review_docs branch in oemof.

The participant who finds the most typos and mistakes will get a little prize.

### DIFF
--- a/doc/about_oemof.rst
+++ b/doc/about_oemof.rst
@@ -15,7 +15,7 @@ The idea of an open framework
 
 The Open Energy System Modelling Framework has been developed for the modelling and analysis of energy supply systems considering power and heat as well as prospectively mobility.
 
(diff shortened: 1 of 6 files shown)


jnnr:
The main change here is replacing optimisation with optimization (British English to American English). We did not see that written down anywhere, but in a webco it was mentioned that there is the convention to use BE in oemof. So the changes here should not be merged.

Originally, it was thought to collect more improvements regarding the docs here, but it seems that is all that was collected.

Therefore, I close this PR.

PR #662, opened 2019-12-05
Fix/review doc - fix some AE/BE typos

joroeder:
Fix some AE/BE typos in the 'getting_startet' and creating pull request for test purpose

### DIFF
--- a/doc/about_oemof.rst
+++ b/doc/about_oemof.rst
@@ -15,7 +15,7 @@ The idea of an open framework
 
 The Open Energy System Modelling Framework has been developed for the modelling and analysis of energy supply systems considering power and heat as well as prospectively mobility.
 
-Oemof has been implemented in Python and uses several Python packages for scientific applications (e.g. mathematical optimisation, network analysis, data analysis), optionally in combination with a PostgreSQL/PostGIS database.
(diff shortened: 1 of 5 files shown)


uvchik:
As we decided to use BE in oemof I thought that optimisation is right. I think even in BE optimization is fine, too, but it seems to be less common in BE.

uvchik:
I just saw that this is already merged. As the 'z'-writing is allowed in BE it might not be worth to revert it but I am still confused.

joroeder:
Oh, sorry. When we saw in the hands-on review docs session on the last dev meeting, that it seemed that both variants were used, we thought, that there was no decision yet. If we want to use BE, we should revert it, I agree.

> As we decided to use BE in oemof
>

Where do we find such decisions? readthedocs or wiki? Actually, to have two places you need to check before doing something, is not very convenient in my opinion. Why don't we put all the stuff from the oemof/oemof wiki in readthedocs?

uvchik:
You are totally right. This should be part of the documentation but it isn't (I thought it is).

There are two sections about developing the documentation which is not good:

- https://oemof.readthedocs.io/en/latest/developing_oemof.html#contribute-to-the-documentation

- https://oemof.readthedocs.io/en/latest/developing_oemof.html#documentation

We should merge them and add the information that we use BE.

The Wiki is only for special or non-static information. For example the "how to release" Wiki is not important for somebody who wants to contribute.

PR #663, opened 2019-12-06
Features/user warnings for debugging

uvchik:
A UserWarning should be raised for untypical uses even though this kind of usage is valid if you really know what you are doing, whereas in 90% of the cases it would be a mistake. This will help users to debug their code but can be turned of for experienced users.

* Related issues #670 

* [x] Add to the documentation!
* [x] Add tests


joroeder:
I think this is a really nice feature! Thanks!

Regarding to #670, I am wondering whether the label of the Transformer/Source/etc. can be included in the warning? Somethink like:

```ruby
            warn(
                "Transformer:  `" + self.label +
                "` constructed without `inputs`.\n",
                debugging.SuspiciousUsageWarning,
            )
```

uvchik:
Good idea. Now it looks like this.

```python
msg = "`Sink` '{0}' constructed without `inputs`."
warn(msg.format(self), debugging.SuspiciousUsageWarning)
```
If you now write `imputs` instead of `inputs`a warning will be raised:

```python
look_out = network.Bus()
network.Sink(label='test_sink', imputs={look_out: "A typo!"})
```
Leads to:
```diff
- SuspiciousUsageWarning: `Sink` 'test_sink' constructed without `inputs`.
```

uvchik:
I wonder where to put this into the documentation.

The most important information is how to use the three modes (warning, error, none).
https://github.com/oemof/oemof/issues/670#issuecomment-572055006

joroeder:
One foolproof option would be to give the additional hint like: "There might be a misspelling in ```inputs``` / ```outputs``` ", because I think that is most likely.


> I wonder where to put this into the documentation.

How about in the general section "How can I use solph?" in a new paragraph like "Handling of errors and warnings" ? Or more specific "Handling of SuspiciousUsageWarning" ?

oemof/tools/debugging.py
```diff
@@ -0,0 +1,28 @@
...
+    Some ways of using `oemof` are not necessarily wrong but could lead to
+    hard to find bugs if you done accidentally instead of intentionally. We
+    use these warnings, and you can do too ;), in your code to warn users about
+    these cases. If you know what you are doing and these warnings point you to
+    things you are doing intentionally, you can easily switch them off.
+
+    Examples
+    --------
+    >>> import warnings
+    >>> warnings.filterwarnings("ignore", category=SuspiciousUsageWarning)
```
p-snft: Replying to https://github.com/oemof/oemof/pull/663#issuecomment-572254355, more documentation might be added here. Also, a link to https://docs.python.org/3.7/library/warnings.html#the-warnings-filter might be put, as functionality is inherited from there. Maybe

> For options different from "ignore", see [Python documentation on warnings](https://docs.python.org/3.7/library/warnings.html#the-warnings-filter).

uvchik:
> How about in the general section "How can I use solph?" in a new paragraph like "Handling of errors and warnings" ?

I like the idea of adding an error section to the solph documentation. We could also add a link to the [FOE wiki](https://github.com/oemof/oemof/wiki/Frequently-occurring-errors).

We should add the following information and a link to the warning section of the python documentation as @p-snft suggested. Within the warning we could think of adding a link to the "Handling of errors and warnings" section in our documentation if it exists.

```python
# switch on SuspiciousUsageWarning
warnings.filterwarnings("always", category=SuspiciousUsageWarning)

# raise an error instead of a warning
warnings.filterwarnings("error", category=SuspiciousUsageWarning)

# switch off SuspiciousUsageWarning
warnings.filterwarnings("ignore", category=SuspiciousUsageWarning)
```

uvchik:
Please check the docs for typos.

I would like to merge soon to make  the warning available for PRs.

uvchik:
Okay, i fixed the conflicts and errors.

@p-snft, @joroeder Please have a brief look at the [docs](https://github.com/oemof/oemof/pull/663/files#diff-68101c16c99df97289607622114f4a02) and [docstring](https://github.com/oemof/oemof/pull/663/files#diff-d593957360bc11347281a9f20ef45b56).

uvchik:
TODO: Make sure that classes that inherit from e.g. Transformer will not raise the warnings.

joroeder:
I approve as well. My only suggestion for improvement is the position of the section in the documentation: maybe, it is better to put it at the end of "How can I use solph?" after "Analysing your results". When somebody starts to read how to use oemof.solph, the handling of warning is maybe not the first think you want to know.

Issue #187, opened 2016-07-11
Entity class still exists


uvchik:
The entity class still exist and is still needed although it should be replaced by the node class.

gnn:
I'll throw this one out for the next release.

gnn:
Cleanup tasks. Not hard, but also never top priority. Next release is a realistic target.

simnh:
Will you do @gnn ?

gnn:
Still on it. Doesn't really change the API, so in order to not block 0.2.0, I'm moving it to 0.2.1.

jnnr:
It seems like the Entity class is still in use. I guess I will shift this issue to the 0.2.2?

beierd:
I looked at the entity class and it indeed seems as if the entitity class is not in use anymore and can be easily deleted. Entitities is still part of the EnergySystem-class but seemingly unrelated to the entity class itself. 

Within the energy_system module parts of the documentation might be misleading as they refer to e.g. <oemof.core.network.Entity>. As I cannot really judge whether the documentation is largely outdated or if minor changes will do the trick I would like to leave this to someone with a deeper understanding. @gnn @uvchik. I blockcommented the relevant docstrings. 

I will create a pull request with the changes so you can have a look.

PR #665, opened 2019-12-10
Fix/remove entity class

beierd:
This pull request is meant to fix the issue #187 ("Entity class still exists") 

Basically I deleted the entity class from network.py and deleted some imports of the class. The comments in basic_tests.py are in order for the test to "run". As the class Entity doesn't exist anymore it cannot be used within the test. Most likely the test whole test is also obsolete? 

Also I'm pretty sure the docstring of the EnergySystem class is outdated, but I lack the understanding to fix the text. 

I would suggest you take it from here @gnn :)

### DIFF
--- a/oemof/energy_system.py
+++ b/oemof/energy_system.py
@@ -81,7 +81,6 @@ class EnergySystem:
     <oemof.core.network.Entity>` will always be grouped by their :attr:`uid
     <oemof.core.network.Entity.uid>`:
 
-    >>> from oemof.network import Entity
     >>> from oemof.network import Bus, Sink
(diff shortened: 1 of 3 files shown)


uvchik:
We have to transfer this PR to the oemof.network package. Do you need help?

uvchik:
https://github.com/oemof/oemof.network/issues/4

uvchik:
I changed the milestone, so that we do not forget about this issue but we really should transfer it because it does no longer belong to oemf-solph.

uvchik:
Transferred.

PR #666, opened 2019-12-10
Features/schedule for flow

c-koehl:
Added a schedule for Flow 
  * Functionality
  Sometimes it´s important to set a schedule for a flow (e.g. CHP electrical output), e.g. for fulfilling contracts. During optimization flow will be pushed to fulfill the schedule, if possible. If it´s not possible a penalty term will be activated and added to the objective expression for each unit of the difference flow-schedule.
  * Changes made in **oemof.solph.network**

    Parameter added to class **Flow**:
    * _schedule_
     Flow has to follow the schedule, but LP won´t be infeasible, if flow can´t fulfill schedule: slack variables will be added (see  _Changes made in **solph.blocks**_)
    * _penalty_pos_
    Describes the costs associated with one unit of the deficit of the flow (:= _slack_pos_) compared to the schedule. Costs will be added to the objective expression.
    * _penalty_neg_
    Describes the costs associated with one unit of the excessed flow (:= _slack_neg_).  Will be added to the objective expression.
 * Changes made in **oemof.solph.blocks**

    Added Set and Variables to class **Flow**:
    * _SCHEDULE_FLOWS_ (**pyomo.Set**)
    contains all flows with schedules

    * _slack_neg_ (**pyomo.Var**, within=NonNegativeReals)
    slack variable for representing excess of flow compared to schedule (see _Added constraint to class **Flow**_)
    * _slack_pos_ (**pyomo.Var**, within=NonNegativeReals)
    slack variable for representing deficit of flow compared to schedule

    Added constraint to class **Flow**
    * flow(i, o, t) - _slack_neg_(i, o, t) + _slack_pos_(i, o, t) = _schedule_(i, o, t)
  
    Term added to objective expression:
    * penalty_pos(i, o, t) * slack_pos(i, o, t) + penalty_neg(i, o, t) * slack_neg(i, o, t)
 * Changes made in **oemof.tests.test_processing**
    New Parameters of class **Flow** had to be added to the test file as well.
 * Changes made in **oemof.tests.constraint_tests**
    Test for scheduled flow was added.
 * Changes made in **oemof.tests.lp_files.flow_schedule.lp**
    Created file.

 * Changes made in **doc.whatsnew.v0-3-3.rst**
    Created file.

* Possible Improvements
    * Default value for _penalty_pos_ and _penalty_neg_ is 0. This could be improved by setting a reasonable default value, but this acquires mathematical research on how to choose appropriate penalties.
    * If it's not possible for the flow to follow the schedule, "non-real" costs will be added to the objective expression. You might not want to interfere the artificial costs (added by the slack variables) with the real costs of the solved LP.


c-koehl:
If you're interested, I could also add an example to oemof-examples.

gplssm:
Once you're done with the implementation, select a reviewer who cross-checks the newly added code. Preferably, select someone who already has some context about this

c-koehl:
@gplssm I'm done with the implementation, but it seems like I don't have the rights to select a reviewer or labels. I don't know what I'm doing wrong? Also I cannot tell, which reviewer is the best to choose

gplssm:
> @gplssm I'm done with the implementation, but it seems like I don't have the rights to select a reviewer or labels. I don't know what I'm doing wrong? Also I cannot tell, which reviewer is the best to choose

I'm not too familiar with permissions granted to individual devs. Maybe @uvchik can help out here.

Meanwhile, I can request a review for you. Did you talk to anyone before/during your implementation of "schedule for flow"? I would prefer to select this person then.

c-koehl:
> > @gplssm I'm done with the implementation, but it seems like I don't have the rights to select a reviewer or labels. I don't know what I'm doing wrong? Also I cannot tell, which reviewer is the best to choose
> 
> I'm not too familiar with permissions granted to individual devs. Maybe @uvchik can help out here.
> 
> Meanwhile, I can request a review for you. Did you talk to anyone before/during your implementation of "schedule for flow"? I would prefer to select this person then.

@gplssm I actually didn't talk to anyone. I would suggest a reviewer with mathematical optimization knowledge, but I think my procedure is also comprehensible for non mathematicians.

gplssm:
I've requested a review by @simnh because he's quite deep into mathematical modeling in oemof.solph.
@simnh please re-request a review by someone else if you can't do it.

oemof/solph/blocks.py
```diff
@@ -28,6 +28,14 @@ class Flow(SimpleBlock):
         Difference of a flow in consecutive timesteps if flow is increased
         indexed by NEGATIVE_GRADIENT_FLOWS, TIMESTEPS.
 
+    slack_pos :
+        Difference of a flow to schedule in consecutive timesteps if flow
+        has deficit to schedule. Indexed by SCHEDULE_FLOWS, TIMESTEPS.
+
+    slack_neg :
```
simnh: Maybe these could be named to schedule_slack_pos, but I am not sure...It's also fine for me like this. Naming is always so controversial ;-)

oemof/solph/network.py
```diff
@@ -117,6 +117,23 @@ class Flow(on.Edge):
         :class:`Flow <oemof.solph.blocks.Flow>`.
         Note: at the moment this does not work if the investment attribute is
         set .
+    schedule : numeric (sequence or scalar)
+        Schedule for the flow. Flow has to follow the schedule, otherwise a
+        penalty term will be activated. If array-like, values can be None
+        for flexible, non-fixed flow during the certain timestep.  Used in
```
simnh: Does this work actually, i.e. setting someting like `schedule=[1, 2, None, 3]`. Has this been tested. I thought pyomo would give an error if None is provided in the equation. Please check if to be sure.

oemof/solph/network.py
```diff
@@ -117,6 +117,23 @@ class Flow(on.Edge):
...
+        combination with :attr:`penalty_pos` and :attr:`penalty_neg`.
+    penalty_pos : numeric (sequence or scalar)
```
simnh: Should there be  a high default value of something like 10e9 force the schedule kind of by default?

oemof/solph/network.py
```diff
@@ -157,11 +174,13 @@ def __init__(self, **kwargs):
...
         scalars = ['nominal_value', 'summed_max', 'summed_min',
                    'investment', 'nonconvex', 'integer', 'fixed']
-        sequences = ['actual_value', 'variable_costs', 'min', 'max']
+        sequences = ['actual_value', 'variable_costs', 'min', 'max',
+                     'schedule', 'penalty_neg', 'penalty_pos']
         dictionaries = ['positive_gradient', 'negative_gradient']
         defaults = {'fixed': False, 'min': 0, 'max': 1, 'variable_costs': 0,
                     'positive_gradient': {'ub': None, 'costs': 0},
                     'negative_gradient': {'ub': None, 'costs': 0},
+                    'penalty_neg': 0, 'penalty_pos': 0, 'schedule': None
```
simnh: I was thinking if this could not rather be implemented like the gradient cost stuff as a dictionary to reduced number of attributes. It seems better to me, but again-> there are always multiple ways. Maybe ask @gnn for this

simnh: Sorry for taking so long to review this one... :-/

oemof/solph/network.py
c-koehl: I tested it and it worked. But I'm confused now, because you're actually right. Usually pyomo is not amused if None is provided. I will add a "return Constraint.Skip" for completeness

oemof/solph/network.py
c-koehl: I was also thinking about adding a default value. It's actually more user friendly. I think it depends on what kind of users are working with oemof. Do you think it will be necessary? I can easily add one :-)

oemof/solph/network.py
c-koehl: Hmmm I do not know about this one. I personally think it can be helpful, if the user can determine during what timesteps he will punish the system harder or being more "relaxed" if flow can not follow the schedule.

No worries, we are all busy and even holidays were in between :-)

oemof/solph/blocks.py
c-koehl: I like your idea. It is always better to be a little bit more specific and if we put "schedule" into the variable names, everybody knows where they belong to.

oemof/solph/network.py
c-koehl: Ah I think it is possible, because of BuildAction no return value has to be provided and the constraints will be added by add() function.
I remember pyomo did expect a return value (and throws an error if you provide None) if you build constraints without BuildAction. 
Maybe with BuildAction it does not expect anything and is fine with only executing the add() function. But I am honestly just guessing, because the documentary of pyomo's BuildAction is not very detailed.

c-koehl: Ok, I tried adding the return value Constraints.Skip, but as soon as it will be returned once, pyomo quits the constraint completely and continues with the next constraint. So in your case (schedule = [1, 2, None, 3]), the flow will only follow the schedule until the first None: [1, 2, None]. Everything following the first None will not be handled.
Interesting, Constraint.Skip behaves differently when building constraints with BuildAction compared to without.

uvchik:
@simnh  Is this ready to merge?

@c-koehl Do you have a good example. I think it is a good idea to publish an example on oemof-examples to make this feature more visible.

c-koehl:
> @c-koehl Do you have a good example. I think it is a good idea to publish an example on oemof-examples to make this feature more visible.

Yes I have a jupyter notebook example. I will create a pull request in oemof/oemof-examples.

src/oemof/solph/network.py
```diff
@@ -189,6 +211,15 @@ def __init__(self, **kwargs):
         if self.investment and self.nonconvex:
             raise ValueError("Investment flows cannot be combined with " +
                              "nonconvex flows!")
+        if (
+            len(self.schedule) != 0 and
+            ((len(self.schedule_cost_pos) == 0 and
+              not self.schedule_cost_pos[0]) or
+             (len(self.schedule_cost_neg) == 0 and
+              not self.schedule_cost_neg[0]))):
+            raise ValueError("The penalty and schedule attribute need "
```
uvchik, line 220: This error is not tested. Please add a test.

uvchik:
I fixed the conflicts caused by the restructuring of oemof.

Please check if everything still works as expected.

@simnh: This PR is related to this example: https://github.com/oemof/oemof-examples/pull/53

simnh:
I checked the example it works as expected. So I think this is fine to merge. The exmaple can be cleaned up a little (pep8 etc.) However, I will suggest this in the other PR.

uvchik:
I tried to understand this part:

```python
if (
        len(self.schedule) != 0 and
        ((len(self.schedule_cost_pos) == 0 and
        not self.schedule_cost_pos[0]) or
        (len(self.schedule_cost_neg) == 0 and
        not self.schedule_cost_neg[0]))):
    raise ValueError("The penalty and schedule attribute need "
                     "to be used in combination. \n Please set "
                     "the schedule attribute of the flow.")
```

My guess:
If `schedule` is set by the user, the user should also set `schedule_cost_pos` and `schedule_cost_neg`.

If  `len(self.schedule) != 0` is True which is necessary to raise the error the schedule is already set, but in the message it says: "Please set the schedule attribute of the flow.".

If the user have to set the `schedule_cost_neg/pos` parameters you should not set a default value, because a default value only makes sense for optional parameter.

c-koehl:
I do not know how to un - fail these tests. I am not sure if they were caused by me... Please correct me if I am wrong

PR #668, opened 2019-12-19
Features/add nonequidistant timeindex

c-koehl:
This PR gives the opportunity to have also a non-equidistant timeindex for the EnergySystem.

If a valid timeindex (no duplicate Datetimeindex in _timeindex_ ; _timeindex_ has to be of type pandas.DatetimeIndex) was found, _timeincrement_ will be calculated automatically by differentiating:

```python
timeindex = self.timeindex.to_series()
timeincrement = timeindex.diff().dropna()
timeincrement_sec = timeincrement.map(dt.timedelta.total_seconds)
timeincrement_hourly = list(timeincrement_sec.map(lambda x: x/3600))
timeincrement_hourly.append(1)
timeincrement = sequence(timeincrement_hourly)
```

If you have a non-equidistant timeindex during differentiation following question/problem occurs: 
How do I fill the last missing value? (see below for explanation)
I choose to fill it with 1. But other values would be valid, like you could fill it with the last known value, but thats a modelling question.

**Why does differentiation for _timeincrement_ returns a value less than in _timeindex_?**
During differentiation you compare the next value with the current value to get the change:
time(t+1) - time(t)
(and divide it by (t+1) - t)
That will always give an result except for the last given value: there is no next value to compare with.

**A non-equidistant timeindex should be handled with care**
  User has to keep the non-equidistant timeindex in mind: e.g. loss_rate(t) of GENERICSTORAGE cannot be scalar anymore. It needs to be adapted to the timeincrement!
  In documentation _loss_rate_ is already defined as 
> fraction of lost energy (e.g. leakage) per time

so I think it should be clear to adapt it in case of a non-equidistant timeindex. But such mistakes can still happen which leads to an incorrect solution of the optimization problem. 
1. Option:
  leave the issue to the user:
  extend documentation to point out possible critical points like in _loss_rate_ for GENERICSTORAGE
2. Option:
  extend code to assure correctness of optimization solution:
  critical points will be calculated automatically by timeincrement


uvchik:
1. I would not calculate the time increment within solph but rather change the API of the EnergySystem to make it possible to pass a Series of time increments instead of a time index. The index could be a numerical index instead of a time index.

2. We already use the timeincrement in [equations](https://github.com/oemof/oemof/blob/dev/oemof/solph/components.py#L380) but it seems to me that we [forgot it at some places ](https://github.com/oemof/oemof/blob/dev/oemof/solph/components.py#L378). In my opinion we should add the time increment to equations wherever it is needed.

I have not checked but critical parts could be gradients of Flows and the dsm-module. We have to check each component and all Flows.

c-koehl:
@uvchik 
1. I wouldn't get rid of the timeindex. It's still the best index especially for displaying and plotting the results. 

2. Yes that is true, in most places you correctly use the timeincrement for also non-equidistant timeincrements. I am gonna have a look tomorrow for more critical places like the _loss_rate_ attribute of GENERICSTORAGE

uvchik:
>     1. I wouldn't get rid of the timeindex. It's still the best index especially for displaying and plotting the results.

I do not want to get rid of it I just wonder if we should allow to use solph without one. It is useful for plotting but people may just want to use time steps and there is no reason to forbid it.

BTW We still can add a function to calculate time increments from time indexes in the oemof.tools package. The API could look like this.

```python
from oemof.tools import calculate_timeincrement

model(timeincrement=calculate_timeincrement(timeindex))

**Comment by c-koehl** [comment]
@uvchik I tried to change everything as you asked for. 
I think if we allow to use solph without a timeindex, an additional issue should be opened, because you are right, there is no reason to forbid it.  :-)

File: oemof/tools/helpers.py [hunk 1]
```diff
@@ -57,3 +59,24 @@ def flatten(d, parent_key='', sep='_'):
...
+        timeincrement = timeindex_sorted.diff().dropna()
+        timeincrement_sec = timeincrement.map(dt.timedelta.total_seconds)
+        timeincrement_hourly = list(timeincrement_sec.map(
+                                    lambda x: x/3600))
+        timeincrement_hourly.append(1.0)
+        timeincrement = sequence(timeincrement_hourly)
+        return timeincrement
+    else:
+        raise AttributeError(
+            "'timeindex' must be of type 'DatetimeIndex'.")
```
oemof-developer: A test is needed for this error.

uvchik:
Looks good to me. Just one small test is missing (sorry, I commented with wrong login).

@simnh Could you have a last brief look.

uvchik:
@c-koehl BTW:
The e-mail address associated with your local git is stored together with the commit. In your case this address is not valid (cons....@...s-MacBook-Pro.local), which might be intended. With this PR you would become an oemof contributor but as your git e-mail address is not associated with your account your commits will be counted as an anonymous contribution.  However, you can do what you want and you can stay as anonymous as you want. I just mentioned it in case you did it by mistake.

uvchik:
@c-koehl BTW:
The e-mail address associated with your local git is stored together with the commit. In your case this address is not valid (cons....@...s-MacBook-Pro.local), which might be intended. With this PR you would become an oemof contributor but as your git e-mail address is not associated with your account your commits will be counted as an anonymous contribution.  However, you can do what you want and you can stay as anonymous as you want. I just mentioned it in case you did it by mistake.

c-koehl:
> @c-koehl BTW:
> The e-mail address associated with your local git is stored together with the commit. In your case this address is not valid (cons....@...s-MacBook-Pro.local), which might be intended. With this PR you would become an oemof contributor but as your git e-mail address is not associated with your account your commits will be counted as an anonymous contribution. However, you can do what you want and you can stay as anonymous as you want. I just mentioned it in case you did it by mistake.

Oh sorry about that. It's my first time contributing. I updated my email adress on my profil and on my local git. I hope it should be fine now. 
I also added the missing test and the whatsnew file.

oemof/solph/components.py
```diff
@@ -358,7 +357,7 @@ def _storage_balance_first_rule(block, n):
             expr = 0
             expr += block.capacity[n, 0]
             expr += - block.init_cap[n] * (
-                1 - n.loss_rate[0])
+                (1 - n.loss_rate[0]) ** m.timeincrement[0])
```
p-snft: This changes the result w.r.t. the old version with constant timeincrement unless `m.timeincrement[t] == 1`. So, the default timeincrement (when not explicitly giving one) should be 1.

oemof/energy_system.py
```diff
@@ -36,7 +36,9 @@ class EnergySystem:
         Stored in the :attr:`entities` attribute.
         Defaults to `[]` if not supplied.
     timeindex : pandas.datetimeindex
-        Define the time range and increment for the energy system.
+        Define the time range for the energy system.
+    timeincrement : numeric (sequence)
```
p-snft: As this PR is filed against 0.3.3, the API should be stable. So, I assume it is still possible to not give a timeincrement. This should be documented better: What would be the default?

oemof/solph/components.py
c-koehl: That's the most difficult part. It depends on loss_rate. Is the given rate describing the loss per hour, 15 mins,...? My assumption is per hour, in documentation: "loss_rate: The relative loss of the storage capacity per timeunit." (you're right, I have to be more clear about the timeunit). In that case there is no need to set a default timeincrement. Please correct me if I'm wrong.

oemof/energy_system.py
c-koehl: If no timeincrement was given, it will be calculated regularly as in former versions. I don't think passing a default timeincrement is a good idea. Right now (according to my changes), there are two possibilities:
1. giving a timeincrement and timeindex: everything is fine, timeincrement will be used for further calculations.
2. giving a timeindex: timeindex has to be valid, it needs to have a 'freq' attribute -> constant timeincrement will be calculated as usual

uvchik: There is no default. If the `timeincrement` is `None` it will be calculated from the time index. This works only with an equidistant time index. The problem is, that the unit of the calculated `timeincrement` is hour even though we normally do not use units. It is possible to use Joule instead of kWh but in that case the `timeincrement` has to be in seconds and the loss rate should be the rate per second.

oemof/solph/components.py
uvchik: It is not possible to define a default `timeincrement`, because it belongs to the time index!

uvchik:
Thank you, @p-snft for bringing up this discussion!

Due to the generic approach we do not use units within solph. By now the user has to take care that all units are in line. But as I mentioned above, the time increment is calculated from the time index in hours, which is a bug in my opinion.

What can we do?

In my opinion we should move the automatic calculation from the model module to the tools module and make the user defining both the index and(!) the time increment. In that case we could allow every kind of index (if it works :smirk:) because the index is only needed for the results (if I remember correctly). This index could be numeric (1, 2, 3,...). It could be even numeric by default even though a datetime index is could to process the results. The real important information is the time increment.

In the tools module we could add a function to calculate the time increment for a equidistant time index (or integrate it in the existing function). The API would be slightly different from now:

```diff
+ from oemof.tools import calculate_timeincrement
+ EnergySystem(timeindex=timeindex,
+              timeincrement=calculate_timeincrement(timeindex, unit='hour'))
- EnergySystem(timeindex=timeindex)
````

In the documentation we could add a chapter about units and add some examples to explain it:

With a time increment of 30min you have to define a loss rate per minute and kilowattminutes [kWmin] or a time increment of 0.5h a loss rate per hour an kWh which might be more intuitive.

**Comment by uvchik** [comment]
We can discuss it here but it does not have to be all done within this PR and not only by @c-koehl.

File: oemof/solph/components.py (hunk 2 above)
Comment (p-snft) [reply oemof/solph/components.py]: Before, the loss_rate (sequence) might have already included a varying time increment: Assuming 10% losses per hour with time steps of `[1 h, 2 h, 1 h]`, I would have passed `[0.1, 0.19, 0.1]`. This allowed us to not define the units. The change per step is already defined.

As we assumed equidistant time steps, timeunit is just the width of the time steps, exactly what would be `timeincrement == 1` in your case.

**Comment by p-snft** [comment]
> We can discuss it here but it does not have to be all done within this PR and not only by @c-koehl, because it is an old bug and not her fault.

Indeed. It brought a problem to my attention which was already present before. However, since now the timeincrement is explicitly used, it becomes very relevant here. Especially, if it's not backward compatible (here: changing the results), it should be a feature for 0.4, not for the 0.3 branch.

**Comment by uvchik** [comment]
@c-koehl  We could move it from the fork to a branch to make it easier to help you.

@oemof/oemof-solph What do you think about my suggestions (see the list below)?

1. Move the calculation of the `timeincrement` to the tools module
2. Make the `timeincrement` mandatory
3. Allow a numeric index e.g. (time-step-index or hour-of-the-year-index)
4. Make the time index optional??? We could create a time-step-index by default but it is a little bit hacky because the length of the index is needed.

Did I forget anything?

I am fine with 1 to 3 but I do not like 4 any more. But I am fine with something like `timeindex=list(range(8760))` or `timeindex=list(range(1, 8761))`. I do not need a datetime index.

**Comment by p-snft** [comment]
Timeincrement plus timeindex is redundant. Actually, I would recommend to disallow giving redundant information, as it either allows to make mistakes or requires to double check whether the information is coherent.

**Comment by uvchik** [comment]
We noticed, that the internal calculation of the `timeincrement` causes problems. But if we remove the internal calculation we do need a second information. Therefore, we could reduce it to the following parameter: `number_of_timesteps` and `timeincrement`. 
If we change it this way we will lose the comfortable time index for the results. So I had the idea that it is still possible to pass an index of datetime and additionally the time increment. The datetime index would give us the information about the number of time steps. But you are right, a datetime index could have a different timeincrement than the `timeincrement` parameter if the user makes a mistake. 

1. We could try to catch that in a Warning. A warning is raised if the equidistant time increment is different from the calculated `timeincrement` in hours, minutes, or seconds.
2. We could leave the internal calculation but make it possible to choose the unit.
3. We could forbid datetime indexes and use only numeric indexes.
4. We could introduce a `ModelTime` class, with methods and attributes to cover all cases.
5. We could use `number_of_timesteps` and `timeincrement` and an optional `start_date` parameter, so that it is possible to create an datetime index in the background if wanted.

File: oemof/solph/components.py (hunk 2 above)
Comment (c-koehl) [reply oemof/solph/components.py]: If you're only considering the calculation above for a default timeincrement, then it might be possible to set a default `timeincrement == 1`, if and only if the user is really aware about calculating the loss_rate for each timestep. Wouldn't it be more user friendly, if the user doesn't have to calculate it all by hand?
Also, if you're considering all the other calculations, which include the timeincrement, it could lead to wrong calculations, e.g. for calculating the summed maximum of a flow:
```python
lhs = sum(m.flow[inp, out, ts] * m.timeincrement[ts]
                   for ts in m.TIMESTEPS)

rhs = (m.flows[inp, out].summed_max *
             m.flows[inp, out].nominal_value)
```

oemof/tools/helpers.py
```diff
@@ -57,3 +59,24 @@ def flatten(d, parent_key='', sep='_'):
...
+    if isinstance(timeindex, pd.DatetimeIndex):
+        if len(set(timeindex)) != len(timeindex):
+            raise IndexError("No equal DatetimeIndex allowed!")
+        timeindex = timeindex.to_series()
+        timeindex_sorted = timeindex.sort_values()
+        timeincrement = timeindex_sorted.diff().dropna()
+        timeincrement_sec = timeincrement.map(dt.timedelta.total_seconds)
+        timeincrement_hourly = list(timeincrement_sec.map(
+                                    lambda x: x/3600))
+        timeincrement_hourly.append(1.0)
```
c-koehl: I found a small bug. I would like to change this line to 
```python 
timeincrement_hourly.insert(0, 1.0)
```
the pandas.Series.diff() function results to a `timeincrement` with length-1 compared to the `timeindex`, because the first value will be missed out.
The output of
```python
timeidxH = pd.date_range('1/1/2017', periods=3, freq='H')
timeidx2H = pd.date_range("1/1/2017 04:00:00", periods = 2, freq="2H")
timeindex = timeidxH.append([timeidx2H])
print(timeindex)
print(timeindex.diff()
```
for `timeindex` is

2017-01-01 00:00:00   2017-01-01 00:00:00
2017-01-01 01:00:00   2017-01-01 01:00:00
2017-01-01 02:00:00   2017-01-01 02:00:00
2017-01-01 04:00:00   2017-01-01 04:00:00
2017-01-01 06:00:00   2017-01-01 06:00:00

and for `timeincrement` is 

2017-01-01 00:00:00        NaT
2017-01-01 01:00:00   01:00:00
2017-01-01 02:00:00   01:00:00
2017-01-01 04:00:00   02:00:00
2017-01-01 06:00:00   02:00:00

So it would make more sense of inserting the missing timeincrement value on the first place, instead on the last one.

I will change it with the following commit.

c-koehl:
With the last commit I changed the calculate_timeincrement function.

It won't be inserted 1.0 for the missing value, the user can now determine the value. If he doesn't give a value, the missing value will be backfilled (the next existing value will be inserted). I like the idea better, what do you think?

oemof/tools/helpers.py
p-snft: Sounds reasonable.

oemof/energy_system.py
p-snft: The PR is now against dev, so API changes are in order.

oemof/solph/components.py
p-snft: This is totally correct. We should just add a comment to the release notes, that the loss rate is now defined per hour and no more per time step.

doc/whatsnew/v0-3-3.rst
```diff
@@ -0,0 +1,52 @@
...
+
+Testing
+#######
+
+* something
+
+Other changes
+#############
+
+* something
```
p-snft: * The `loss_rate` of `GenericStorage`s is now defined per time increment.

doc/whatsnew/v0-3-3.rst
```diff
@@ -0,0 +1,52 @@
+v0.3.3 (??, ??)
```
p-snft: This will be part of v0.4.0.

oemof/solph/components.py
```diff
@@ -247,7 +246,7 @@ class GenericStorageBlock(SimpleBlock):
 
     Storage balance :attr:`om.Storage.balance[n, t]`
         .. math:: E(t) = &E(t-1) \cdot
-            (1 - \delta(t))) \\
+            (1 - \delta(t))^{\tau(t)/(1 h)} \\
```
p-snft: 1 hour is the default time unit, but not the only possible one. Maybe, introduce a symbol t_u for 1 time unit.

uvchik: At the moment the unit for the time increment is hour by default. But according to the discussion above we should make the API cleaner. So I agree that per time unit is better than per hour. It is possible to use seconds for the time increment to get Joul a energy unit instead of Wh.

uvchik:
@p-snft : Feel free to merge if you are happy, too.

c-koehl:
Thank you guys! But I just moved the improvements to the v0.4.0 whatsnew file, as you said. But it will be now part of version 0.3.3. Do you want me to create a new pull request to fix this?

p-snft:
It was merged into dev, so it will not be in v0.3.x as v0.3 has been branched a while ago.

uvchik:
Thank you @c-koehl :tada:

PR #392, opened 2017-12-05
Revise text for feedinlib

birgits:
@gnn or @uvchik please check :) Thanks!

PR #675, opened 2020-02-05
Update feedinlib information in oemof documentation

SabineHaas:
Addressed #392


uvchik:
Looks good even though I did not check if it compiles correctly. It would be nice to link to an openFred map and documentation. 

You could also change the text on the [oemof web page](https://oemof.org/libraries/#feedinlib). I think you can use basically the same text.

Thanks for your effort.

uvchik:
Sorry, @SabineHaas the doc moved to oemof/oemof: You have to copy it there.

uvchik:
This is not an oemof-solph issue.

Issue #648, opened 2019-11-11
Improve documentation of GenericCHP


ckaldemeyer:
As we got some questions concerning parameters such as Beta, etc. via the mailing list, we should improve the documentation in order to make the functionality more transparent:

* Add underlying paper of Mollenhauer et al. already in the general documentation and not only within the class documentation
* Outline calculation of relevant parameters such as Beta i.e. "Stromverlustkennziffer" already within the documentation by porting them from the paper e.g. from equation 3 on page 3 in this case

Maybe we can add a simple PR on the dev-meeting and fix it for the next version..

p-snft:
As this is targeted for 0.4, you should also consider to use more talking names (e.g. `power_loss_factor` instead of `beta`). For the formulas, you would still use the beta, so it can be easily identified with the corresponding one from the paper.

jnnr:
To add another observation: I just noticed that in the case of a backpressure turbine, the 'without district heating' specification in `Eta_el_max_woDH`, `Eta_el_min_woDH` is quite confusing.

uvchik:
We should also have a look at the symbols used in `ExtractionTurbineCHP`. We could use the same terms if they are the same in both classes. It doesn't matter if we change it in `GenericCHP` or in `ExtractionTurbineCHP` as long as they are the same.

jnnr:
Another observation: The order of equations, parameter table, figure and code snippets is changed comparing between the ExtractionTurbineCHP and GenericCHP.

I would find the following order natural and good to read:

1. figure
2. equations
3. parameter table
4. code snippets
5. (plots of results)

The figure gives an immediate impression, the equations tell the concept, next would be the parameter table if you want to look up what some parameter means, then the code snippets that show how to use the component.

uvchik:
Is this fixed by #677?

p-snft:
Seems not to be an issue anymore.

PR #677, opened 2020-02-10
Features/docs improve extraction turbine chp

jnnr:
I adapted the figure of the ExtractionTurbine component to make it more clear. Before, the blue background was not constrained by the nominal_value of the fuel input. I added an explanation of the iso-input lines and cleaned up the graphic appearance. I include it here so you can immediately have a look (Edit: Proportions are adapted to match typical operation ranges of extraction turbines):

ExtractionTurbine_range_of_operation

* Related issues?
#661 
#648  

Still open:

* [x] Re-read the text and check for typos and passages that can be described more precise.

* Share your knowledge: Inkscape is a great tool.

@nesnoj: I started this partly because of our discussion. If you like, have a look!

### DIFF
--- a/doc/_files/ExtractionTurbine_range_of_operation.svg
+++ b/doc/_files/ExtractionTurbine_range_of_operation.svg
@@ -6,11 +6,37 @@
    xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
    xmlns:svg="http://www.w3.org/2000/svg"
    xmlns="http://www.w3.org/2000/svg"
+   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
+   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
    id="svg8"
    version="1.1"
    viewBox="0 0 205.05406 99.122871"
    height="99.122871mm"
-   width="205.05406mm">
+   width="205.05406mm"
+   sodipodi:docname="ExtractionTurbine_range_of_operation.svg"
+   inkscape:version="0.92.3 (2405546, 2018-03-11)">
+  <sodipodi:namedview
+     pagecolor="#ffffff"
+     bordercolor="#666666"
+     borderopacity="1"
+     objecttolerance="10"
+     gridtolerance="10"
+     guidetolerance="10"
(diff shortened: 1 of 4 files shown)


nesnoj:
Hey, thanks for improving these docs, to my mind the usage is now more clear.

1. I added the backpressure coefficient `C_b` to equation (2) and the text in d5d6ce3d1fb0a8229ffbd45cbf239476094c7861, since it is one of the params found in literature and may help if you're not too familiar with CCGT. Feel free to accept or reject.

2. List of symbols: Maybe I do not fully unterstand the sense of the table. The component is parametrized by those vars and params but why is `beta` part of the symbols list? It is rather calculated using the conversion factors, you cannot pass it explicitly as in `GenericCHP`, right?
If so, I'd remove it from the table.
If not, I'd consider to add `C_b` to the table too.

3. On more thing: I recommend to place the `ExtractionTurbineCHP` docs right after those of `GenericCHP`..

jnnr:
>     1. I added the backpressure coefficient `C_b` to equation (2) and the text in [d5d6ce3](https://github.com/oemof/oemof/commit/d5d6ce3d1fb0a8229ffbd45cbf239476094c7861), since it is one of the params found in literature and may help if you're not too familiar with CCGT. Feel free to accept or reject.

>     2. List of symbols: Maybe I do not fully unterstand the sense of the table. The component is parametrized by those vars and params but why is `beta` part of the symbols list? It is rather calculated using the conversion factors, you cannot pass it explicitly as in `GenericCHP`, right?
>        If so, I'd remove it from the table.
>        If not, I'd consider to add `C_b` to the table too.

Beta is created in the background (named `main_flow_loss_index`) when the constraints are defined. Even if you cannot specify it directly, it is part of the component.

>     3. On more thing: I recommend to place the `ExtractionTurbineCHP` docs right after those of `GenericCHP`..

Yes, I thought of putting them together as well. It could be achieved by making the ordering consistent (basic, component, custom and then an alphabetical order in these categories).

        Sink (basic)
        Source (basic)
        Transformer (basic)
        ExtractionTurbineCHP (component)
        GenericCHP (component)
        GenericStorage (component)
        ElectricalLine (custom)
        GenericCAES (custom)
        Link (custom)
        SinkDSM (custom)

PR #679, opened 2020-02-24
Features/docs improve sequence data type

### LINKED ISSUE #673: Specify sequence data type in Flow's docstring
As discussed in https://github.com/oemof/oemof-thermal/issues/51 recently, in the params of [Flow()](https://github.com/oemof/oemof/blob/0a503aa492b6b8e76c2a22d1ab51987aa9a1ee95/oemof/solph/network.py#L49) it is not fully clear what "sequence" in `numeric (sequence or scalar)` refers to (e.g. in `actual_value`) - can you use a Pandas Series? Yes, but this is not covered by the docs (actually a DF is allowed too although it doesn't make much sense to provide a multi-dimensional type).
Reading the docstring I'd assume it expects a builtin type which are - according to the [docs](https://docs.python.org/3/library/stdtypes.html?highlight=sequence#sequence-types-list-tuple-range) - list, tuple and range. 

Apparently, all [iterable](https://docs.python.org/3/library/collections.abc.html#collections.abc.Iterable) objects are allowed here according to [this line in sequence()](https://github.com/oemof/oemof/blob/0a503aa492b6b8e76c2a22d1ab51987aa9a1ee95/oemof/solph/plumbing.py#L40) (except for string).

You may check with
```
>>> from collections.abc import Iterable
>>> from pandas import Series
>>> isinstance(Series([1, 2, 3]), Iterable)
True
```
The docstring should be something like
```
...
max : numeric (sequence or scalar)
      sequence must be iterable (instance of :class:`Iterable <collections.abc.Iterable>`) except :obj:`str`
...
```
(a little cumbersome though)

Do you agree?
If so, I could prepare a PR..

nesnoj:
The parameters' data type in the docstrings is changed from `numeric (sequence or scalar)` to `numeric (iterable or scalar)`, closes #673.

Beside the docstrings, I also adapted the param name of the function `sequence()` in 043a02b3e2c0045d2cdf9a1fb197e6ad87a56911. This is an API change, right? Should I include this too in the API changes list in what's new? (actually this param renaming is more a cosmetic matter, not actually needed) 

**EDIT:** The API change disallows this PR to be included in v0.3.3, is this correct? I assume v0.4.0 is the target then. Do I have to use `dev` or `0.4` (does not exist yet) as base?

### DIFF
--- a/doc/whatsnew/v0-4-0.rst
+++ b/doc/whatsnew/v0-4-0.rst
@@ -46,8 +46,12 @@ Other changes
 
 * The ``loss_rate`` of :class:`~oemof.solph.components.GenericStorage`
     is now defined per time increment.
+* The parameters' data type in the docstrings is changed from
+  `numeric (sequence or scalar)` to `numeric (iterable or scalar)`
+  (`Issue #673 <https://github.com/oemof/oemof/issues/673>`_).
 
 Contributors
 ############
 
 * Caterina Köhl
+* Jonathan Amme
--- a/oemof/network.py
+++ b/oemof/network.py
@@ -235,6 +235,8 @@ class Node:
 
 
 EdgeLabel = NT("EdgeLabel", ['input', 'output'])
+
+
 class Edge(Node):
     """ :class:`Bus`es/:class:`Component`s are always connected by an :class:`Edge`.
 
--- a/oemof/solph/components.py
+++ b/oemof/solph/components.py
@@ -58,7 +58,7 @@ class GenericStorage(network.Transformer):
 
     initial_storage_level : numeric
         The content of the storage in the first time step of optimization.
-    balanced : boolian
+    balanced : boolean
         Couple storage level of first and last time step.
         (Total inflow and total outflow are balanced.)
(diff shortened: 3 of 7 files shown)


uvchik:
I doubt that there will be v0.3.3 because there are already plans to publish v0.4.0 this spring.

I do not see any API changes but this PR is branched from dev, so it includes API changes from other PRs. Therefore, we have to merge into dev to avoid API changes in the `v0.3` branch.

nesnoj:
Thanks @uvchik for your feedback!
I thought any changes of params are API changes as all functions/objects such as `sequence()` (which is not private) can be used externally. But apparently I'm mistaken here..

Could you please approve this PR? (if not, I'll find someone else :)

uvchik:
Strictly spoken you are right. I just thought it is unlikely that somebody calls a function with one positional argument by naming the parameter explicitly but if the the function is used like this `sequence(sequence_or_scalar=5)` your are right. In this case it will break the API.

nesnoj:
Thank you! :+1:

Issue #547, opened 2018-12-10
nominal_value and actual_value


jnnr:
Flows have attributes `nominal_value` and `actual_value`. The latter, `actual_value`, gives information about the temporal profile of the flow. It is internally multiplied by `nominal_value`.

* Remark: A power timeseries does not necessary attain the maximal possible value or the nominal value in a given time interval. Scaling a timeseries so its maximum is not the same as scaling it to its general nominal value.
*  Terminology: Nennleistung is not necessarily the same as maximal power (https://www.energie-lexikon.info/leistung.html, section "Nennleistung"). The [docu](https://oemof.readthedocs.io/en/stable/oemof_solph.html#source-basic) does not draw this distinction . This means that a non-normalized power timeseries may contain values above the Nennleistung. E.g., PV can in some conditions produce more than its peak power which is measured under standard test conditions.

How to handle this?

My opinion: As flows are pretty flexible and can be used in many contexts, I suggest to avoid restricting its use just to meet the terminology of special cases. However, I would clearly state that `nominal_value` and `actual_value` are internally multiplied.

In this case, actual_value should be allowed to be >1. Without investment, this is already possible. In investment mode, it will lead to zero flow.

It seems to be caused by this constraint in blocks.py:

>
        def _max_investflow_rule(block, i, o, t):
            """Rule definition of constraint setting an upper bound of flow
            variable in investment case.
            """
            expr = (m.flow[i, o, t] <= (
                (m.flows[i, o].investment.existing + self.invest[i, o]) *
                 m.flows[i, o].max[t]))
            return expr
        self.max = Constraint(self.FLOWS, m.TIMESTEPS,
                              rule=_max_investflow_rule)

simnh:
Generally I agree, but I do not see why the constraint above should lead to 0:

if `max` in some t = 100, there is no problem for the flow to be > 0 and < inf+ depending on the value of invest (leaving the existing investment out for now).

jnnr:
I have to add that the case I am talking about is `fixed=True`. This leads to two constraints: `_investflow_fixed_rule` and `_max_investflow_rule`. 

Which means:
> 
flow = (invest + existing) * actual
flow <= (invest + existing) * max

where max seems to be set to 1 in the background. In the LP-file, this looks like this:

> 
c_e_InvestmentFlow_fixed(wind_electricity_17)_:
-1.089 InvestmentFlow_invest(wind_electricity)
+1 flow(wind_electricity_17)
= 0

[...]

PR #680, opened 2020-02-24
Revise API and behaviour of fixed flows

jnnr:
Implements the changes discussed in #547.

The steps I took according to the discussion in the issue:
* [x] Rename `actual_value` to `fix`,
* [x] Write a test that fails when a value in the `fix` sequence >1 in investment mode,
* [x] Fix it by not building max constraint for fixed flows,
* [x] I did not remove 'fixed', but it does not have to be passed explicitly as it is set to True in the 
background when fix is passed. Is that ok so? And shall I remove the explicit passing of 'fixed' from the tests?

Open TODOs in the docs:
* [ ] Replace `actual_value` in the docs
* [ ] Explain in the docs that `fix`, `min` and `max` are relative to `nominal_value`
* [ ] Explain in the docs that `fix` can be larger than 1.  

Can you have a look @oemof/oemof-solph?


p-snft:
I second @uvchik's comments but would be more strict: This is (also) meant to be a simplification. At the moment, setting `actual_value` has no effect if you do not also set `fixed` – this is unintuitive and may happen unnoticed. So, the name `fixed` is chosen so that you do not need redundant parameters.

Just closely related (but an easy fix within this PR): As `fix`, also `max` can be  >1. (I consider this very useful, please adjust the documentation an do not remove the feature.)

uvchik:
> I second @uvchik's comments but would be more strict: This is (also) meant to be a simplification. At the moment, setting `actual_value` has no effect if you do not also set `fixed` – this is unintuitive and may happen unnoticed. So, the name `fixed` is chosen so that you do not need redundant parameters.

As I understand it, we already decided to remove the `fixed` attribute and to rename the `actual_value` attribute to `fix`.

I agree that it should be possible to set `max` and/or `fix` > 0, but if `max` and `fix` is not set, `max` is set to 1 by default (as shown in the code example above). You are still free to set `max` to 100 if you want.

jnnr:
Thanks for your comments! I will continue working on this next week.

uvchik:
Okay, I think the code should work now. We should create an example that produces wrong results on the dev-branch but correct results (renaming `actual_value` to `fix`) on this branch. @jnnr Could you do that?

I removed the `fixed` attribute. The new `fix` attribute is `None` by default and if it is not None the variable of the Flow will be fixed to this value. It is not allowed to set `fix` and (`max` or `min`) at the same time, because this makes no sense.

I do not know why codecov fails, I do not think that I added uncovered lines and coveralls is okay.

We do need to check the documentation. I renamed `actual_value` to `fix` but there still might be some occurrences of `fixed` or `fixed=True`.

p-snft:
> We still have to adapt the documentation.

Thanks, I missed these few lost occurrences in my review. Should be removed now.

uvchik:
Thank you for fixing that right away.

PR #681, opened 2020-02-25
Features/rename storage capacity

jnnr:
Until now, the current content of the storage at time t is called `capacity`. This is confusing, as the maximum storable amount is called `nominal_storage_capacity`, which is similar. In general, one might think that capacity means the max. amount of energy (water, stuff, etc.) a storage can hold. So `nominal_storage_capacity` seems to be an adequate name for its purpose, but `capacity` not. 

The user does not define this as an input parameter, but rather sets `initial_storage_level`, which is relative to `nominal_storage_capacity`. However, it is an important part of the results. As a developer, you might look into GenericStorageBlock or GenericStorageInvestmentBlock and be confused by the name.

The current filling level is sometimes called 'state of charge' or SOC. But this quantity is relative to the nominal storage capacity. We were discussing about a suitable name in #538.

Therefore, in this PR I renamed capacity -> storage_content, as we decided in #538.

Thanks again for the discussion. Can you please have a look if I did not omit something? I hope this step helps to make things clearer and avoid confusions in the future. 

* [x] Review the docs
* [ ] Explain storage content in the docs (tell that storage content is an absolute value, other than e.g. initial_storage_level and max_storage_level)
* [x] Write changes to what's new


oemof/solph/components.py
```diff
@@ -477,17 +479,17 @@ class GenericInvestmentStorageBlock(SimpleBlock):
           (InvestmentFlow.invest(n, target(n)) + existing) *
           invest\_relation\_input_output(n) \\
           \forall n \in \textrm{INVEST\_REL\_IN\_OUT}
-
-    Maximal capacity :attr:`om.InvestmentStorage.max_capacity[n, t]`
-        .. math:: E(n, t) \leq invest(n) \cdot c_{min}(n, t), \\
-            \forall n \in \textrm{MAX\_INVESTSTORAGES,} \\
+    # TODO:
```
uvchik: What does that mean?

jnnr: Good question. That was there already. I will check the blame.

jnnr: Ah, no. That was me :) I guess that I should check that again.

jnnr: I compared the docstring with the constraint defined below. Seems consistent. I removed the TODOs.

jnnr:
> There might be a quick fix that could be done with this PR, however, you did not introduce the problem.

I thought that the problem was introduced sufficiently in the issue. I added some more explanation, hope that helps.

doc/oemof_solph.rst
```diff
@@ -508,7 +508,7 @@ GenericStorage (component)
 In contrast to the three classes above the storage class is a pure solph class and is not inherited from the oemof-network module.
 The ``nominal_storage_capacity`` of the storage signifies the storage capacity. You can either set it to the net capacity or to the gross capacity and limit it using the min/max attribute.
 To limit the input and output flows, you can define the ``nominal_storage_capacity`` in the Flow objects.
```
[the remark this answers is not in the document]
jnnr: You are right. I corrected the name.

joroeder:
@jnnr Could you have a look at this PR again? ☺️

oemof/solph/components.py
```diff
@@ -51,13 +51,14 @@ class GenericStorage(network.Transformer):
...
-        The content of the storage in the first time step of optimization.
+        The relative storage content in the timestep before the first
+        time step of optimization (between 0 and 1).
     balanced : boolean
         Couple storage level of first and last time step.
         (Total inflow and total outflow are balanced.)
     loss_rate : numeric (iterable or scalar)
-        The relative loss of the storage content per time unit.
-    fixed_losses_relative : numeric (iterable or scalar)
+        The relative loss of the storage content per timeunit.
```
jnnr:
```suggestion
        The relative loss of the storage content per time unit.
```

jnnr: This should be two words.

p-snft: This is a regression: 'time unit' was correct.

uvchik:
I hope I fixed everything in the right way. Could you have a brief look at your changes.

uvchik:
I tried to exclude the conf.py file from the pep8-Checker but I did not succeed. If anybody has an idea, please go ahead.

joroeder:
> 
> 
> I tried to exclude the conf.py file from the pep8-Checker but I did not succeed. If anybody has an idea, please go ahead.

Sorry, I first need to get familiar to all these checker. I think, we should not worry to much for now in every branch. We might need to have a look at all the checker issues in dev anyway, right?

docs/usage.rst
```diff
@@ -561,7 +561,26 @@ The following code block shows an example of the storage parametrization for the
...
 
-For more information see the definition of the  :py:class:`~oemof.solph.components.GenericStorage` class or check the `example repository <https://github.com/oemof/oemof-examples>`_.
+If you want to view the temporal course of the state of charge of your storage
+after the optimisation, you need to check the ``storage_content`` in the results:
+
+.. code-block:: python
+
+    results = oemof.outputlib.processing.results(om)
+    column_name = (('your_storage_label', 'None'), 'storage_content')
+    SC = views.node(results, 'your_storage_label')['sequences'][column_name]
```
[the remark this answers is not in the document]
uvchik: It works for me like this.

```python
from oemof.solph import views
views.node(..)
```

src/oemof/solph/components.py
```diff
@@ -685,52 +756,82 @@ def _create(self, group=None):
...
+                for n in group
+                if n.invest_relation_output_capacity is not None
+            ]
+        )
+
+        self.INVEST_REL_IN_OUT = Set(
+            initialize=[
+                n for n in group if n.invest_relation_input_output is not None
+            ]
+        )
```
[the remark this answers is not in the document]
uvchik: It is a typical style but not the only one. It might be a good idea to use Black style because some packages of oemof already do.

uvchik:
We should worry if it is caused by this PR but `conf.py` and the missing docstring are issues we will not fix anyway. You are right, we can fix it somewhere else.

docs/usage.rst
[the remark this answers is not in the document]
joroeder: yes, it is working like that. I will adapt the example accordingly

src/oemof/solph/components.py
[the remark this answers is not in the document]
joroeder: Okay, I don't mind ;)

docs/usage.rst
```diff
@@ -561,7 +561,26 @@ The following code block shows an example of the storage parametrization for the
         initial_storage_level=0.5, balanced=True,
         inflow_conversion_factor=0.98, outflow_conversion_factor=0.8)
 
-For more information see the definition of the  :py:class:`~oemof.solph.components.GenericStorage` class or check the `example repository <https://github.com/oemof/oemof-examples>`_.
+If you want to view the temporal course of the state of charge of your storage
+after the optimisation, you need to check the ``storage_content`` in the results:
+
+.. code-block:: python
+
+    results = oemof.outputlib.processing.results(om)
```
joroeder, line 569: So, should be put it like that?

```suggestion
    from oemof.solph import processing, views

    results = processing.results(om)
```

PR #683, opened 2020-02-27
Make v0.3 development available in actual dev

uvchik:
As the dev branch is heading to be the new v0.4. release we should integrate all festures from the v0.3 branch.


uvchik:
I think it is easy to fix the merge conflicts in `whats_new.rst` and `__init__.py` but there are contradictory changes in the `components.py` module.

Basically we do have the following changes and I added my idea to merge it. I had just a brief look so I am not sure if I got it right. @jnnr Could you merge dev into v0.3-dev to solve the conflicts?

v03:
`(1 - beta(t)) - gamma(t) * E_nom - delta(t)`

dev:
` (1 - delta(t)) ** (tau(t))`

my merge approach
`((1 - beta(t)) - gamma(t) * E_nom - delta(t)) ** (tau(t))`

**The full code of the equations** 

## v03

grafik

## dev

grafik

## my merge approach

grafik

uvchik:
I think it is easy to fix the merge conflicts in `whats_new.rst` and `__init__.py` but there are contradictory changes in the `components.py` module.

Basically we do have the following changes and I added my idea to merge it. I had just a brief look so I am not sure if I got it right. @jnnr Could you merge dev into v0.3-dev to solve the conflicts?

v03:
`(1 - beta(t)) - gamma(t) * E_nom - delta(t)`

dev:
` (1 - delta(t)) ** (tau(t))`

my merge approach
`((1 - beta(t)) - gamma(t) * E_nom - delta(t)) ** (tau(t))`

**The full code of the equations** 

## v03

grafik

## dev

grafik

## my merge approach

grafik

p-snft:
@uvchik: I guess the hardest thing is to understand that beta in v03 is called delta in dev. Your equations look reasonable.

jnnr:
I guess it would be rather

E(t) = E(t-1) * (1 - beta(t)) ** tau(t) - (gamma(t) * E_nom + delta(t))) * tau(t) + ...

The loss factor, beta, is taken to the power of tau, but the relative (gamma) and absolute (delta) constant losses are multiplied by tau.

jnnr:
Looks good now (to me). All tests pass. However, it seems that there is no test for the non-equidistant timeindex. At least, I did not find one in the PR #668.

oemof/solph/components.py
```diff
@@ -199,6 +207,13 @@ def _check_invest_attributes(self):
             e2 = ("Overdetermined. Three investment object will be coupled"
                   "with three constraints. Set one invest relation to 'None'.")
             raise AttributeError(e2)
+        if (self.investment and
+                self.fixed_losses_absolute != solph_sequence(0) and
+                self.investment.existing == 0 and
+                self.investment.minimum == 0):
+            e3 = ("With fixed_losses_absolute > 0, either investment.existing "
```
uvchik: This error is not tested, but I do not know who forgot it. We should add the test.

doc/whats_new.rst
```diff
@@ -8,7 +8,7 @@ These are new features and improvements of note in each release
     :local:
     :backlinks: top
 
-.. include::  whatsnew/v0-4-0.rst
+.. include::  whatsnew/v0-3-3.rst
```
p-snft: In dev, we should have both whatsnews.

jnnr: As I understood, we will skip v0.3.3, won't we?

jnnr: In that case, I would just rename the file v0.3.3 -> v0.4.0

jnnr:
Thanks for the test, @uvchik! And it actually caused you to fix something :) Shall we open an issue for the missing test for nonequidistant timestep and merge this here?

uvchik:
Yeah, I just fixed the pep8 issues (everything I could fix). I think it is not possible to break the tables. It does not bother me but it makes it difficult to find the real conflicts because you get all the messages from the docstring table.

I think it is good to merge it fast to avoid more conflicts :smile: 

Thank you @p-snft and @jnnr

uvchik:
> Thanks for the test, @uvchik! And it actually caused you to fix something :) Shall we open an issue for the missing test for nonequidistant timestep and merge this here?

YES :exclamation:

jnnr:
> I think it is not possible to break the tables. It does not bother me but it makes it difficult to find the real conflicts because you get all the messages from the docstring table.

That's true. Maybe there is a way to tell pep8speaks to ignore the tables?

PR #692, opened 2020-03-23
Add confirmed mac installation instructions

ChillkroeteTTS:
I just installed oemof and the cbc solver on my mac and it worked out of the box. Thus I want to remove the remark, that it is not tested yet. On the way I tried to conform the installation instructions for mac to the scheme which is used for Windows and Linux.


uvchik:
Thank you, @ChillkroeteTTS .

It is good to hear, that everything went fine :smile:

PR #697, opened 2020-03-31
Install cbc solver on appveyor to make test possible

uvchik:
* The appveyor test cannot pass because by now it is not possible to use cbc.

* Somehow it is possible to use the windows power shell to install the cbc solver. See the appveyor docs for more information.


chrklemm:
Hello everyone, I saw in the user-meeting protocol the following question with the note to this post: 

"Who can help to install the cbc-solver on a windows system automatically to make windows tests possible?"

I am not sure if this helps, but with the following cmd command you can download a file. It will be stored in the directory, where you execute the command:

curl url --output cbc.exe

Since the CBC-solver does not have to be installed under windows - it is sufficient if it the "cbc.exe" is located in the same folder as the Python document that accesses cbc - would it be possible to execute the following command in the appropriate folder?

curl https://www.coin-or.org/download/binary/CoinAll/COIN-OR-1.7.4-win32-msvc10.exe --output cbc.exe

.appveyor.yml
```diff
@@ -19,7 +55,7 @@ install:
   - '%PYTHON_HOME%\Scripts\easy_install --version'
   - '%PYTHON_HOME%\Scripts\pip --version'
   - '%PYTHON_HOME%\Scripts\tox --version'
-  - ps: (new-object net.webclient).DownloadFile('https://osf.io/rmfb7/download', 'my_test_solver.exe')
+  - ps: (new-object net.webclient).DownloadFile('https://osf.io/xyqre/download', 'cbc.exe')
```
uvchik, line 58: This is the interesting line. It already downloads the solver. The problem is that we need to copy it to PATH-dir or to add the present folder to the PATH. We do not want to execute the solver but to make it available. It will be executet by the python script.

chrklemm: What about this python command?

import os
os.environ["PATH"] += os.pathsep + 'C:\\Program Files (x86)\\Graphviz2.38\\bin'

uvchik: https://www.appveyor.com/docs/build-configuration/#installing-additional-software

uvchik: > What about this python command?

Where do you want to put it? I think we need the same command for the power-shell (ps) as shown above.

uvchik:
Idea: Use github actions instead of appveyor.

uvchik:
Once PR #853 is merged the failing tests will be fixed.

PR #700, opened 2020-04-09
Remove tests from the oemof-network package

uvchik:
As oemof-network is now a separate package the test should be there and not here.


uvchik:
I just moved tests to oemof.network. Coverage should stay the same because solph code is not touched.

Issue #684, opened 2020-03-04
Missing test for non-equidistant timeindex


jnnr:
The new feature (added in #668) is not tested anywhere yet. Please add one ore more tests.

@c-koehl, @oemof/oemof-solph

c-koehl:
What kind of tests are you thinking of? You can find my tests [here](https://github.com/c-koehl/oemof/blob/ac76b7346e2f78cde658bce0bc1ef68b3d1683f2/tests/test_models.py#L47).

jnnr:
Hey there, sorry for the late reply. You are right, there are already some tests, and they are also merged together with PR #668. It seems like I have overlooked them. Sorry, my bad! 

Looking at the tests i noticed that despite them being in test_models, they only test the function calculate_timeincrement of oemof.tools.helpers. Another test that actually tests how a model with a nonequidistant timeindex is build (and comparing lp files) could be added? I have no complete overview over all kinds of tests we have in oemof, so someone else might have a better idea :)

c-koehl:
I created a pr for this issue (see above). Is it sufficient?

uvchik:
The PR is move to the next milestone, so I will move this issue, too.

p-snft:
There is one now.

PR #701, opened 2020-04-10
Test for Features/add nonequidistant timeindex

c-koehl:
This pr relates to issue [#684](https://github.com/oemof/oemof-solph/issues/684).

A constraint test with comparing a lp file was added to the new feature of a nonequidistant timeindex.
I hope the tests validates the functionality. (e.g. the costs in the objective expression)

Additionally an unecessary import was deleted which can lead to circular imports.


uvchik:
I think you did not use the most actual state of the dev branch. Please check if your remote dev branch is in line with the upstream dev branch. A merge might be difficult, but as you mostly just added code (not changed code) you can just save your code somewhere, update your dev and open a new PR. I think this would be the fastest way.

joroeder:
> 
> 
> Everything seems fine. Nevertheless I had some issues adapting to the new structure of oemof. I could not run the nosetest command `nosetests -w "/path/to/oemof" --with-doctest`, because I received multiple import errors, e.g. in file solph/custom.py:
> 
> ```
> from oemof.network.network import Transformer as NetworkTransformer
> ModuleNotFoundError: No module named 'oemof.network'
> ```

``tox`` is the new test tool ([https://oemof-solph.readthedocs.io/en/v0.4.0/contributing.html#tests](https://oemof-solph.readthedocs.io/en/v0.4.0/contributing.html#tests)) you can run for example ```tox -e "clean, check, docs, py37"``` to specifiy the test run, e.g. just test with python 3.7 , or just the docs or the tests ...

uvchik:
Shouldn't we add a component to the test where the constraints that differs with a non-equidistant time index (e.g. GenericStorage).

c-koehl:
I deleted the import and use of sequence in helpers.py because it was leading to circular imports (in older versions, I don't know about the newer, flat ones) and I don't think it's necessary

uvchik:
> I deleted the import and use of sequence in helpers.py because it was leading to circular imports (in older versions, I don't know about the newer, flat ones) and I don't think it's necessary

I agree, in your case it is always a list.

Could you please solve the merge conflicts.

p-snft (approved):
Looks very good and also answers questions I had about this feature.

uvchik:
Thank you both for the test and the review. 

The most important thing is that you check the results for a storage in an energy system with a nonequidistant timeindex - especially the time sensitive results such as the loss rate. For example an example that leads in a  table that shows that the losses for a three hour time interval is three times higher than for a one hour interval.

After that we can merge it right away. The code base is fine.

c-koehl:
It is not 3 times higher.
E.g. if you have a 0.1 loss rate per hour, after one hour the storage would have a capacity of 1 - 0.1 = 0.9 times it former capacity.
The capacity after 3h would be (1 - 0.1)  ^ 3 =  0.729 times it former capacity.
It is like the compound interest (zinseszins) formula

uvchik:
> It is not 3 times higher.
> E.g. if you have a 0.1 loss rate per hour, after one hour the storage would have a capacity of 1 - 0.1 = 0.9 times its former capacity.
> The capacity after 3h would be (1 - 0.1) ^ 3 = 0.729 times its former capacity.
> It is like the compound interest (zinseszins) formula

Yes, I know. Sorry, I was not exact in my example. I just wanted to say that this has to be checked. If it works correctly for all these time sensible results I am fine with it.

p-snft:
Continued in #851.

PR #704, opened 2020-05-13
Disable GH PR and issue link checks outside of Travis

gnn:
Our changelog contains so many links to GitHub issues and pull requests that checking them all locally causes GitHub do respond with a 429 error because of too many requests, but the requests from Travis seem to be whitelisted. So this PR is about finding a way to disable requests to the most frequented URLs when testing locally, but keeping everything switched on when building on Travis.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -47,4 +47,11 @@ napoleon_use_ivar = True
 napoleon_use_rtype = False
 napoleon_use_param = False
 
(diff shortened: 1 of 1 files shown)


gnn:
Hey @uvchik, it seems to work. Looking at the Travis build log shows that all the the links are checked, while they are skipped while building locally. Please try it out so that I know that I didn't make any mistakes. :)

PR #705, opened 2020-05-25
Docs/improve documentation

uvchik:
As solph is now a separate package the documentation has to be revised.

* [x] Remove old stuff, that does not belong to oemof-solph but other packages
* [x] Move content of outputlib to oemof-solph
* [x] Add authors to header because due to the restructuring of oemof the information about the contributors got lost.


joroeder:
Some notes to the current state:
* the section Overview > Installation > Installing python3 should completely be moved to the oemof/oemof docs.
* the section "Contributing" should also go to the oemof/oemof docs. The sectin Overview > Contributing could be left, but then the link must be replaced by the link to the future Contribution section in oemof/oemof
* Should the section "Authors" be kept? Actually, I like it. Then people see that there are many people involved, and everybody can be part of it.

PR #710, opened 2020-06-23
Fix SolverStatus and '_bound' problem of pyomo 5.7.0

### LINKED ISSUE #709: 'SolverStatus' object has no attribute 'key'
**Describe the bug**
When solving an energy system in a newly installed virtual environment, an unhandled exception

> 'SolverStatus' object has no attribute 'key'

is thrown. The reason probably is an (undocumented) change in pyomo.

**To Reproduce**
1. Install current pygmo
2. Install oemof.solph
3. Try to solve some energy system 
4. See error also occurs in GitLab's CI pipelines, which is where I first found it.

p-snft:
There seems to be an (undocumented) change to Pyomo's API. (They never provide a changelog, so one might argue, every change is undocumented.) This commit allows to live with the issue.

Testing would require using different versions on Pyomo, and it would require testing to find out which version introduced the change.

Fix #709

### DIFF
--- a/docs/whatsnew/v0-4-1.rst
+++ b/docs/whatsnew/v0-4-1.rst
@@ -26,12 +26,15 @@ Documentation
 Bug fixes
 ^^^^^^^^^^^^^^^^^^^^
 
-* something
+* Fixed incompatibility with recent Pyomo release (5.7)
 
 Known issues
 ^^^^^^^^^^^^^^^^^^^^
 
-* something
+* Results of one-time-step optimisation counterintuitive
+    If an optimisation with one time-step is performed, at the processing of the
+    results, the scalars of the results is stored in the dict of the sequences.
+    (See `Issue #693 <https://github.com/oemof/oemof-solph/issues/693>`_)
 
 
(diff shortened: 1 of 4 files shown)


uvchik:
I would put the future version in the `try`-part and the old version in the `except`-part. People will use the new version in the future and then the old version will be the exception.

I do not have a idea how to simulate an older pyomo version in the tests, so these lines will remain untested I guess.

@gplssm In the `SinkDSMDelayBlock` class a protected attribute is accessed. You can do it but it is always dangerous. You do not break the API if you change a protected attribute and this is what happend [here](https://github.com/oemof/oemof-solph/blob/dev/src/oemof/solph/custom.py#L1320) with the new pyomo version 5.7. Could you try to fix it?

uvchik:
I got an idea how to fix the `SinkDSMDelayBlock`. @gplssm Can you check the fix!

p-snft:
Should we require pyomo 5.7 in the setup.py and use the new API instead of the try/catch block?

PR #711, opened 2020-06-23
Pin Pyomo below 5.7

p-snft:
This is an alternative approach to #710, which solves #709 but seems to have other issues. (Probably more Pyomo changes in new point release, and bad testing options.)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -87,6 +87,7 @@ setup(
         "numpy",
         "pandas",
         "pyomo >= 4.4.0, < 5.7",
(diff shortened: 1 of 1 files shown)


uvchik:
I have no idea what happened here but I get the same error when I update to pyomo 5.7.0 and downgrade again to 5.6.9.

uvchik:
It works but #710 is more promising. I do not understand why Scrutinizer is failing.

p-snft:
I do also prefer #710. Thus I withdraw this PR.

Issue #715, opened 2020-07-07
GenericStorage only supports one input/output


psychon:
Hi,

**Describe the bug**
solph non-deterministically picks one input and one output of a GenericStorage and ignores the others.

**To Reproduce**
I'm running the following program with oemof 0.3.2:
```python
import oemof.solph as solph
import pandas as pd

### Don't ask
import pyutilib.services
import os
os.environ['PATH'] += os.pathsep + 'C:\\Users\\UlSchl\\Desktop\\cbc\\'
pyutilib.services.register_executable(name="cbc")
### End of "don't ask"

es = solph.EnergySystem(timeindex=pd.date_range('1/1/2017', periods=2, freq='min'))

input1 = solph.Bus(label="input1")
input2 = solph.Bus(label="input2")
output = solph.Bus(label="output")

es.add(input1, input2, output)

# In time step #1: source1 produces one energy, sink consumes one energy
# In time step #2: source2 produces one energy, sink consumes one energy

es.add(solph.Source(
    label='source1',
    outputs={input1: solph.Flow(
        actual_value=[1, 0],
        nominal_value=1,
        fixed=True,
    )}
))

es.add(solph.Source(
    label='source2',
    outputs={input2: solph.Flow(
        actual_value=[0, 1],
        nominal_value=1,
        fixed=True,
    )}
))

es.add(solph.Source(
    label='sink',
    inputs={output: solph.Flow(
        actual_value=[1, 1],
        nominal_value=1,
        fixed=True,
    )}
))

# Everything is connected to a generic storage
es.add(solph.components.GenericStorage(
    label='storage',
    nominal_storage_capacity=1337,
    initial_storage_level=42,
    inputs={
        input1: solph.Flow(),
        input2: solph.Flow(),
    },
    outputs={output: solph.Flow()},
))

# Time to solve this
model = solph.Model(es)
model.write("model2.lp", io_options={"symbolic_solver_labels": True})
model.solve()
```
Output is something like:
```
WARNING: Constant objective detected, replacing with a placeholder to prevent
    solver failure.
WARNING: Constant objective detected, replacing with a placeholder to prevent
    solver failure.
[redacted]\venv\lib\site-packages\oemof\solph\models.py:216: UserWarning: Optimization ended with status ok and termination condition infeasible
  UserWarning)
```
Infeasible? That sounds wrong. The system just has to pass the energy from source to sink. Let's take a look at the lp file:
```
[stuff]
[...]

PR #716, opened 2020-07-08
Docs/fix generic storage docs

uvchik:
* Make it clear that the GenericStorage is designed for only one input and one output flow.

* This PR is part of issue #715

### DIFF
--- a/src/oemof/solph/components.py
+++ b/src/oemof/solph/components.py
@@ -33,7 +33,7 @@ from pyomo.environ import Set
 from pyomo.environ import Var
 
 
-class GenericStorage(network.Transformer):
+class GenericStorage(network.Node):
     r"""
     Component `GenericStorage` to model with basic characteristics of storages.
 
--- a/src/oemof/solph/components.py
+++ b/src/oemof/solph/components.py
@@ -29,7 +29,7 @@ from pyomo.environ import Set
 from pyomo.environ import Var
 
 from oemof.network import network
-from oemof.solph.network import Transformer as solph_Transformer
+from oemof.solph import network as solph_network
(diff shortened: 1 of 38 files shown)


uvchik:
A new `isort` issue occurred, so I fixed it within this PR.

src/oemof/solph/components.py
```diff
@@ -246,6 +249,13 @@ def _check_invest_attributes(self):
 
         self._set_flows()
 
+    def _check_number_of_flows(self):
+        msg = "Only one {0} flow allowed in the GenericStorage {1}."
+        if len(self.inputs) > 1:
+            raise AttributeError(msg.format("input", self.label))
+        if len(self.outputs) > 1:
```
psychon, line 259: Should this be `== 1` instead of `> 1` (in both checks)? I haven't checked, but I would expect that without inputs or outputs, the code does the equivalent of `list()[0]`, causing an exception to be thrown.

uvchik: You cannot do it because it is allowed to initialise a GenericStorage without inputs and outputs even though it is unwanted in most situation. For any reason users may want to add these attributes later.

I added a `SuspiciousUsageWarning` which users can switch off if they know what they are doing. This warning is raised if no inputs or outputs are defined (including misspelling).

uvchik:
I know this is not how it should be, but I added a missing test that does not belong to this issue. But I did not want to open an extra PR for one additional test. Sorry :blush:

uvchik:
@psychon I just wanted to bring everything in line but we can still talk about future changes.

uvchik:
If there are no objections I will merge this PR next week.

PR #717, opened 2020-08-17
Fix/new pandas version changed indexing

uvchik:
  * The new pandas version does not allow tuples as index but we use tuples as index

The following test method from `class TestParameterResult` (test_processing.py) fails but worked with old python versions.

```python
def test_parameter_with_node_view(self):
        param_results = processing.parameter_as_dict(
            self.es, exclude_none=True)
        bel1 = views.node(param_results, 'b_el1')
        assert (
            bel1['scalars'][(('b_el1', 'storage'), 'variable_costs')].values
            == 3
        )
  ```

The problem is that `bel1['scalars'].index[13]` returns:
(('b_el1', 'storage'), 'variable_costs')

But the following does not work anymore:
```python
key = bel1['scalars'].index[13]
bel1['scalars'][key]
```
The reason is that key is a tuple you have to use the following code:
```python
key = bel1['scalars'].index[13]
bel1['scalars'][[key]]
```

But this is not very intuitive.

In the test above you have to make the following changes:

```diff
- bel1['scalars'][(('b_el1', 'storage'), 'variable_costs')].values
+ bel1['scalars'][[(('b_el1', 'storage'), 'variable_costs')]].values 
```

I fixed the test but we should think about the concept or we should add that to the documentation.

### DIFF
--- a/tests/test_processing.py
+++ b/tests/test_processing.py
@@ -242,10 +242,13 @@ class TestParameterResult:
         param_results = processing.parameter_as_dict(
             self.es, exclude_none=True)
         bel1 = views.node(param_results, 'b_el1')
-        eq_(bel1['scalars'][(('b_el1', 'storage'), 'variable_costs')], 3)
+        assert (
+            bel1['scalars'][[(('b_el1', 'storage'), 'variable_costs')]].values,
+            3
+        )
 
         bel1_m = views.node(param_results, 'b_el1', multiindex=True)
-        eq_(bel1_m['scalars'].loc[('b_el1', 'storage', 'variable_costs')], 3)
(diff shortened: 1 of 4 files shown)


uvchik:
We may want to merge these fixes quickly, to help developers to avoid additional fails.

If we do so we should open another Issue to keep the general issue in mind.

simnh:
This pandas stuff recently also broke one of my models..(not inside oemof). I think for now the best solution is to go with the way pandas wants us to do it. For all users that are somewhat into pandas, the indexing will make sense, for others not so much, but they will get used to it :-D

simnh:
For the long term, I think we might discuss another API for results

uvchik:
Okay, I started an extra discussion (#718) about this topic, so that I can merge this PR now.

PR #720, opened 2020-09-01
Feature/unidirectional link

p-snft:
For quite some time, there has been a "Link" component in "custom" that is meant for lossy transport of electricity between two buses (typically locations). This component is now enhanced to allow exclusive Flow in either one or the other direction.

* [ ] Documentation
* [ ] Tests


p-snft:
@simnh: I would like to merge this and start raising quality to a mainline component later. What do you think?

PR #609, opened 2019-06-13
update docstrings of some classes in components.py

oakca:
Related issues #597 

- sets are removed from `GenericStorage` and `GenericInvestmentStorage`
- table is added to `GenericStorage` and  `OffsetTransformer` 
- some small changes for the length of the "===" in tables

PR #721, opened 2020-09-02
Follow up of PR 609

uvchik:
#609


uvchik:
I think there is no additional use, so I will not resolve the conflicts.

PR #722, opened 2020-09-02
GasLine

ghost:
Piecewise linear GasLine model


joroeder:
Hi Philipp, thank you for opening this PR! Looks very interesting! However, it would be really nice, if you provide further descriptions of your component and some more details, like more extensively the general idea, what are the required parameters, what should the component can do, and the mathematical background the component should be based on. Schemes are very welcome as well  :wink:

philipp235:
> Hi Philipp, thank you for opening this PR! Looks very interesting! However, it would be really nice, if you provide further descriptions of your component and some more details, like more extensively the general idea, what are the required parameters, what should the component can do, and the mathematical background the component should be based on. Schemes are very welcome as well 😉

philipp235:
The goal of the GasLine object is to model a non-linear gas flow between two nodes. For this purpose a potential in form of a pressure is assigned to each node and the gas flow in the pipeline is the result of a pressure difference between the nodes

joroeder:
Hi Philipp, are you still working on that? If you have additional attributes eg like pressure at each Bus, how do you plan to connect the bus to the rest of an oemof-solph energy system?

philipp235:
> 
> Hi Philipp, are you still working on that? If you have additional attributes eg like pressure at each Bus, how do you plan to connect the bus to the rest of an oemof-solph energy system?

Hi Joroeder it is possible to use the GasLineBus like any other Bus in your energy system.  I also added a file with an example. To be honest I never shared something on github or any other platfrom before therefore I am not sure what I can do to to get the GasLineModell in the base version of oemof. I am glad for any feedback.

philipp235:
The basic idea behind the GasLineModel is that an energy flow only occurs when there is some form of potential difference like temperature, voltage, pressure etc. Therefore you only have to find the mathematical connection between the potential difference and the resulting energy flow. With the piecewise function, any of these relationships can be modeled and used in oemof. The GasLineModel should therefore only show how this approach can be used. The GasLineBus has the property of the potential and the energy flow that occurs is calculated based on the difference of these potentials. Thepotential of the GasLineBus can be fixed, or gets calculated from the energy flows.

uvchik:
I really like the idea of integrating piecewise linearisation :smile: 

I wonder just if it is possible to create a more generic component from your idea. I could think of something like a Piecewise Linear Converter. If it is to generic for some users we could then add some typical functions for e.g. gas transport etc..

Furthermore, I wonder if it would be easier to add the potential difference to the component not to the bus. Your Converter could also be an electrolyzer with electricity on the one side an hydrogen on the other side.

Such a generic component could be used by many users while a gas line is more specific.

philipp235:
> I really like the idea of integrating piecewise linearisation 😄
> 
> I just wonder if it is possible to create a more generic component from your idea. I could think of something like a Piecewise Linear Converter. If it is to generic for some users we could then add some typical functions for e.g. gas transport etc..
> 
> Furthermore, I wonder if it would be easier to add the potential difference to the component not to the bus. Your Converter could also be an electrolyzer with electricity on the one side an hydrogen on the other side.
> 
> Such a generic component could be used by many users while a gas line is more specific.

I could rename the objects to make them sound more generic. In the implementation you pass a list with x and y values which are used to create the piecewise function. The reason why the nodes have the property of a potential is that changing this property affects all lines from that point. For example, if you lower the pressure at a node so that all lines leading away from that point are affected.

jnnr:
> I really like the idea of integrating piecewise linearisation smile
> 
> I just wonder if it is possible to create a more generic component from your idea. I could think of something like a Piecewise Linear Converter. If it is to generic for some users we could then add some typical functions for e.g. gas transport etc..


There is an open PR #592 which introduces a PiecewiseLinearTransformer. We (stefansc1 and jnnr) are about to finish this up soon (finally!). To be more precise, stefansc1 is doing the last steps on a fork (https://github.com/stefansc1/oemof-solph/tree/features/piecewise_linear_transformer).

There is a PR on oemof_examples https://github.com/oemof/oemof-examples/pull/67 that showcases the new component. 

If any of you would be interested to review, please assign yourself!

philipp235:
I think the PiecewiseLinearTransformer would be also pretty nice in oemof. If I understand der PiecewiseLinearTransformer right it models the conversion loss  non linear. 

In compersion to this the GasLineModel is not capable of doing this. Instead it calculates the energy flow as a non linear function of the potential difference between nodes.
oemof

jnnr:
> In compersion to this the GasLineModel is not capable of doing this. Instead it calculates the energy flow as a non linear function of the potential difference between nodes.

Interesting! I am not sure if I fully understand, though. Is Delta_p given exogenously? Which choices does the solver then have?

philipp235:
> > In compersion to this the GasLineModel is not capable of doing this. Instead it calculates the energy flow as a non linear function of the potential difference between nodes.
> 
> Interesting! I am not sure if I fully understand, though. Is Delta_p given exogenously? Which choices does the solver then have?

It works like the ElectricalLine Object. You need to define a slack an the other potentials get calculated by the solver.

philipp235:
import oemof.solph as solph
import oemof.solph.exnet as ex
from oemof.outputlib import *
import math
import pandas as pd
import time



datetimeindex = pd.date_range('1/1/2017', periods=3, freq='H')

es = solph.EnergySystem(timeindex=datetimeindex)

# This is how you can define the function between delta_p and E
# -1<=delta_p<=1
# input_list = x_values
# output_list = y_values

input_list = []
input_list.append(-1)
i = 0
while i < 101:
    ob = i / 100
    input_list.append(ob)
    i = i + 1
output_list = []
output_list.append(0)

for ob in input_list:
    if ob != -1:
        s = math.sqrt(ob)
        output_list.append(s)
# rest of the example
b_gas1 = ex.GasBus(label='b_gas1', slack=True)
b_gas2 = ex.GasBus(label='b_gas2')
b_gas3 = ex.GasBus(label='b_gas3')


gas_line_12 = ex.GasLine(label='gas_line_12',
                         inputs={b_gas1: solph.Flow(nominal_value=200)},
                         outputs={b_gas2: solph.Flow(nominal_value=200)},
                         input_list=input_list,
                         output_list=output_list,
                         K_1=100,
                         conv_factor=0.99)

gas_line_13 = ex.GasLine(label='gas_line_13',
                         inputs={b_gas1: solph.Flow(nominal_value=200)},
                         outputs={b_gas3: solph.Flow(nominal_value=200)},
                         input_list=input_list,
                         output_list=output_list,
                         K_1=100,
                         conv_factor=0.99)

gas_line_23 = ex.GasLine(label='gas_line_23',
                         inputs={b_gas2: solph.Flow(nominal_value=200)},
                         outputs={b_gas3: solph.Flow(nominal_value=200)},
                         input_list=input_list,
                         output_list=output_list,
                         K_1=100,
                         conv_factor=0.99)

gas_line_32 = ex.GasLine(label='gas_line_32',
                         inputs={b_gas3: solph.Flow(nominal_value=200)},
                         outputs={b_gas2: solph.Flow(nominal_value=200)},
                         input_list=input_list,
                         output_list=output_list,
                         K_1=100,
                         conv_factor=0.99)

source_1 = solph.Source(label='source_1',
                        outputs={b_gas1: solph.Flow(nominal_value=300)})


sink_1 = solph.Sink(label='sink_1',
                    inputs={b_gas2: solph.Flow(nominal_value=100,
                                               actual_value=[1, 1, 0],
                                               fixed=True)})

sink_2 = solph.Sink(label='sink_2',
                    inputs={b_gas3: solph.Flow(nominal_value=100,
                                               actual_value=[1, 0, 1],
                                               fixed=True)})

es.add(b_gas1)
es.add(b_gas2)
es.add(b_gas3)
es.add(gas_line_12)
es.add(gas_line_13)
es.add(gas_line_23)
es.add(gas_line_32)

es.add(source_1)
es.add(sink_1)
es.add(sink_2)

model = solph.Model(es)
model.solve(solver='cbc')
model.results()
results = processing.results(model)

p-snft:
Closed due to inactivity.

PR #727, opened 2020-10-14
Reject 'variable_cost' on Flow objects

psychon:
It took me way too long to find this missing typo: 'variable_cost' when
'variable_costs' was meant.

This commit changes the error mode for this typo: Instead of silently
ignoring the costs, the user would now get a loud complaint about this
typo.

Signed-off-by: Uli Schlachter <psychon@znc.in>

-----

Feel free to take this PR as an issue instead and fix it in a different way if you prefer that.
Also, no idea if the unit test works. I was not able to get things to run (likely because I tried running oemof-solph dev against the latest release of oemof-network).

-----

    * Depending on how you count: This either does not implement any functionality, or this implements a quality-of-life improvement that avoids many wasted hours chasing a typo (after one finally notices that there is a problem at all)
  * Where is it located?
    * I'm sorry, but I do not understand the question.
  * How does the API look?
    * User makes a typo. User gets an exception saying so.
  ...

* Related issues?
  * None. Feel free to treat this as an issue instead.

  * I like strongly typed languages. Dynamic types feel like "please continue even if things go wrong".
  * More seriously: What are other similar typos? Should this be extended into a list of common papercuts that should be dealt with? Does it perhaps even make sense to complain if any attributes are specified which are not understood / which will be ignored? However, no idea how this would be implemented.

  * None, thanks for asking.

* [] For new features: Remember the documentation!
  * Uhm... and don't forget the fifth of November...

### DIFF
--- a/src/oemof/solph/network.py
+++ b/src/oemof/solph/network.py
@@ -169,6 +169,11 @@ class Flow(on.Edge):
                     'negative_gradient': {'ub': None, 'costs': 0}}
         keys = [k for k in kwargs if k != 'label']
 
+        if 'variable_cost' in keys:
+            raise AttributeError(
+                "There is no `variable_cost` attribute,"
+                " did you mean `variable_costs`?")
+
         if 'fixed_costs' in keys:
             raise AttributeError(
                 "The `fixed_costs` attribute has been removed"
--- a/tests/test_components.py
+++ b/tests/test_components.py
@@ -20,6 +20,15 @@ from oemof.solph import NonConvex
 from oemof.solph import components
 from oemof.tools.debugging import SuspiciousUsageWarning
 
+# ********* Flow *********
+
+
+def test_flow_typo_variable_cost():
+    msg = r"did you mean `variable_costs`"
+    with pytest.raises(AttributeError, match=msg):
+        Flow(variable_cost=42)
+
+
 # ********* GenericStorage *********
 
 


p-snft:
We talked about this type of issues several times (e.g. typing "imputs" instead of "inputs"). I advocated the solution to make keyword arguments explicit. That would also have prevented your problem. However, others advocated a post-check after the initialisation, so we now have a 'SuspiciousUsageWarning' instead of raising errors.

psychon:
> so we now have a 'SuspiciousUsageWarning' instead of raising errors.

Uhm, sorry, but where? Or are you suggesting "in general" and I am only misunderstanding you as "this error case is already reported as a warning"?
`git grep Suspicious src` says that it only occurs three times.

The first match is similar to what I am proposing here (but for a removed attribute/argument instead of a typo):
https://github.com/oemof/oemof-solph/blob/ea3e7970ab9545774579323cb158db8de6ce93c6/src/oemof/solph/network.py#L183-L187
The second match seems... well, slightly unrelated. Sure, this would catch your `imputs` example, but this approach does not seem to work with optional arguments like `variable_costs`:
https://github.com/oemof/oemof-solph/blob/ea3e7970ab9545774579323cb158db8de6ce93c6/src/oemof/solph/network.py#L335-L341

p-snft:
The point is that it is allowed to have arbitrary names for (additional) keywords. Blacklisting would interrupt with this design choice. I also think that replacing this feature by an "custom_args" agumement that may contain a dict would make things much easier, but it is also possible in another way:

Analogue to
https://github.com/oemof/oemof-solph/blob/ea3e7970ab9545774579323cb158db8de6ce93c6/src/oemof/solph/network.py#L335-L341
a warning could be raised if an unknown argument is given.

uvchik:
>  I advocated the solution to make keyword arguments explicit. That would also have prevented your problem.

As `variable_costs` is an optional variable and as we still allow `**kwargs` this would not prevent the problem. A wrongly spelled parameter would be accepted as a new keyword argument.

uvchik:
> The first match is similar to what I am proposing here (but for a removed attribute/argument instead of a typo):
> 
> https://github.com/oemof/oemof-solph/blob/ea3e7970ab9545774579323cb158db8de6ce93c6/src/oemof/solph/network.py#L183-L187

We added this warning do make users aware of an API change. This is not for wrongly spelled parameters and I think we will remove it in the future.

I would prefer a function to scan a whole EnergySystem for suspicious attributes. I like the idea to use a dictionary for that.

```python
suspicious attributes = {
    "Flow":  {
        "variable_cost": "variable_costs",
        "varible_costs": "variable_costs",
        "normal_value": "nominal_value",
    "Transformer": {
        "conversion_factors": "conversion_factor",
        "...."...
    "class_name": {
        "misspelled name": "correct name"
```

The API of the function could be like this:
```python
def check_attributes(energy_system, user_dict=None):
    # Add the user_dict to the default dict (optional)
    # Walk through all classes and check the attributes
    # return a dict/DataFrame/Series with suspicious attributes.
```

p-snft:
> > I advocated the solution to make keyword arguments explicit. That would also have prevented your problem.
> 
> As `variable_costs` is an optional variable and as we still allow `**kwargs` this would not prevent the problem. A wrongly spelled parameter would be accepted as a new keyword argument.

It would, as `**kwarg` would have been replaced by some `custom_args={}`. However, we decided for post-ante checks, so, we should proceed with it.

At the moment, both solutions, the one of @psychon and the one of @uvchik, just catch "known typos". A fail-safe version would do proper book-keeping of all known arguments not only in the docstring. This is done, e.g. here:
https://github.com/oemof/oemof-solph/blob/ea3e7970ab9545774579323cb158db8de6ce93c6/src/oemof/solph/network.py#L163-L166

Doing so, one can easily check if a kwarg is in one of the lists and complain if it is not.

uvchik:
> It would, as `**kwarg` would have been replaced by some `custom_args={}`. However, we decided for post-ante checks, so, we should proceed with it.

You are right, if we do not use `**kwargs` but a parameter to catch the custom attributes it would work.

p-snft:
At the developer meeting, we decided to move from **kwargs to explicitly named arguments. So, I will close this PR.

PR #738, opened 2020-12-15
Features/piecewise linear transformer

stefansc1:
* Update PiecewiseLinearTransformer
  * fix implementation in custom.py to use Transfomer from oemof.network as base
  * fix test to use fixed demand value
  * skip auxiliary variables from pyomo.Piecewise when creating dataframe for result

### DIFF
--- a/src/oemof/solph/processing.py
+++ b/src/oemof/solph/processing.py
@@ -80,18 +80,9 @@ def create_dataframe(om):
     # get all pyomo variables including their block
     block_vars = list(set([
         bv.parent_component() for bv in om.component_data_objects(Var)]))
-
     var_dict = {}
     for bv in block_vars:
-        if isinstance(bv.parent_block(), _PiecewiseData):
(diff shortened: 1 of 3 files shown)


src/oemof/solph/processing.py
```diff
@@ -80,18 +80,9 @@ def create_dataframe(om):
...
-        if isinstance(bv.parent_block(), _PiecewiseData):
-            for i in getattr(bv, '_index'):
-                key = (
-                    str(bv.parent_block()).split('.')[0],
-                    str(bv).split('.')[-1] + '_' + str(i),
-                    bv[i].parent_block().index())
-                value = bv[i].value
-                var_dict[key] = value
-        else:
+        if not isinstance(bv.parent_block(), _PiecewiseData):
```
jnnr, line 85:
```suggestion
        # Drop the auxiliary variables introduced by pyomo's PIecewise
        if not isinstance(bv.parent_block(), _PiecewiseData):
```

Issue #664, opened 2019-12-06
New custom DSM component could benefit from extension / further modelling approaches


jokochems:
Hi there,

as I presented in the dev meeting, I'm interested in further developing the DSM component. My idea is twofold:

- First, I'd like to introduce an investment possibility for DSM and to distinct between load shifting and load shedding units. This could / should be realized as an addition to the new DSMSink component integrated in v0.3.2.

- Second, I want to include (slightly) different modelling approaches for modelling DSM which I came accross in the literature in order to do a systematic benchmarking of these. As far as I know, this is lacking in the literature so far and would be quite interesting. My approach will be to keep every single implementation encapsulated in a separate custom component which inherits from a Sink (the new custom DSMSink from v0.3.2 serving as a model).

Details can be found in the presentation I held in the DSM session of the latest dev meeting. which will soon be available (here is the link: https://oemof.files.wordpress.com/2019/12/2019-12-06_demand_response_tub-er.pdf). There are already some code snippets which I will soon tidy up and make available. Anyone, who has tips / comments or who'd like to contribute is warmly welcome.

Tanks a lot
Johannes

jokochems:
Hi there,
I'm terribly sorry that it took around forever and that I'm far behing the original schedule. But here's an overview on the current status quo:
* I've implemented four approaches for load shifting as well as load shedding, each one as a separate component. These are: DIW - Zerrahn and Schill 2015; DLR - Gils (2015); IER - Steurer (2017) and TUD - Ladwig (2018) -> see the link below for the literature info.
* I've added a jupyter notebook for some tests and comparison of the implementations.
* I've put the work in progress into the following repo: https://github.com/jokochems/DR_modeling_oemof
* Investments are still to be added. I'll do that when I finished the actual model comparison (August/September).

Some words on the distinction between shedding and shifting, I chose:
* One unit should be able to do both, load shifting and load shedding.
* Therefore, I introduced a second variable for shedding and adapted the constraints accordingly. This seems to work out fine for all approaches except for the one from DLR.
* An alternative would by to introduce separate shifting components (which could be modelled as Sinks or even as Sources). I did not (yet) implement that.

So far, two issues arose, where your help might be precious:
1. If I try to integrate a second DSM component, I get a KeyError. This is also the case for the current SinkDSM custom component. It seems to have something to do with the grouping and/or maybe not indexing properly. I had a look into the grouping implementation in oemof.solph and oemof.network but did not get everything here.
[...]

PR #740, opened 2021-01-09
Extend custom SinkDSM functionaliy

jokochems:
Hi guys,

I've included a couple of additions to the custom SinkDSM component (in oemof.solph.custom).

**Overview**
These comprise:
* Including load shedding
* Including an investments possibility
* Including the modeling approach from the PhD thesis of Hans Christian Gils (2015) at DLR

**Related issues**
This one is related to
#632
#664

**Insights/Remarks**
* I've made a modeling comparison of Demand Side Management modeling approaches and presented it at the INREC 2020.
I'll add the slides to the upcoming description.
* The conclusion is that the DLR modeling approach outperforms the one currently implemented from the DIW in terms of computation and functionality while its shortcoming can easily be tackled (which I already did by including one constraint from the DIW approach).

**Other comments and questions**
* Some minor changes on the existing component were necessary / made. I'll describe them in depth in an upcoming message (I'm running short in time now).
* One could discuss whether it does make sense to have everything in a (huge) single component or whether to implement different components for every approach (which I did in the first place).

Best regards,
Johannes


jokochems:
Okay, here is a more in-depth description of what is new:

**Changes on existing SinkDSM**
- Rename parameter `method` to `approach`
- Rename methods `interval` and `delay` to `oemof` and `DIW`. Add a new approach `DLR`.
The naming is up for discussion.
- Parameters `demand`, `capacity_up` and `capacity_down`: Define all of those to be normalized, .i.e. . This is necessary for proper investment modeling and also consistent with most other parts of oemof.solph. In a dispatch model, these normalized sequences are multiplied by parameters specifying the maximum, i.e. `max_demand`, `max_capacity_down` and `max_capacity_up`. (Note that this is similar to usign the `fix` attribute and specifying a `nominal_value` when e.g. creating a Sink with fixed inflow in a dispatch model.)
- Add parameter `shed_time` for modeling load shedding, i. e. a process, where load is reduced without compensating for it in a later process. So far, only load shifting has been depicted.
- Include a differentiation in costs as well: The former `cost_dsm_down` became `cost_dsm_down_shift`. The parameter `cost_dsm_down_shed` has been added.
- Parameters `efficiency` as well as `recovery_time_shift` and `recovery_time_shed` have been introduced. These are used for modeling the extensions already described in the paper from Zerrahn and Schill (2015) (DIW).
- Boolean parameters `shed_eligibility` and `shift_eligibility` control whether a unit is eligible for load shifting, load shedding or both. Since I need demand reponse units to be eligible for both in some cases and include their interrelation, I chose this approach over depicting load shedding via some kind of modified source (I discussed this with @uvchik a while ago).
- For the simple `oemof` (resp. `interval`) approach, load shedding is not (yet) depicted.

**Including investments modeling**
- The idea is to use the normalized demand prior to demand response as well as normalized timeseries (resp. sequences) for upshift and downshift capability. So far nothing new, except for the normalization (see above). Investments occur in (maximum) capacity while a fixed percentage share of that capacity is assumed to be eligible for upshift resp. downshift. This is depicted by the parameters `flex_share_up` and `flex_share_down`, both .
- Include a checking for investment attributes in `oemof.solph.custom.SinkDSM`: When using an investment model, `flex_share_up` and `flex_share_down` should be defined. Defining `max_demand`, `max_capacity_down` and `max_capacity_up` in addition (which are used for a dispatch model) would cause overdetermination. Hence, an Error is thrown if this is done.
- Besides that, there is an investment block (for each approach / method) which basically carries pretty much the same constraints as for dispatch modeling. Differences are that an investment investment variable `invest` is introduced and provided with investment bounds. In the constraints formulations `g.demand[t] * g.max_demand[t]` (g is the group, t is the respective timestep) is substituted by `g.demand[t] * self.invest`.
- Of course, an investment costs term is added to the objective function.

**Including the approach from Gils (2015) - DLR**
- The approach is described in the PhD thesis of Gils, Hans Christian (2015):  'Balancing of Intermittent Renewable Power Generation by Demand Response and Thermal Energy Storage', Stuttgart, <http://dx.doi.org/10.18419/opus-6888>, accessed 08.01.2021, pp. 67-70.
- There are some similarities to the approach from Zerrahn and Schill (2015) which is currently integrated in the custom SinkDSM component, but there are differences as well. In particular, the DLR approach does not have that much interlinkage of timesteps in the constraints formulations which leads to computational performance improvements over the DIW approach. Furthermore, it includes an interference time or `shift_time` which limits the time for one upshift resp. one downshift.
- The approach uses balancing variables: So the variable `dsm_up` for initial shift gets an additional `balance_dsm_up` variable which balances the upshift via a downshift in the given delay time.
- In order to depict flexible delay times - a unit with a delay time of 4 is assumed to be eligible for a shift of 1, 2 or 3 timesteps as well - I defined `delay_time` to be an iterable, a list to be precise. Each SinkDSM unit is then linked to that list of delay times. This is depicted via a mapping of units (set `DR`) and delay times (set `H`). A pyomo set `DR_H` is the product from that mapping. All variables except for the shedding variable and the demand response storage level variables are indexed with that set.
- There is one variable `dsm_up_level` depicting a demand response upwards storage level. `balance_dsm_do_level` is depicting a demand response downwards storage level.
- There are optional constraints, such as an overall limit for demand response activations and a daily limit. These introduction of those can be controlled via the boolean parameters `ActivateYearLimit` and `ActivateDayLimit`.
- If the parameter `addition` is set to True, a constraint equivalent to Eq. 10 from Zerrahn and Schill (2015) is introduced which limits the amount of up- and downshifts happening both at a time. This is an extension of the basic formulation which in my view does make sense.
- If the parameter `fixes` is set to True, shifts which cannot be compensated for, are forbidden. If this is not done, there can be some strange effects, esp. at the end of the optimization timeframe. While it isn't explicitly stated in the description, from correspondence with Hans Christian Gils I happen to know that there are other ways they tackle that in their (GAMS-based) model.

Since this got pretty lengthy, I'll lay down the discussion and definitions in another comment.

jokochems:
**Some points for discussion**:
* Fundamental one: Should we put everything in one component or define multiple SinkDSM components differing by approach?
* Naming: Does my renaming make sense or do you want me to undo it for the sake of continuity in using the existing component? If I'd change the naming back, it should be possible to use the component in the same way as before if the default values of `max_demand`, `max_capacity_down` and `max_capacity_up` were all set to 1. In that case, it would be necessary to explicitly set or force them to None for investment modeling where you'd rather want to use `flex_share_up` and `flex_share_down`.

**Issues where help and feedback is wanted**:
* PEP 8 line length: I exceeded it at some points. In my view, this can be tolerated since otherwise it can get even harder to read the constraints formulations. (There is a lot of identation already because of iterations, checking conditions, ...)
* Two of the tests fail: AppVeyor and TravisCI. To me it seems that this has something to do with the python versions. But to be honest, I do not really have experience with the autochecks and would appreciate your help to understand why this happens and how to fix that.
* My implementation of the mapping of demand response units and delay times in the DLR modeling approach leads to the following pyomo warning getting thrown: 
> WARNING: Initializing ordered Set SinkDSMDLRBlock.H with a fundamentally unordered data source (type: set).  This WILL potentially lead to nondeterministic behavior in Pyomo

In my view, this can be ignored since the mapping does work the way it should. Do you think it would be appropriate to suppress (only this) warning?

@uvchik, @gplssm and @p-snft: I'm asking for your review and feedback since we already had some discussions and/or you're familiar with the existing custom SinkDSM component. Thank you very much in advance! :-)

jokochems:
Please find the slides from the INREC conference here: https://github.com/jokochems/DR_modeling_oemof/blob/master/Kochems_Demand_Response_INREC.pdf

- At the INREC conference, I presented an in-depth analysis of how different modeling approaches behave in a (widely) harmonized setting and had a look at some effects of certain constraint formulations.
- I evaluated the modeling approaches from the PhD theses of Martin Steurer (IER) and Theresa Ladwig (TUD) as well.
- I decided not to include the approaches from IER and TUD for the following reasons: The approach from TUD shows some similiarities to the `oemof` resp. `interval` approach already implemented. In addition to that, there are some additional shortcomings in my view compared to the DLR or the DIW approach. The approach from IER is quite hard to parameterize. In my view, it demands a high level of information on generation and demand patterns which you normally do not have upfront and need to gain in an extensive residual load analysis.
- Please find the main results of the comparison on slides 18-34. A simplified mathematical description of the DLR approach is given on slides 50-52.

jokochems:
Here's an overview on the definitions I used for a simple case where demand prior to demand response, capacity for upshifts and capacity for downshifts are all kept constant (of course they may vary in time):
grafik

jokochems:
> **Issues where help and feedback is wanted**:
>     * Two of the tests fail: AppVeyor and TravisCI. To me it seems that this has something to do with the python versions. But to be honest, I do not really have experience with the autochecks and would appreciate your help to understand why this happens and how to fix that.

Looking into the TravisCI outcome, this is probably because I haven't yet adjusted the existing tests, right? I'm on it and try to add updated tests.

p-snft (changes requested):
> Fundamental one: Should we put everything in one component or define multiple SinkDSM components differing by approach?

I would suggest to have multiple DSM classes. You should derive from a basic implementation, if that helps to avoid code duplications.


> Naming: Does my renaming make sense or do you want me to undo it for the sake of continuity in using the existing component? If I'd change the naming back, it should be possible to use the component in the same way as before if the default values of max_demand, max_capacity_down and max_capacity_up were all set to 1. In that case, it would be necessary to explicitly set or force them to None for investment modeling where you'd rather want to use flex_share_up and flex_share_down.

Since you are in `custom`, that's no problem. If you find a better wording, use it. (I have some problems with `max_capacity_down`: At the first glance, it sounds like a limit of shifted energy before the DSM has to be rebalance. Is it possible to use `flex_share_down` if not in invest mode?)


> PEP 8 line length: I exceeded it at some points. In my view, this can be tolerated since otherwise it can get even harder to read the constraints formulations. (There is a lot of identation already because of iterations, checking conditions, ...)

For docstrings, exceeding the maximum number of characters per line is sometimes hard to avoid. I can live with that if it helps making the code readable. (You can add comments to switch specific PEP checks on and off.)

>Two of the tests fail: AppVeyor and TravisCI. To me it seems that this has something to do with the python versions. But to be honest, I do not really have experience with the autochecks and would appreciate your help to understand why this happens and how to fix that.

The test say that there it is "missing 1 required positional argument: 'approach'". This is, as the test still defines e.g. `method='delay',` You should fix that.

> My implementation of the mapping of demand response units and delay times in the DLR modeling approach leads to the following pyomo warning getting thrown:

I think that can be ignored (in custom).

jokochems:
Thanks for your quick response and the feedback, @p-snft!

> > Fundamental one: Should we put everything in one component or define multiple SinkDSM components differing by approach?
> 
> I would suggest to have multiple DSM classes. You should derive from a basic implementation, if that helps to avoid code duplications.

Thanks! I would do so if that is a common agreement. @uvchik, @gplssm: What do you think? 

> > Naming: Does my renaming make sense or do you want me to undo it for the sake of continuity in using the existing component? If I'd change the naming back, it should be possible to use the component in the same way as before if the default values of max_demand, max_capacity_down and max_capacity_up were all set to 1. In that case, it would be necessary to explicitly set or force them to None for investment modeling where you'd rather want to use flex_share_up and flex_share_down.
> 
> Since you are in `custom`, that's no problem. If you find a better wording, use it. (I have some problems with `max_capacity_down`: At the first glance, it sounds like a limit of shifted energy before the DSM has to be rebalance. Is it possible to use `flex_share_down` if not in invest mode?)

Yes, you could use that in the dispatch mode as well. It basically depends on your data input:
* For dispatch modeling, I started with use cases as Guido and Julian have produced, where there are actual upper and lower demand boundaries given.
* Flex shares is what you come accross in the literature. But basically it is just calculating the one from the other (see my scheme above). So if there's a preference, I could adjust that.

> > Two of the tests fail: AppVeyor and TravisCI. To me it seems that this has something to do with the python versions. But to be honest, I do not really have experience with the autochecks and would appreciate your help to understand why this happens and how to fix that.
> 
> The test say that there it is "missing 1 required positional argument: 'approach'". This is, as the test still defines e.g. `method='delay',` You should fix that.

Thanks. I'm in the process of fixing the test.

src/oemof/solph/custom.py
```diff
@@ -1567,9 +2210,2143 @@ def _objective_expression(self):
...
+        (5) \quad DSM_{tt}^{up}  + \sum_{t=tt-L}^{tt+L} DSM_{t,tt}^{do} \leq
+        max \{ E_{tt}^{up}, E_{tt}^{do} \}\quad \forall tt \in \mathbb{T} \\
+        &
+        (6) \quad \sum_{tt=t}^{t+R-1} DSM_{tt}^{up} \leq E_{t}^{up} \cdot L
+        \quad \forall t \in \mathbb{T} \\
+        &
+
+
+
+    **Table: Symbols and attribute names of variables and parameters**
```
p-snft, line 2254: This table reads almost the same as the previous one. Can you work with just one of the tables to avoid code duplication?

src/oemof/solph/custom.py
```diff
@@ -1567,9 +2210,2143 @@ def _objective_expression(self):
...
+            load shift"
+            ":math:`demand_{t}` ",":attr:`demand[t]`","P","Electrical
+            demand series"
+            ":math:`E_{t}^{do}` ",":attr:`capacity_down[t]`","P","Capacity
+            DSM down shift "
+            ":math:`E_{t}^{up}` ", ":attr:`capacity_up[t]`", "P","Capacity
+            ":math:`\eta`",":attr:`efficiency`","P", "Efficiency loss for
+            load shifting processes"
+            ":math:`\R`",":attr:`recovery_time_shift`","P", "Minimum time
+            between the end of one load shifting process and the start of another"
```
p-snft, line 2279:
```suggestion
            ":math:`\R`",":attr:`recovery_time_shift`","P",
                "Minimum time between the end of
                one load shifting process and the start of another"
```

src/oemof/solph/custom.py
```diff
@@ -1567,9 +2210,2143 @@ def _objective_expression(self):
...
+                        lhs = (sum(self.dsm_up[g, h, t]
+                                   + self.balance_dsm_do[g, h, t]
+                                   + self.dsm_do_shift[g, h, t]
+                                   + self.balance_dsm_up[g, h, t]
+                                   for h in g.delay_time)
+                               + self.dsm_do_shed[g, t])
+
+                        # maximum capacity eligibly for load shifting
+                        rhs = (max(g.capacity_down[t] * g.flex_share_down,
+                                  g.capacity_up[t] * g.flex_share_up)
```
p-snft, line 4313:
```suggestion
                        rhs = (max(g.capacity_down[t] * g.flex_share_down,
                                   g.capacity_up[t] * g.flex_share_up)
```

p-snft (changes requested):
While I do not see PEP errors as show stopper for staging ("custom") components, it should be easy to avoid them, here.

src/oemof/solph/custom.py
jokochems: Yes, there are quite some duplicates, thanks! I did the suggested quickfix, but will have a look on the tables and constraint formulations again, since I noticed that they need an update (I have not yet included all the adjustments I needed for investment modeling). But one table with all vars and params is a good suggestion.

jokochems:
I'm sorry, I left this one open for quite a long time because I was focussing on the development in #750. 
I'll resume and provide a proper Demand Response implementation here. I'll take care of potential conflicts with the multiperiod additions. Thanks for the valuable hints and change requests @p-snft.

jokochems:
In my view, this one is finished (eventually). I'm sorry that it took so long.
@p-snft: I included the changes you requested and cleaned up the docs as well as some small bugs.

I fixed the constraints tests and added some usage example here: https://github.com/jokochems/DR_modeling_oemof/blob/master/oemof_PR/DSM-Modelling_oemof-integration-test.ipynb (it's not so very user-friendly, but helped to check if every constraint works the way it should and to spot bugs)

To be honest, I don't understand the outcome of the failing tests, i. e. black, tox checks and appveyor. I'd be happy to receive some help on that since I'm not very experienced with CI/CD and autotests (yet). Thanks!

jokochems:
Okay, now this one is really finished. I udated the changelog and the readthedocs documentation on the SinkDSM component.
I still need help with the failing CI tests, though. Two of them seem to be somewhat flake8-related, but I don't know how to tackle that. Please let me know if something is still missing. Thank you!

jokochems (commented):
This should all be resolved.

jokochems:
Alright, thanks @p-snft for the hint that the CI output only indicated some PEP8 issues which were easy to fix. :-D
The black linters still fail, but I interprete this to be an issue with working on forks vs. the remote which still needs fixing, right?
If this is the case, in my view, this one is now ready for a final review and a merge.

src/oemof/solph/custom/sink_dsm.py
```diff
@@ -15,7 +15,9 @@
 SPDX-License-Identifier: MIT
```
p-snft, line 15: You might want to change the copyright to name our contribution.

p-snft (approved):
Great, thanks! (You should not forget to add your copyright notes to the files you changed.)

src/oemof/solph/custom/sink_dsm.py
jokochems: Right, thanks! It's pushed now. Thanks very much for approving and advising on this one!

jokochems:
Thanks @p-snft for the help on this one and thanks @nailend, @gplssm and @uvchik for the foundation that was build upon.
Looking forward to further tackling #750 now.

PR #741, opened 2021-01-14
Replaced private class import with public one

stefansc1:
This addresses https://github.com/oemof/oemof-solph/pull/592#pullrequestreview-566257556

In processing.py, a private class had been imported from pyomo.core.base.piecewise. This is now replaced by an import of the public class IndexedPiecewise.


p-snft (approved):
Looks good, thank you!

PR #746, opened 2021-02-16
Replace travis with workflows, add 3.9 and remove 3.6

uvchik:
* solph will have to pay for Travis CI in the future
* Therefore we try to use github workflows instead
* We will test Python 3.9 from now on and drop 3.6 support


uvchik:
This will add the following tests:

* Pytest from tox.ini for 3.7, 3.8, 3.9 with coverage
* Checks from tox.ini for 3.7, 3.8, 3.9 (twine,  check-manifest, flake8,  isort )
* Packaging for 3.9

Pytest/Checks is what Travis has done before, packing is new.

I would suggest to add a black-style checker, but @oemof/oemof-solph should agree to that.

uvchik:
[image](https://github.com/oemof/oemof-solph/actions?query=workflow%3A%22tox+checks%22) [image](https://github.com/oemof/oemof-solph/actions?query=workflow%3A%22tox+checks%22) [image](https://github.com/oemof/oemof-solph/actions?query=workflow%3Apackaging)

uvchik:
I just added black to give you an idea how it would work. I fixed all issues but I can revert it if it is unwanted.

I think auto-fixing is nice, too but you will have a Bot at the contributors. After a while the Bot may be one of the top ten developers. This is what I do not like, so I tend to not using the Bot.

You have to change [this value](https://github.com/oemof/oemof-solph/pull/746/files#diff-107e910e9f2ebfb9a741fa10b2aa7100cc1fc4f5f3aca2dfe78b905cbd73c0d2R41) to true, to enable auto fixing.

gnn (approved):
That is pretty cool. As I said, the checks overview and the details for the checks could be a little tidier for me, but as you explained that's not on you, but rather on how GitHub picks what it wants to display. Concerning checking for properly formatted code in the tests, I'm all for it. This is a pretty cool PR. Thanks for doing this.

setup.cfg
```diff
@@ -4,6 +4,7 @@ universal = 1
 [flake8]
 max-line-length = 79
 exclude = */migrations/*, docs/conf.py
+ignore = E203, W503
```
p-snft, line 7: Why is E203 ignored?

p-snft (approved):
…and do a lot of style improvements.

setup.cfg
uvchik: There are a lot of discussions about that in the web. Black decided to use a whitespace before and after a colon if it is used for slicing but flake8 complains about it even though it is in line with PEP8. [[Source]](https://github.com/PyCQA/pycodestyle/issues/373)

p-snft: I see, thank you!

Issue #751, opened 2021-03-09
Use submodules


p-snft:
The current structure could be split into submodules in a way still API-compatible to v0.4. This is easily done and improves readability by a lot. Also, it will help preparing and restructuring the new (private) submodules as suggested in #588.

p-snft:
Done.

PR #752, opened 2021-03-09
Split code into a submodules

p-snft:
This is refactoring only. Every change in this PR should be fully transparent and invisible to the users. In the end, every class should sit in their own file while the API is unchanged.

Implements #751.


p-snft:
The problem detected by the pipelines is due to Black and isort not agreeing on formating imports.

The decreased coverage seems to be an artifact, as nothing really changed.

p-snft:
* The decreased coverage seems to be an artifact, as nothing really changed.
* Codacity dislikes unused imports in the __init__ files.

The second problem is actually not a problem, so I would like to merge.

p-snft:
Todo:

- [ ] Update Docs
- [ ] Adjust copyright notes (?)

src/oemof/solph/blocks/bus.py
```diff
@@ -0,0 +1,70 @@
...
+"""Creating sets, variables, constraints and parts of the objective function
+for Bus objects.
+
+SPDX-FileCopyrightText: Uwe Krien <krien@uni-bremen.de>
+SPDX-FileCopyrightText: Simon Hilpert
+SPDX-FileCopyrightText: Cord Kaldemeyer
+SPDX-FileCopyrightText: Patrik Schönfeldt
+SPDX-FileCopyrightText: Birgit Schachler
+SPDX-FileCopyrightText: jnnr
+SPDX-FileCopyrightText: jmloenneberga
```
p-snft, line 12: These copyright notes are taken from blocks.py. Some of these authors might not have contributed to the new blocks/bus.py. Should we adjust this? (Might be tedious in some cases.)

uvchik (approved):
You changed the structure as decided at the developer meeting. It is a good idea to merge the API-stable changes in a first PR. Thnak you. Go on :smile:

simnh:
I think its good!

jokochems:
> 
> 
> I think its good!

Yeah, me too. Thanks for making the structure of solph clearer and more readable.

uvchik:
I just found out that the references to the docstrings in the documentation are all broken now.

p-snft:
Can you give an example? There is a test and I fixed a lot of references.

PR #753, opened 2021-03-13
Features/refine investments

p-snft:
This PR suggests a new API for investments:
* `Investment` has a new keyword argument `other_needs`. It replaces custom kwargs that served the same purpose.
* The constraints `investment_limit` and `additional_investment_flow_limit` are unified. If a keyword is named, the corresponding entry in `other_needs` is considered, else the summed `ep_costs` are limited.


p-snft:
I am still struggling with the naming of the „other needs“. There are a couple of options:
* Demands (should probably be reserved for energy covered by flows).
* Needs
* Requirements
* Prerequisites

jokochems:
> 
> 
> I am still struggling with the naming of the „other needs“. There are a couple of options:
> 
>     * Demands (should probably be reserved for energy covered by flows).
> 
>     * Needs
> 
>     * Requirements
> 
>     * Prerequisites

If I understood it correctly, you use `other_needs` to impose other limiting factors to investments than the total investment costs, right?
In that case, I would rather not associate it with requirements or needs, but rather suggest
* limiting_factors
* other_aspects
* boundary_args

or something like that.

p-snft:
> If I understood it correctly, you use `other_needs` to impose other limiting factors to investments than the total investment costs, right?

It will be used for limiting factors that are shared between `Storages` and `Flows`. However, there is no strict requirement that a limit is set. Thus, it can be used to calculate things for information only. (I.e. to have the space demand at hand, saved in a variable.)

uvchik:
I cannot see the changes made by this PR due to hundreds of changed lines.

However, I like ```custom_kwargs``` as it is a placeholder for custom keyword arguments that become attributes and can be used in custom constraints.

p-snft:
> I cannot see the changes made by this PR due to hundreds of changed lines that do not belong to this PR.
I based this one on https://github.com/oemof/oemof-solph/pull/752. (We agreed on the general idea, so it will probably avoid double work.) I hope to have that one finished soon, so all changes specific to this branch can be seen.

> However, I like `custom_kwargs` as it is a placeholder for custom keyword arguments that become attributes and can be used in custom constraints.
Generally, I do agree. However, `custom_kwargs` would be an argument to the solph non-Block-class. But as far as I see it, for investments to storages there has to be an attribute to `GenericInvestStorageBlock`. So, to allow identifying keywords that need to be handed to the Block, I suggest to have this extra argument. (Also: investment limits are part of mainline solph, so it's not really a "custom" argument.)

joroeder:
With the API of `other_needs` a nested structure is introduced. Maybe, I did not follow all of the discussion, but the purpose with that is to implement explicitly named arguments (#727), right?

p-snft:
> With the API of `other_needs` a nested structure is introduced. Maybe, I did not follow all of the discussion, but the purpose with that is to implement explicitly named arguments (#727), right?

There are two reasons:
1. Arguments providing "core" functionality should have explicit names. (#727)
2. We need a way to signify which arguments need to generate a variable for the optimiser.

PR #764, opened 2021-05-25
Merge v0.4.2 release into dev

p-snft:
This is a pyomo hotfix just done at master.


p-snft:
This does not work, as it disregards splitting into submodules.

PR #767, opened 2021-05-29
Improve SinkDSM formulation (formatting, fixes, tests)

### LINKED ISSUE #761: Minor fixes are sensible in SinkDSM
Hi everyone,

I just came across some minor easy to implement quick fixes that would be sensible, I'll implement as soon as I find the time:
* If you specify `flex_share_up` and `flex_share_down`, but set `max_demand`, `max_capacity_up` and `max_capacity_down` all to `None` in a dispatch model, this is not caught in an Error which provides useful user feedback, though it should be.
* In the `objective_value` of the `SinkDSMDIWBlock`(s) the summation should be for the first, not the second timeindex. This won't make a difference as long as your cost values are time-independet, but matters if you consider e.g. different periods or changes in costs over time.

jokochems:
This one adds some minor improvements to the SinkDSM formulation:
* Extend the checks of infeasible attribute combinations
* Fix the minor bug in the objective_value formulation for `approach="DIW"`, mentioned in #761
* Improve (consistency and hopefully readability) of code formatting
* Add constraint tests for the investment modeling implemented 

Resolves #761

I guess there is no need to adapt anything in the changelog since it is only some minor fixes and reformatting.

### DIFF
--- a/src/oemof/solph/blocks/non_convex_flow.py
+++ b/src/oemof/solph/blocks/non_convex_flow.py
@@ -10,12 +10,14 @@ SPDX-FileCopyrightText: Patrik Schönfeldt
 SPDX-FileCopyrightText: Birgit Schachler
 SPDX-FileCopyrightText: jnnr
 SPDX-FileCopyrightText: jmloenneberga
+SPDX-FileCopyrightText: Johannes Kochems (jokochems)
 
 SPDX-License-Identifier: MIT
 
 """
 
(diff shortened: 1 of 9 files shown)


jokochems:
@p-snft: Would you be so kind to quickly review this one and merge it in case it is okay? Thank you a lot!
I think the failing black issue is the same as last time (problem with forks).

PR #769, opened 2021-05-29
Features/nonconvex gradients

### LINKED ISSUE #768: NonConvexFlow is missing gradient constraints
I noticed that there is an open TODO in the code for this.

Actually, this should be fairly simple to tackle: If I'm not mistaken, the constraint formulation is pretty much the same as for continuous variables except that the flow values have to be multiplied with the binary status variable.

I'll self assign and open a PR for this.

jokochems:
This introduces gradient attributes and constraints for nonconvex flows:
* Enhance `solph.options.NonConvex` by `positive_gradient` and `negative_gradient` dicts.
* Add gradient constraints in `solph.blocks.non_convex_flow.NonConvexFlow`.
* Add checks for infeasible attribute combinations.
* Add tests.

This one resolves #768.

I'm sorry for including the commit history for #767 with it.


jokochems:
@p-snft: Would you please review this one? I think you have been working with the nonconvex option quite a lot. To be honest, I have rarely used it so far, but just wanted to provide this small enhancement.

jokochems:
I rethought this one:

Problems:
* While the general architecture is working, the constraints themselves need adjustment.
* There is a problem with switching a plant off for positive gradients resp. switching it on for negative gradients. The left-hand side term gets smaller than zero in both cases.

Ideas to tackle:
* One option would be to use the existing formulation and to restrict to the absolute values of the left-hand side term.
* Another option would be to use a slightly different constraint formulation, such as the one in Meus et al. (2017), eq. (10) and (11). DOI: https://doi.org/10.1109/TPWRS.2017.2736441

PR #775, opened 2021-06-05
Docs/revise docstrings

jokochems:
Adapt docstrings of the following modules:
* _models.py
* _options.py
* components.experimental._sink_dsm.py
* components._sink.py
* components._source.py

PR #779, opened 2021-07-05
Docs/revise docstring - OffsetTransformer

joroeder:
This PR fixes the docstring of OffsetTransformer.

 * [x] Improve table
 * [x] Add symbols to parameter
 * [ ] Fix link to external base class


joroeder:
There seems to be a problem with (all?) links to external classes in the docstring, e.g. ``pyomo.core.base.block.SimpleBlock`` I did not figure out how to solve that. Could somebody help?

For the `OffestTransformer`, I changed the base class, because this seems to make sense anyway to me, as there are some general checks in ```oemof.solph.network.transformer.Transformer```, the Offset Transformer is now benefitting from. I did not encounter a problem with that and all tests passed.

jokochems:
I'll close in favor of #775 where I copied the changes to (due to not having access to the fork).

Issue #758, opened 2021-04-21
Time step of "initial_storage_level"


p-snft:
At the moment, "initial_storage_level" signifies the storage level before the first time step. I's suggest to change that to the 1st time step, as this would simplify formulating the constraints.

(Is there any reason to stick to the current definition?)

jokochems:
Hi @p-snft,

I don't know the history, but I always found that very confusing as well and wondered why it had been implemented that way in the first place. Thanks for bringing it up.

p-snft:
By now, I have probably found a reason to have it the current way: It might save one constraint per storage.

uvchik:
If you use the `initial_storage_state` to the first time step you need a `n+1` time step to store the storage state at the end.

I recently discussed that topic with some other users. They suggested to distinguish between state variables and interval variables. For an hourly year we would have 8760 intervals and 8761 states. If we use a datetime index the state variables will be quite clear. For the interval variables we have to make clear if they right-stamped or left-stamped.

So in future we would have to pass a datetime index with n+1 values which also makes it also possible to get the intervals from the datetime index. At the moment assume that the last interval is the same as all the others but this is not possible for non-equidistant time increments.

jokochems:
We had a discussion on this and the difference between time stamps (points in time) and time intervals (time spans) on the dev meeting https://github.com/oemof/oemof/issues/96. So when you simulate a year with 8,760 hours in hourly frequency, you actually have 8,761 values for the storage level when you count the initial one.

@uvchik will provide a suggestion on how to deal with that.

But I tink, for the solution, we have to be aware that the `initial_storage_level` defined so far will be active for the hour **before** the start of the simulation, so the timestep -1 which is not contained within the models' TIMESTEPS (as @p-snft mentioned when creating this issue). I think this was not 100% made clear in our discussion on the dev meeting. But anyways, it should be very clearly documented in the best case using some sort of timeline visualization in the docs.

jokochems:
I think, #811 will resolve this, won't it?

p-snft:
Done.

PR #782, opened 2021-07-21
A suggestion for incresing the users understanding of timeindex and timeincrement

jjurczyk:
Hello everybody,

I am coming back to issue #758. Looking at all tests and the general usage
of omoef-solph I think for a user it is the most convenient way to just
initialize the energy system with a timeindex.

But one has to be very clear about the definition of all linear equations and
their meaning. Looking at a storge, one realises that the solutions of the
(MI)LP problem rely internally on the soly timeincrement $\tau_{i}$ variable. Therefore
the results have the following interpretation:

All *flows* in the system represent the value **during** the timeincrement $\tau_{i}$,
while all *energy states* represent the value at the **end** of the
timeincrement $\tau_{i}$

In order to pin this behavior, following changes were implemented:

The `oemof.solph.models.BaseModel` now calculates the timeincrements from the
timeindex, if it is given by a user and no timeincrement is given. Then
the resulting timeindex attribute is truncated at the initial timeindex.

If a timeincrement is set, the timeindex argument is ignored and reset with a
timeindex inline with the timeindex.

This means when using it with an timeindex

```python
>>> import pandas as pd
>>> index = pd.date_range("1/1/2021",period = 8760, freq ="H")
```

not a lot would change. The result timeindex first item would be '2021-01-01 01:00:00'
and the last '2021-12-30 23:00:00'. If one wants to include the last hour of the
year one needs to set period to 8760+1.

Also the `auto_construct` argument, in `omoef.solph.models.Model` is now always
set to false when initializing the parent class. After finishing the initialization
of the parent class the _construct this called. This probably needs discussion, since I
was unsure about the internal mechanism of initialization. To me it seems, the construct
should reside only in `omoef.solph.models.Model`.

The pyomo construction now builds upon `oemof.solph.models.BaseModel.timeindex`
and `oemof.solph.models.BaseModel.timeincrement` and not `oemof.solph.models.BaseModel.es`.

Furthermore all tests with period were set to period+1.

This is a first suggestion on how to improve the interpretation of results, which
was a little fuzzy before, since providing a timeindex it resulted in a timeincrement of the
same length.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+
 *.py[cod]
 __pycache__
 
@@ -69,3 +70,5 @@ docs/_build
 
 # Mypy Cache
 .mypy_cache/
+# vs code
+.vscode
--- a/src/oemof/solph/models.py
+++ b/src/oemof/solph/models.py
@@ -337,7 +337,8 @@ class Model(BaseModel):
 
     def __init__(self, energysystem, **kwargs):
         # the auto_construct shows underdefined behavior since the parent is
-        # not fully initialised => timeindex & timeincrement are not set properly
+        # not fully initialised => timeindex & timeincrement are not
+        # set properly
         # There fore this "hack"
         kwargs_ = kwargs.copy()
         kwargs_.update({"auto_construct": False})
--- a/src/oemof/solph/models.py
+++ b/src/oemof/solph/models.py
@@ -21,6 +21,7 @@ from pyomo.opt import SolverFactory
 from oemof.solph import blocks
 from oemof.solph import processing
 from oemof.solph.plumbing import sequence
+import pandas as pd
 
 
 class BaseModel(po.ConcreteModel):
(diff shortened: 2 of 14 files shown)


p-snft:
Hi there, sorry for not recognising your PR so far. I am currently in the process of shaping the API for the upcoming release of _solph v0.5_, which focuses on simplifying things. Thus, your contribution might be a valuable addition. However, I am currently wondering weather pandas should be a fixed dependency or not. (This might be something to discuss at https://github.com/oemof/oemof/issues/96.)

uvchik:
@jjurczyk In PR #811 you will find our new approach. We discussed it at the last developer meeting and decided to break the API to make the indexing more consistent. I think the new approach will also meet your needs. Please let me know if you have problems.

uvchik:
As there was no reaction from the original author I will close this PR in favour of PR #811

Issue #787, opened 2021-09-14
Add CITATION.cff


p-snft:
I think, we should add a [CITATION.cff](https://citation-file-format.github.io/). It adds "human- and machine-readable citation information" and is supported i.e. by GitHub and Zenodo.

jokochems:
Good idea, I just did so for another repository. It is easy to do and kind of straightahead.

PR #788, opened 2021-09-14
Create CITATION.cff

p-snft:
Implements #787.


uvchik:
Do you have an idea on how to deal with the github-names?

p-snft:
I see two possible extreme options:
1. Cite using nicknames.
2. Do not cite.

I started with option 2, as I did not want to add something and delete it shortly afterwards. But at the moment, I'd opt for Option 1. Whoever wants to change to the real name can do that later.

PR #797, opened 2021-12-02
Raise error if minimum storage level is greater than initial one

jokochems:
Raises a value error if minimum storage level is greater than initial one.


jokochems:
I'll close this one and directly integrate into the v0.5.0 branch.

Issue #796, opened 2021-12-01
InvestmentFlow wrapper


p-snft:
Formulate https://github.com/oemof/oemof-solph/blob/v0.5/src/oemof/solph/flows/_investment_flow.py analogue to https://github.com/oemof/oemof-solph/blob/v0.5/src/oemof/solph/flows/_non_convex_flow.py.

p-snft:
Thanks. This has become obsolete with #856.

PR #799, opened 2021-12-02
Implement Investmentflow wrapper

joroeder:
Implements #796. And aims to implement (the new) a template for docstrings.

 * [ ] implement
 * [ ] add test
 * [ ] update whatsnew
 * [ ] update documentation

### DIFF
--- a/src/oemof/solph/flows/_investment_flow.py
+++ b/src/oemof/solph/flows/_investment_flow.py
@@ -75,6 +75,7 @@ class InvestmentFlow(Flow):
             "max",
             "positive_gradient",
             "negative_gradient",
+            "investment"  # needs to be removed ?
         ]
 
         for key in flow_kwargs.keys():
--- a/src/oemof/solph/flows/_investment_flow.py
+++ b/src/oemof/solph/flows/_investment_flow.py
@@ -32,7 +32,33 @@ class InvestmentFlow(Flow):
     r"""
     Flow with a variable flow capacity usually for investment optimisations.
 
-    """
+    Parameters
+    ----------
+    maximum : float, :math:`P_{invest,max}` or :math:`E_{invest,max}`
+        Maximum of the additional invested capacity
+    minimum : float, :math:`P_{invest,min}` or :math:`E_{invest,min}`
+        Minimum of the additional invested capacity. If `nonconvex` is `True`,
+        `minimum` defines the threshold for the invested capacity.
+    ep_costs : float, :math:`c_{invest,var}`
+        Equivalent periodical costs for the investment per flow capacity.
+    existing : float, :math:`P_{exist}` or :math:`E_{exist}`
+        Existing / installed capacity. The invested capacity is added on top
+        of this value. Not applicable if `nonconvex` is set to `True`.
+    nonconvex : bool
+        If `True`, a binary variable for the status of the investment is
+        created. This enables additional fix investment costs (*offset*)
+        independent of the invested flow capacity. Therefore, use the `offset`
+        parameter.
+    offset : float, :math:`c_{invest,fix}`
+        Additional fix investment costs. Only applicable if `nonconvex` is set
+        to `True`.
+
+
+    For the variables, constraints and parts of the objective function, which
+    are created,
+    see :class:`oemof.solph.flows._investment_flow.InvestmentFlowBlock`.
+
+    """  # noqa: E501
 
     def __init__(
         self,
--- a/src/oemof/solph/flows/_investment_flow.py
+++ b/src/oemof/solph/flows/_investment_flow.py
@@ -52,7 +52,8 @@ class InvestmentFlow(Flow):
     offset : float, :math:`c_{invest,fix}`
         Additional fix investment costs. Only applicable if `nonconvex` is set
         to `True`.
-
+    **kwargs :
+        Parameter of the :class:`oemof.solph.flows._flow.Flow`
 
     For the variables, constraints and parts of the objective function, which
     are created,
--- a/src/oemof/solph/flows/_investment_flow.py
+++ b/src/oemof/solph/flows/_investment_flow.py
@@ -69,7 +69,7 @@ class InvestmentFlow(Flow):
         existing=0,
         nonconvex=False,
         offset=0,
-        **kwargs,
+        **flow_kwargs
     ):
         investment = Investment(
             maximum=maximum,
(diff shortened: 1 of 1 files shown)


joroeder:
I am wondering:

1) how to avoid docstring duplications? Or rather reference docstring of ´Investment´ object?

2) I am implementing **kwargs ? Or didn't we want to avoid them?

src/oemof/solph/flows/_investment_flow.py
```diff
@@ -23,16 +23,63 @@
...
+        self,
+        maximum=float("+inf"),
+        minimum=0,
+        ep_costs=0,
+        existing=0,
+        nonconvex=False,
+        offset=0,
+        **kwargs,
+    ):
+        instvestment = Investment(
```
jokochems, line 74: There is a typo in variable name `investment`

joroeder:
Feedback from dev meeting:
 
>     1. how to avoid docstring duplications? Or rather reference docstring of ´Investment´ object?

Reference Investment class.

For flow attributes, reference flow.
 
>     2. I am implementing **kwargs ? Or didn't we want to avoid them?

Make them explicit.

src/oemof/solph/flows/_investment_flow.py
joroeder: yes, thanks!

src/oemof/solph/flows/_investment_flow.py
```diff
@@ -69,7 +53,7 @@ def __init__(
         existing=0,
         nonconvex=False,
         offset=0,
-        **kwargs,
+        **flow_kwargs
```
p-snft, line 56: Here, you just rename kwargs to flow_kwargs but still allow arbitrary keyword arguments.

joroeder:
Should the test be updated? (Remove all investment attributes from the test) This has partly be done. However, should there still be the options of "manually" putting the investment attribute in the Flow? Then, it would be good, if there remain the "old" tests. And additionally, write new tests for the InvestmentFlow-"wrapper" ... What do you think?

p-snft:
> Should the test be updated? (Remove all investment attributes from the test) This has partly be done. However, should there still be the options of "manually" putting the investment attribute in the Flow? Then, it would be good, if there remain the "old" tests. And additionally, write new tests for the InvestmentFlow-"wrapper" ... What do you think?

It's still possible to use the option object, so it can be used in the tests.

src/oemof/solph/flows/_investment_flow.py
joroeder: Yes, however I added a check for the flow attributes below - so does this not count as pseudo- explicit? ;-)

Actually, I am wondering how to proceed here probably, and not repeat the whole default kwargs setting from the Flow class: [https://github.com/oemof/oemof-solph/blob/0c5031daf3ef816f9cb6f6305ea5b2dff84a24fd/src/oemof/solph/flows/_flow.py#L141-L209](https://github.com/oemof/oemof-solph/blob/0c5031daf3ef816f9cb6f6305ea5b2dff84a24fd/src/oemof/solph/flows/_flow.py#L141-L209)

In the end, it would make sense to make the kwargs of the Flow class explicit first. Then, we could adapt the flow in another PR, and remove the kwargs from the NonConvexFlow and the InvestmentFlow in following PRs. What do you think?

joroeder: I am wondering how to make the flow kwargs explicit at all as we allow additional arbitrary keywords as "emission_factor" and for additional constraints. So, we need to have **kwargs anyway?

uvchik: This is the idea of `flow_kwargs`, but I wonder why there are still stars. In my opinion it should be like this:

```python
    def __init__(
        self,
        maximum=float("+inf"),
        minimum=0,
        ep_costs=0,
        existing=0,
        nonconvex=False,
        offset=0,
        flow_kwargs)
 ```
 The new API would therefore look like this:
 ```diff
- Flow(variable_costs=5, emissions=10)
+ Flow(variable_costs=5, flow_kwargs={"emissions": 10})
```

joroeder: That sounds reasonable. Therefore, this API needs still to be implemented in v0.5 branch, or am I missing something?

[https://github.com/oemof/oemof-solph/blob/fec3f07d3d3a5e560ae09ef4b78743abc08fd27c/src/oemof/solph/constraints/integral_limit.py#L87](https://github.com/oemof/oemof-solph/blob/fec3f07d3d3a5e560ae09ef4b78743abc08fd27c/src/oemof/solph/constraints/integral_limit.py#L87)

p-snft: As you do not touch the class `Flow`, you do not break the API of `Flow`. (However, I already did by moving it to a submodule `flows`.)

jokochems: > This is the idea of `flow_kwargs`, but I wonder why there are still stars. In my opinion it should be like this:
> 
> ```python
>     def __init__(
>         self,
>         maximum=float("+inf"),
>         minimum=0,
>         ep_costs=0,
>         existing=0,
>         nonconvex=False,
>         offset=0,
>         flow_kwargs=None)
> ```
> 
> The new API would therefore look like this:
> 
> ```diff
> - Flow(variable_costs=5, emissions=10)
> + Flow(variable_costs=5, flow_kwargs={"emissions": 10})
> ```

I like the idea of having to explicitly place the additional arguments in a dictionary. This raises awareness for the user. What is more, is that typos for the standard attributes are caught, when you e.g. type `ep_cost` and expect `ep_costs` being set. This is not caught in the current implementation, though. So I suggest to include the change proposed here by @uvchik.

joroeder: Okay, how should we proceed?

In my opinion it makes sense to first implement the explicit keywords in the the Flow, and then proceed with this PR. (And then, also the NonConvexFlow)

p-snft: I'd just follow @jokochems' suggestion. It's easier to address the explicit kwargs in a second PR when this one is already merged.

uvchik:
I talk to @jokochems, @p-snft , @gnn and we think that `options` is more common than `flow_kwargs`.

p-snft:
Closed in favour of #856.

PR #805, opened 2021-12-02
Features/raise error for summed min and max

### LINKED ISSUE #755: UserWarning if summed_max/min without nominal_value is given
If somebody wants to use the `summed_max` (or `summed_min`) contraint for flows, a `nominal_value` must be given. (In the case that it is not an investment flow.)

If only the `summed_max` attribute is given, it is ignored (because value * inf = inf).

So, raising a Warning could help to identify errors in your model setup.

What do you think?

jokochems:
Resolve #755 

If `summed_max` or `summed_min` is set nominal value must be not None.

### DIFF
--- a/src/oemof/solph/flows/_flow.py
+++ b/src/oemof/solph/flows/_flow.py
@@ -221,7 +221,7 @@ class Flow(on.Edge):
 
         if not self.investment:
             warn_msg = (
-                "If {} is set in a dispatch model, "
+                "If {} is set in a flow (except InvestmentFlow), "
                 "nominal_value must be set as well.\n"
                 "Otherwise, it won't have any effect."
             )
--- a/src/oemof/solph/flows/_flow.py
+++ b/src/oemof/solph/flows/_flow.py
@@ -225,14 +225,13 @@ class Flow(on.Edge):
                 "nominal_value must be set as well.\n"
                 "Otherwise, it won't have any effect."
             )
-            if self.summed_max is not None:
-                if self.nominal_value is None:
+            if self.nominal_value is None:
+                if self.summed_max is not None:
                     warn(
                         warn_msg.format("summed_max"),
                         debugging.SuspiciousUsageWarning,
                     )
-            if self.summed_min is not None:
-                if self.nominal_value is None:
+                if self.summed_min is not None:
                     warn(
                         warn_msg.format("summed_min"),
                         debugging.SuspiciousUsageWarning,
--- a/src/oemof/solph/flows/_flow.py
+++ b/src/oemof/solph/flows/_flow.py
@@ -221,20 +221,23 @@ class Flow(on.Edge):
             )
 
         if not self.investment:
-            assumption_msg = (
+            warn_msg = (
                 "If {} is set in a dispatch model, "
-                "nominal_value must be set as well."
+                "nominal_value must be set as well.\n"
+                "Otherwise, it won't have any effect."
             )
             if self.summed_max is not None:
-                assert (
-                    math.isfinite(self.nominal_value),
-                    assumption_msg.format("summed_max")
-                )
-            if self.summex_min is not None:
-                assert (
-                    math.isfinite(self.nominal_value),
-                    assumption_msg.format("summed_min")
-                )
+                if self.nominal_value is None:
+                    warn(
+                        warn_msg.format("summed_max"),
+                        debugging.SuspiciousUsageWarning,
+                    )
+            if self.summed_min is not None:
+                if self.nominal_value is None:
+                    warn(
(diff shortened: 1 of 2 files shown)


p-snft (approved):
Sounds reasonable.

jokochems:
@p-snft: I stated that I don't see the need for checkin that something is finite. Hence checking for None will do the trick. Would you please re-review it? Thank you! :-)

p-snft:
Do you mean, it's too obvious that a nominal value of Inf will break things?

jokochems:
> Do you mean, it's too obvious that a nominal value of Inf will break things?

Hum, I don't now why anyone should place infeasible as a parameter value here anyways. I don't think, we need to catch that. What is more is that the assert statement you proposed doesn't work for NoneType.

fwitte:
I think @p-snft refers to inf as infinite/infinity, or?

src/oemof/solph/flows/_flow.py
```diff
@@ -220,6 +219,24 @@ def __init__(self, **kwargs):
                 + "nonconvex flows!"
             )
 
+        if not self.investment:
+            warn_msg = (
+                "If {} is set in a dispatch model, "
```
joroeder, line 224:
```suggestion
                "If {} is set in a flow (except InvestmentFlow), "
```
I mean you could also think of "mixed" models (but thats really a small thing ;-)

jokochems:
> I think @p-snft refers to inf as infinite/infinity, or?

Yeah, I'm sorry, it was just a typo, I meant infinity as well. And yes, I do not really see why anyone would want to use this in a way like `nominal_value=float('inf')`. Correct me if you see the need to catch it. If I'd require a large value, I'd simply go for someting like `nominal_value=10e10`. I discussed with @joroeder who mentioned it in #755.

p-snft:
> Correct me if you see the need to catch it. If I'd require a large value, I'd simply go for someting like `nominal_value=10e10`. I discussed with @joroeder who mentioned it in #755.

So, you mean, it's too obvious that a nominal value of infinity will break things. Especially when not defining it is disallowed, I believe people might try that. Imagine you don't care what the instantaneous values are but just want to limit the sum. Personally, I try to make a difference between infinity and "some big number" whenever possible. So, my first approach would be to not give a limit. (That's caught now.) If that leads to an error message telling that I need to define a nominal value, I'd set it to infinity. (That would lead to the same internal problem but now without an error message. (FYI: Next, I'd try `Flow(nominal_value="1", max=np.inf)`.)

jokochems:
> > Correct me if you see the need to catch it. If I'd require a large value, I'd simply go for someting like `nominal_value=10e10`. I discussed with @joroeder who mentioned it in #755.
> 
> So, you mean, it's too obvious that a nominal value of infinity will break things. Especially when not defining it is disallowed, I believe people might try that. Imagine you don't care what the instantaneous values are but just want to limit the sum. Personally, I try to make a difference between infinity and "some big number" whenever possible. So, my first approach would be to not give a limit. (That's caught now.) If that leads to an error message telling that I need to define a nominal value, I'd set it to infinity. (That would lead to the same internal problem but now without an error message. (FYI: Next, I'd try `Flow(nominal_value="1", max=np.inf)`.)

Yep, agreed. I'll re-integrate the assert check for a finite values as well.

jokochems:
As you mentioned, @p-snft, and as stated in #755, I also included the checks for `nominal_value` and `max` to be finite as a prerequisite. Would you please review and merge if it is okay (i.e. provided there is no rare "Günni" usecase for infinite values)?

uvchik:
I wonder why you raise a warning instead of an error. It is not possible to defined a `summed_max` without a `nominal_value`.
The problem are the following lines:

https://github.com/oemof/oemof-solph/blob/cd975976073c7734c58c9a38977611cfcf34954c/src/oemof/solph/flows/_flow.py#L361-L379

I would remove the nominal value check. In that case the following error will be raised:
```TypeError: unsupported operand type(s) for *: 'int' and 'NoneType'```
This is not very obvious so you can change your warning to an error and everything is fine. Ignoring a parameter is very hard to debug :exclamation: 

```diff
        # ########################## SETS #################################
        # set for all flows with an global limit on the flow over time
        self.SUMMED_MAX_FLOWS = Set(
            initialize=[
                (g[0], g[1])
                for g in group
                if g[2].summed_max is not None
-               and g[2].nominal_value is not None
            ]
        )


        self.SUMMED_MIN_FLOWS = Set(
            initialize=[
                (g[0], g[1])
                for g in group
                if g[2].summed_min is not None
-               and g[2].nominal_value is not None
            ]
        )
```

uvchik:
Closed in favour to #819

PR #772, opened 2021-06-03
Revise docstring

uvchik:
[Post updated and edited by @jokochems on 10.11.2022]

The actual documentation of this PR: https://oemof-solph.readthedocs.io/en/docs-revise-docstrings/reference/

We have dozen of wrong references in our docstrings and some equations are unreadable. It is a lot of work for one person to fix all the docstrings but it is NOT so much work to fix one or two docstrings. So please help.

1. clone solph
2. checkout `docs/revise-docstrings`

Use `tox` to build the documentation and see the warnings.

1. Open your Python environment
2. Uninstall solph (if it is installed)
2. Use `pip install -e /your/path/oemof-solph` to install a modifiable version of solph
2. Move to your clone of the solph repository (`cd /your/path/oemof-solph`)
3. Install/update `tox`: `pip install tox --upgrade`
4. Use `tox -e docs` to build the documentation and see the warnings 
5. You will find the html-version of the docs in `/your/path/oemof-solph/dist/docs/index.html`

You want to help but you are unsure what to do? Write a comment or contact me. If there are many questions we could make a Q&A-WebCo. It is a good first contribution but we need the help of all contributors, so that it is less work for each person.

You can also join the chat to ask questions: https://matrix.to/#/#oemof.solph:matrix.org?via=matrix.org

Try to finish your work within one month after you assigned yourself.

I did not check all docstrings, so you do not have to do anything if everything is fine :exclamation: 

You should fix all failing references. References like `:py:attr:` will never work. You can use `:py:class:` or `:py:func` e.g.:
:py:class:`~oemof.solph.components.generic_storage.GenericStorageBlock`

The eqautions should be readable e.g. https://oemof-solph.readthedocs.io/en/latest/reference/oemof.solph.html#oemof.solph.components.offset_transformer.OffsetTransformerBlock

Everything else should be readable, look nice, and contain all needed information.

:red_circle: open
:large_orange_diamond: assigned
:heavy_check_mark: done

 * [x] bus.py (all) (@uvchik, June-3) :heavy_check_mark:
 * [x] flow.py (all)  (@uvchik, June-3) :heavy_check_mark:
 * [x] investment_flow.py (@uvchik, June-3) :heavy_check_mark:
 * [x] transformer.py (all) (@uvchik, June-3) :heavy_check_mark:
 * [x] non_convex_flow (@p-snft ) :heavy_check_mark:
 * [x] extraction_turbine_chp.py (@lensum) :heavy_check_mark:
[...]

PR #807, opened 2021-12-02
Fix flow api build

joroeder:
  * [ ] Fixes the API build of the restructured modules
      * [ ] Flows
      * [ ] Buses
      * [ ] ...
  * [x] Updates the version numer to v0.5.0.dev0
      * [x] __init__
      * [x] setup.py
      * [x] VERSION
      * [x] docs/config.py
      * [x] whatnew (was already updated)

### DIFF
--- a/VERSION
+++ b/VERSION
@@ -1,1 +1,1 @@
-__version__ = "0.4.3.dev0"
+__version__ = "0.5.0.dev0"
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -30,7 +30,7 @@ project = "oemof.solph"
 year = "2014-2021"
 author = "oemof-developer-group"
 copyright = "{0}, {1}".format(year, author)
-version = release = "0.4.5.dev0"
+version = release = "0.5.0.dev0"
 
 pygments_style = "trac"
 templates_path = ["."]
--- a/docs/reference/oemof.solph.rst
+++ b/docs/reference/oemof.solph.rst
@@ -264,43 +264,49 @@ oemof.solph.buses.electrical_bus
     :undoc-members:
     :show-inheritance:
 
+Flows
+-----
 
-oemof.solph.custom.ElectricalLine
----------------------------------
+oemof.solph.flows.experimental.ElectricalLine
+^^^^^^^^^^^^^^^^^^^^^
 
-.. automodule:: oemof.solph.custom.electrical_line
+.. automodule:: oemof.solph.flows.experimental.ElectricalLine
     :members:
     :undoc-members:
     :show-inheritance:
 
-oemof.solph.custom.GenericCAES
-------------------------------
 
-.. automodule:: oemof.solph.custom.generic_caes
+Components
+----------
+
+oemof.solph.components.experimental.GenericCAES
+^^^^^^^^^^^^^^^^^^^^^
+
+.. automodule:: oemof.solph.components.experimental.GenericCAES
     :members:
     :undoc-members:
     :show-inheritance:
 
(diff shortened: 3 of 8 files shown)


joroeder:
To also build the block docstrings (that are not directly include as they are not in the init), it is necessary to include them "manually". This leads to ```oemof.solph.flows._flow.Flow```

Alternatively, we could add the Block classes to the init(s)? Is there anything against? And if it is just for buiding the API docu.

joroeder:
An futher question is, should the API documentation be divided in non-experimental and experimental?

p-snft:
Concerning blocks in the API reference: Strictly speaking, blocks are not part of the API. When just these two options are possible, I'd rather make the blocks part of the public API then displaying protected members. If possible, the block documentation should be inlined into the bus/ flow/ component classes docstring. (@jnnr, @gnn, didn't one of you try that?)

> An futher question is, should the API documentation be divided in non-experimental and experimental?

I'd suggest to have the documentation of an experimental bus/ link/ component within that category. The class documentation itself should state that it is experimental - optimally already in the headline (and thus in the table of contents).

joroeder:
> 
> I'd suggest to have the documentation of an experimental bus/ link/ component within that category. The class documentation itself should state that it is experimental - optimally already in the headline (and thus in the table of contents).

I agree, that is also a fine solution. So, you would also not differentiate within the API indenture level and mark as "experimental"? By importing the user also needs to import from "experimental", right? So this should be clear anyway.

joroeder:
> Concerning blocks in the API reference: Strictly speaking, blocks are not part of the API. When just these two options are possible, I'd rather make the blocks part of the public API then displaying protected members. If possible, the block documentation should be inlined into the bus/ flow/ component classes docstring. (@jnnr, @gnn, didn't one of you try that?)

As far as I remember from the last dev meeting, this is diffucult or not possible. But I agree this would be the most elegant solution. Providing the Bocks docstring (so the formulas and the tables with variables and parameter) only in the documentation (which we are doing for some components anyway) would be another alternative.

jnnr:
> Concerning blocks in the API reference: Strictly speaking, blocks are not part of the API. When just these two options are possible, I'd rather make the blocks part of the public API then displaying protected members. If possible, the block documentation should be inlined into the bus/ flow/ component classes docstring. (@jnnr, @gnn, didn't one of you try that?)

There is something similar already, not inlining one docstring in another, but inlining a part of a docstring in the docs/usage.rst-file:

```
.. include:: ../src/oemof/solph/components/generic_chp.py
  :start-after: _GenericCHP-equations11:
  :end-before: """
```

uvchik:
@jnnr Would be nice if there is something like this within the docstring. Any ideas?

The following workaround does not show the function as headline so that the docstring (only!) is displayed below the docstring of the Bus class. It looks like one docstring.

```python
oemof.solph.buses.Bus
---------------------

.. automodule:: oemof.solph.buses._bus
    :members: Bus
    :undoc-members:

.. automodule:: oemof.solph.buses._bus.BusBlock
    :members:
    :undoc-members:
```

grafik

Even though I am not sure if it is cleaner to link to the docstring of the Block class. On the other hand, not showing the API like in the example above makes clear that you should not use the API.

joroeder:
Summary -- questions:
1) _ underscore in the API reference could be misleading for users.
2) divide into experimental or not. Decisions: no, as you need to import from experimental anyway

Further work is done in #772

p-snft:
As "Further work is done in https://github.com/oemof/oemof-solph/pull/772", I will close this PR.

PR #810, opened 2021-12-10
Features/multi-period

### LINKED ISSUE #793: Multiperiod modeling should be supported
This relates to ongoing work in #750 (now #810). I just noticed that a dedicated issue was missing.

This also resolves related outdated pull requests:
#352
#342

The draft pull request in #750 in principle is functional. I've currently started the process of re-aligning it with the existing solph architecture instead of creating a parallel one. After carefully rethinking it, in my view this is the only approach that makes sense in terms of maintainability and re-usability.

jokochems:
Hello everyone,

this one replaces #750 and reincludes the possibility to model multi-period investments, i. e. investments may not only occur at the start time of the optimization, but are allowed in every period.

The idea is to directly integrate this into the framework, i.e. to make single period-modeling a special case of multi-period modeling.

It has now been adjusted to the new API structure `v0.5` will bring along. 
Also, processing has been adjusted such that the old structure can be kept for standard models.

Related issue:
- closes #793

Checklist of missing elements to be added:

- [x] Include lifetime / period conversion
- [x] Handle investment storage initial content (by adjusting the definition for multi-period models in the first place)
- [x] Adjust tests
- [x] Add new tests
- [x] Include SinkDSM results processing
- [x] Adjust docs

I'd be happy for any kind of support and interest in testing this under development feature.


jokochems:
I've gone through it again and don't know whether I can manage to make periods flexible in the first place:
* The problem is that I need some linkage with the lifetime tracking and discounting resp. calculating annuities. Therefore, I'd like to stick to the annual definition in the first place. Flexibilizing it could potentially be an upcoming development step (dependend on how users want to use it in the future).
* I've linked the periods directly with timesteps when creating it in the energy system. Thus, it should be possible for an advanced user to deviate from the annual definition. If she/he does so, she/he needs to be aware that the unit for lifetime definition which defaults to years needs to be adjusted and annuities will be calculated on a period basis rather than an annual one.

If you have any idea on how to easily make periods flexible, please let me know. I think providing plain pyomo tuples (period, timestep) could be an alternative option. But also using something linke a `period_timeincrement` seemed non-trivial as you'd have to ensure that this is not out of sync with the regular one.

jokochems:
A short heads up:
* It is still a draft, but I'm on it and want to get it done ASAP.
* Tests and docs need attention and some debugging needs to be done in the course of that.

jokochems:
I'm in the process of finally getting this one done. I'm planning to write tests and adjust docs in the next couple of days/weeks. (A conference contribution and the [AMIRIS open source release](https://dlr-ve.gitlab.io/esy/amiris/home/) blocked me for a while.)

My idea is then to merge this feature from my fork first into a dedicated feature branch in the repo here once the CI requirements are fulfilled.
I'll then create another merge request from the feature branch to integrate it into a stable branch of the framework.

This way, we can then take time to discuss, tidy up all the "teething" and make it a nice feature for everyone while I'm already working on setting up `pommesinvest` with it and get my PhD modelling stuff on track.

jokochems:
Hum, I wanted to get the tests working on this one before adjusting docstrings as a last step (for the preliminary version of this feature; see above).

I managed to get all the pytest suceeding with my local installation.  But once I tried running it all with tox (locally and using the CI pipeline), I end up with a weird pandas error which has to do with multi-indexing / concatenation. The error is the following 
```TypeError: Must pass list-like as `names`.``` in pandas.core.indexes.base.py:1813 (pandas v1.4.1), probably as a consequence of an error in pandas.core.indexes.multi.py:3817.

I hae been using pandas v1.1.4 locally (so kind of an outdated one). But both, the local tox run (with pandas v1.2.4 as well as the CI one with pandas v1.4.1) failed. I'll try to set up another local environment using the latest pandas installation to dig into that and will have a look in the debug mode  myself.

I briefly spoke to @p-snft on this one. In case I can't find / fix it and need help, I'll let you know.

jokochems:
> Hum, I wanted to get the tests working on this one before adjusting docstrings as a last step (for the preliminary version of this feature; see above).
> 
> I managed to get all the pytest suceeding with my local installation. But once I tried running it all with tox (locally and using the CI pipeline), I end up with a weird pandas error which has to do with multi-indexing / concatenation. The error is the following `` TypeError: Must pass list-like as `names`. `` in pandas.core.indexes.base.py:1813 (pandas v1.4.1), probably as a consequence of an error in pandas.core.indexes.multi.py:3817.
> 
> I hae been using pandas v1.1.4 locally (so kind of an outdated one). But both, the local tox run (with pandas v1.2.4 as well as the CI one with pandas v1.4.1) failed. I'll try to set up another local environment using the latest pandas installation to dig into that and will have a look in the debug mode myself.
> 
> I briefly spoke to @p-snft on this one. In case I can't find / fix it and need help, I'll let you know.

I have been able to fix it myself. If you are interested, here is what happened:
* Since I introduce different indexation for some variables (i.e. all flows and all investment-related ones), I have to take care of this in the results processing. Basically, I do so by explaning the model, what is indexed by timesteps, periods (for investment) or a tuple of both. The problem occured in my adjustments in the results processing for standard models (so not actual multi-period models, but the special case that we only have one period 0).
* I then tried to harmonize indexing to get a common results DataFrame (per node). I did that by replacing index values to tuples of period and timestep, like (0, 3). There is only one period for a standard model, so this value is always 0.
* I separated the overall DataFrame into three, one for each kind of indexation, harmonized the indices and then wanted to combine by a simple call of `pd.concat(list_of_dfs, axis=1)`. Here, the error was thrown.
* I found out that the reason for that was that all DataFrames had an index named `timeindex` consisting of tuples such as (0, 3). As a result, pandas tried to build a MultiIndex in the combined DataFrame, which succeeds in versions prior to v1.2. The `names` attribute is then simply a FrotenList[(None), (None)]. But in versions starting from v1.2, the `names` attribute tries to use the DataFrames index names and ends up being set to `timeindex`, so a simple str value instead of something list-like which is expected and needed.

Fix:
* Instead of trying to bring everything into this tuple shape, it is sufficient to just go with the timesteps values since we only have one period (0) for a standard model that we are used to.
* After doing so, the index of the EnergySystem is used anyways. So it really doesn't matter what is chosen here in this intermediate state. The only thing necessary is to have a proper mapping - which is taken care of.

jokochems:
Uff, results `processing` has really been one of the thoughest ones on this PR. Here is a heads up with some really good news:
* I started by trying to handle all the different variable indexations separately which has proven to be tedious and besides would have been computationally inefficient.
* Luckily, in the process of documenting the whole thing, I came up with an idea on how to vastly simplify again and use the processing as it is / has been.
* From a performance point of view, it now really shouldn't make much difference whether you run your standard model with oemof's status as of v0.5 or this pull request. It is now really just a few minor if / else checks in the investment components (`InvestmentFlowBlock`, `GenericStorageInvestmentBlock` and the `SinkDSMInvestmentBlock`s) as well as one single additional operation in the processing module to have the periods index removed from the "flow" variable again and retrieve what you are used to. So it should be really fast, even for large models (which was my major concern when I introduced so much additional computation in the processing which luckily could all be removed again 😄).

jokochems:
Okay, this one is finished now. 🥳 

As I stated above, I'm going to change merge it into a feature branch in the framework itself to have a working and reusable version here and start building / restructuring our `pommesinvest` model on top of that.

I'll then open another pull request for integration into oemof v0.5 (ideally) where we can discuss the implementation and possible adaptions and changes that you suggest.

PR #811, opened 2021-12-13
Revise indexing of date and time

uvchik:
#### There are two ways to define an index
*It is not possible to define both, because they might contradict each other.*
a. Define a valid pandas.DatetimeIndex.
b. Define an iterable object with the time increments. The number of values is the number of time intervals

#### The system needs three attributes:
1. TIMESTEPS (set), which is the number of intervals
2. TIMEPOINTS (set), which is the number of points, which is one more than the intervals (TIMESTEPS+1)
3. timeincrements (iterable), which is the length of each interval

All three attributes can be taken from a DatetimeIndex (a) or from an iterable with time increments (b).

###New parameter for the EnergySystem
The parameter `infer_last_interval` is `True` by default. This corresponds to the actual behaviour but it only works with equidistant time intervals with a `freq` attribute in the DatetimeIndex. People may implicitly think of 8760 intervals while defining 8760 time points. This is the reason why we keep this behaviour.
Setting the parameter to `False` allows the user the explicitly defines a time index. This works also for intervals with a different length. A numeric index defined by an iterable is always explicit and therefore this parameter does not have any effect.

### Impact on components
All components can now use TIMESTEPS and TIMEPOINTS. Which leads to variables with a different length.
That means that a storage has e.g. 8761 values for the storage content but only 8760 values for the charge/discharge flow.

#### Actual and future default behaviour:
The parameter `remove_last_time_point` of the `results` function is `True` by default if `infer_last_interval` is also `True`. This correspondents with the actual behaviour. If the time index is defined explicitly the parameter is `False` by default because this way the results are also explicit. It is always possible to change the default behaviour. Regardless of the mode, all indexes have the same length.

If the parameter `remove_last_time_point` is `True` the last storage content of a storage can not be found in the results. In case of a balanced storage it is equal to the storage content of the first time step, so it can be found there. Otherwise it has be calculated or ignored. The time stamp of an interval is the start-time of the interval: 00:00 - 01:00 is stamped with 00:00. For time points the stamp is clear. 

If the parameter `remove_last_time_point` is `False` variables that are related to the interval will have one row with nan-values.

### Important
The actual behaviour
A pandas.DatetimeIndex with 8760 periods will lead to 8759 intervals in the `explicit` mode and to `8760` intervals in the `implicit` mode because people may implicitly think of 8760 hours while defining 8760 periods. But the length of one interval is missing, so that the `implicit` behaviour together with a DatetimeIndex does not work without a `freq` attribute in the DatetimeIndex and therefore will only work for equidistant time intervals.

### Discussion
This topic was already addressed  in #782. We had a long discussion about this topic at the developer meeting. The compromise was to allow two modes (`explicit` and `implicit` time mode).

* Now the definition within each mode is consistent.
* It is not possible to define conflicting values anymore
* It is now possible to define non-equidistant intervals using a pandas.DatetimeIndex using the `explicit` mode
* It is now possible to use solph without a timeindex by defining the length of the time intervals and using a numeric index

### API
```python
EnergySystem(timeindex=None, timeincrement=None, infer_last_interval=True, **kwargs)
```

The following might be the content of a following PR.

Valid kwargs for the `solph.EnergySystem` class are:

* `results`: I am not sure what results to pass to the init method of the class and why. A formerly idea was to create a `compute` method to the energy system, which builds a model, solves the model and write the results back to the EnergySystem. Such a method does not exist by now. It is okay to have an `result` attribute but I would remove it as a parameter
* `temporal`: I have no idea what this is for and I could not find anything in the code, so I would remove it.
* `groupings`: Add groupings for external components or for other purpose
* `entities`: You can add objects to the EnergySystem directly with the initialisation using the entities parameter.

#### My conclusion

I would add groupings and entities and remove the kwargs:

```python
EnergySystem(timeindex=None, timeincrement=None, infer_last_interval=True, entities=None, groupings=None)
    if groupings is None:
        groupings = []
    if entities is None:
        entities = ()
```


uvchik:
The test are still failing but I tested the basic concept and it works. I would like to get some feedback before I go on.

Please mark that we now have Variables of the length TIMESTEPS (intervals) and TIMEPOINTS (points). In the storage the content as one more timestep than the (dis-)charge flow. In the end all sequences will have the same length. Either by cutting the first point of the content variable or by adding a nan-row to the interval variables. This could effect other components as well (DSM?).

src/oemof/solph/_energy_system.py
```diff
@@ -32,12 +34,51 @@ class EnergySystem(es.EnergySystem):
...
+            timeindex = timeindex.union(
+                pd.date_range(
+                    timeindex[-1] + timeindex.freq,
+                    periods=1,
+                    freq=timeindex.freq,
+                )
+            )
+
+        if timeincrement is not None and timeindex is not None:
+            raise AttributeError("Don't do it.")
```
jokochems, line 70: "Specifying a timeincrement and a timeindex at the same time is not allowed since these might be conflicting to each other."

src/oemof/solph/_energy_system.py
```diff
@@ -32,12 +34,51 @@ class EnergySystem(es.EnergySystem):
...
+
+        elif timeincrement is None and timeindex is None:
+            pass
```
jokochems, line 73: An error should be raised here, shouldn't it?

src/oemof/solph/_energy_system.py
uvchik: Of course. This was just a placeholder as it is still a draft :smile:

src/oemof/solph/_energy_system.py
uvchik: If you really know what you are doing this could be fine. So we could add a warning. As I remember @gnn needs it like this.

uvchik:
> I meanwhile got a little confused with the usage of `implicit` and `explicit`, though:
> 
>     * `explicit` means that there is an additional timepoint explicitly displayed and you have to provide it in with your input data, right?
> 
>     * `implicit` means that everything is taken care of under the hood and results have 8,760 values for 8,760 data points fed in as they used to.
>       Is it just me or would another wording like `add_timepoint` / `no_add_timepoint` be more intuitive?

We can talk about naming but my basic idea is the following:

Lets talk about an non-leap year with hourly values. Defining a DatetimeIndex with 8760 periods will define 8759 intervals but people implicitly think of 8760 intervals because the last interval is just the same, so the last interval will be added under the hood by adding the 8761st point. But you cannot add an interval if you do not know how long it is. Therefore the `implicit` does only work if you have a `freq` attribute in the DatetimeIndex which defines the length for all(!) intervals. If you have different length it is impossible to know the length of the last time step.
The `explicit` mode explicitly defines the time range with 8761 points and 8760 intervals. That's why I think we should use the implicit mode as default mode because otherwise people would end with 8759 intervals without noticing :thinking: The explicit mode therefore works with different interval length. You can define your DatetimeIndex however you like.
In the end only the length of the interval and the number of intervals is needed by the model, nothing else. The DatetimeIndex is only used for the results.

I hope I could clarify your confusion :smile:

src/oemof/solph/_energy_system.py
```diff
@@ -32,12 +34,51 @@ class EnergySystem(es.EnergySystem):
...
+            )
+            raise TypeError(msg.format(type(timeindex)))
+
+        self.timemode = timemode
+
+        if timemode == "implicit" and timeindex is not None:
+            # Add one timestep to the timeindex.
+            timeindex = timeindex.union(
+                pd.date_range(
+                    timeindex[-1] + timeindex.freq,
```
p-snft, line 63: I know, this is the way, it's currently done, as you also state:

> By now the first TIMEPOINT has been ignored. So all results do have the length of the number of intervals. This behaviour will be kept as default. For a storage the initial storage content can not be found in the results.

However, I consider this very confusing and would recommend changing this when the definitions are changed anyway.

src/oemof/solph/_energy_system.py
```diff
@@ -32,12 +34,51 @@ class EnergySystem(es.EnergySystem):
     oemof.network directly.
     """
 
-    def __init__(self, **kwargs):
+    def __init__(
+        self, timeindex=None, timeincrement=None, timemode="implicit", **kwargs
```
p-snft, line 38: As "explicit"/"implicit" is a binary decision, maybe an option "infer_additional_timestamp=True" would be an alternative option. (I have the feeling that hiding the extra time step from the results is actually a bit misleading.)

src/oemof/solph/components/_generic_storage.py
```diff
@@ -421,70 +424,29 @@ def _storage_content_bound_rule(block, n, t):
...
-        self.balance_first = Constraint(
-            self.STORAGES, rule=_storage_balance_first_rule
-        )
-
-        # storage balance constraint (every time step but the first)
         def _storage_balance_rule(block, n, t):
             """
             Rule definition for the storage balance of every storage n and
-            every timestep but the first (t > 0).
+            every timestep.
```
p-snft, line 444: I really like that simplification!

p-snft (commented):
Please consider always showing the full results (including extra time steps). I cannot see any value in hiding time steps (besides backwards compatibility).

uvchik:
Sorry, I will try again to clarify the problem.

I will use 8760 hours as an example.

1. From now on solph will differ between TIMESTEPS (time intervals) and TIMEPOINTS. This will allow variables with (let's say) 8760 values (variables of intervals) and variables with 8761 values (variables of points). By now the only variable of points is the storage content. Most or all of the other variables are variables of intervals. Now we have three ways to handles this in the results:
     a) Allow results with different length
     b) Cut the first element of the point variables (actual behaviour) 
     c) Add a nan-row to all interval variables
  Some users like b) and some c) so we decided to allow both variants. I really would not allow a) because it is error-prone   
  especially for a numeric index.
2. From now on there are two ways to define the model index (iterable or DatetimeIndex).
   While the definition with am iterable is quite clear there were different opinions on how to define a DatetimeIndex. If you
   want to model with different  interval length you have to define 8761 points to clearly define the length of 8760 intervals. As   
   most people model equidistant intervals some users find it intuitive to define just the start of each interval so the last interval  
   has to be added "under the hood". This is `infer_additional_timestamp=True/False` from @p-snft.
   a) Add an interval at the end
   b) Do not add an interval at the end

My first approach was to combine 1b and 2a, which is the actual behaviour to the `implicit` mode and make it the default behaviour, so for most users everything stays the same. Furthermore, I added an `explicit mode`. which combines 1c and 2b.

Due to the comments above I realised that it is of course possible to combine 1c and 2a, which is a `mixed mode` (implicit index but explicit results).

Think of two thing we could also add two parameters and drop the `mode` idea. One parameter to decide whether to add an interval or not and one parameter to decide whether a nan-row is added or one value is cut off.

We have to decide what kind of design we like better and than talk about the names.

p-snft:
I do understand this. However, is there a good reason against
1 d) Cut the last element of the point variables?

This way, the `initial_storage_level` would be in the (shown) results.

uvchik:
> Thanks for your verbose explanation. I do understand this. However, is there a good reason against
> 1 d) Cut the last element of the point variables?

The weather data is typically right indexed. 1:00 -> 4m/s means that in the hour from 0:00 to 1:00 the average wind speed was 4m/s. If you calculate feed-in power from the wind speed you get 'xy_from4m/s' kW for the index 1:00. If you know use a `Source` you can use the time series for the `fix` attribute. The length of this iterable object should be equal to the number of intervals. An implicit DatetimeIndex looks like this:

```
2012-01-01 00:00:00
2012-01-01 01:00:00
2012-01-01 02:00:00
2012-01-01 03:00:00
...
```
In the approach of this PR the one time point at the end will be added: `2013-01-01 00:00:00`. For the implicit results the first(!) time point is removed.

Now you get a shifted time index:
```
2012-01-01 01:00:00  'xy_from4m/s' kW
2012-01-01 02:00:00 ....
```

As I think about it this is not the actual behaviour but in line with the weather data. 

I am not sure if left/right indexed is the right expression.

@p-snft Your approach would lead to a left indexed result, starting with 2012-01-01 00:00:00 meaning the first interval from 0:00 to 1:00. This is also possible. We just have to document it well. For me the other way is more intuitive but it might be a matter of taste.

uvchik:
I totally agree. Intuitive use always depends on the person. 

In the end people define the intervals and not the points, so everything should be fine.

I am fine with a left index we just have to make it clear in the docs. This is what we have. So we add a time point at the end and remove it afterwards. The problem is that the final storage content is cut of, which makes it more difficult for shifting horizons. But everybody is free to use the `explicit` mode.

p-snft:
Maybe, we can find out what people expect by looking on the current usage. In matplotlib, both options are offered for steps plots:
* 'pre': The y value is continued constantly to the left from every x position, i.e. the interval `(x[i-1], x[i]]` has the value `y[i]`.
* 'post': The y value is continued constantly to the right from every x position, i.e. the interval `[x[i], x[i+1])` has the value `y[i]`.

Thus, I assume that plotting with `steps-post` means left indexing. This is done in:
* [activity_costs/activity_costs.py](https://github.com/oemof/oemof-examples/tree/master/oemof_examples/oemof.solph/v0.4.x/activity_costs/activity_costs.py)
* [basic_example/basic_example.py](https://github.com/oemof/oemof-examples/tree/master/oemof_examples/oemof.solph/v0.4.x/basic_example/basic_example.py)
* [basic_example/basic_example_tuple_as_label.py](https://github.com/oemof/oemof-examples/tree/master/oemof_examples/oemof.solph/v0.4.x/basic_example/basic_example_tuple_as_label.py)
* [generic_chp/bpt.py](https://github.com/oemof/oemof-examples/tree/master/oemof_examples/oemof.solph/v0.4.x/generic_chp/bpt.py)
* [generic_chp/ccet.py](https://github.com/oemof/oemof-examples/tree/master/oemof_examples/oemof.solph/v0.4.x/generic_chp/ccet.py)
* [generic_chp/mchp.py](https://github.com/oemof/oemof-examples/tree/master/oemof_examples/oemof.solph/v0.4.x/generic_chp/mchp.py)
* [invest_non_convex/invest_non_convex.py](https://github.com/oemof/oemof-examples/tree/master/oemof_examples/oemof.solph/v0.4.x/invest_non_convex/invest_non_convex.py)
* [jupyter_tutorials/2_Processing_results_and_plotting.ipynb](https://github.com/oemof/oemof-examples/tree/master/oemof_examples/oemof.solph/v0.4.x/jupyter_tutorials/2_Processing_results_and_plotting.ipynb)
* [min_max_runtimes/min_max_runtimes.py](https://github.com/oemof/oemof-examples/tree/master/oemof_examples/oemof.solph/v0.4.x/min_max_runtimes/min_max_runtimes.py)
* [plotting_examples/storage_investment_plot.py](https://github.com/oemof/oemof-examples/tree/master/oemof_examples/oemof.solph/v0.4.x/plotting_examples/storage_investment_plot.py)
* [start_and_shutdown_costs/startup_shutdown.py](https://github.com/oemof/oemof-examples/tree/master/oemof_examples/oemof.solph/v0.4.x/start_and_shutdown_costs/startup_shutdown.py)

On the other hand, `steps-pre` can only be found in
* [schedule_flow/flow_schedule.ipynb](https://github.com/oemof/oemof-examples/tree/master/oemof_examples/oemof.solph/v0.4.x/schedule_flow/flow_schedule.ipynb)

Explicitly note that it is not consistently used within the group of developers (but at least within the individual examples). So, whatever we do, documentation needs to be improved. (This would also warn people about inconsistent usage of input data.)

jokochems:
> Ah, now I understand! The problem arises as - to my understanding - market data is left indexed. The time stamp gives the point in time where the price changes to the given level. (Honestly, I never thought about right indexed data. For me, it was totally natural that a value indexed 2023-01-01T00:00:00 does not belong to the year 2022 any more. Obviously, I was wrong.)
> 
> PS: I wonder, how often people take time series that are shifted by one your as parallel.

Here's my (totally subjective) intuitive thoughts  - provided sticking to the newly introduced time points:
* For me personally, the left indexing would also be more intuitive and what I would have expected, same as as for @p-snft a. Hence, the start of the first interval, e.g. `2021-01-01 00:00:00` for me intuitively would be the time point, the `initial_storage_level` is valid for.
* To be honest, I found it kind of weird once I detected that the `initial_storage_level` was defined to be before the actual optimization time frame.
* Since it is not clear as of now, I also think this may have led to quite some off-by-one modeling "errors" / other expectations.

Some discussion on how to come to a definition - provided sticking to the newly introduced time points:
* Is there some "standard" storage pattern data available anywhere one could refer to?
* Sponteneously, I'm only aware of ENTSO-E's publication of average weekly filling rates for hydro storages: https://transparency.entsoe.eu/content/static_content/Static%20content/knowledge%20base/data-views/generation/Data-view%20Water%20Reservoirs%20and%20Hydro%20Storage%20Plants.html. But again, these are average values for time intervals, not for (hourly) time points.

A simpler alternative suggestion  - getting rid of the newly introduced time points and potentially increasing consistency:
* You could also think of storage levels as average storage levels over a given time interval which then would also be consistent with the definition of other variables. Thus, this would remove the need for indexing the `storage_content` with the new `TIMEPOINTS` set and lead to it being indexed by `TIMESTEPS`. Also we wouldn't have to take care of an additional point (since actually there are none anymore).
* I've spoken to others and found out that they are actually doing this in such a way. Besides when you read other model documentations, this is what you'll come across, e.g. here https://www.diw.de/documents/publikationen/73/diw_01.c.558112.de/diw_datadoc_2017-088.pdf (page 12/13) or here https://neon.energy/emma-documentation.pdf (page 9)
* So in my view, it might just have been a "historical" matter of how it has been defined in the first place that led to all the confusion ...

Rethinking it, I'd strongly favour the simpler - and I personally think more common - alternative.

Thanks a lot @uvchik and @p-snft for the fruitful discussion.

p-snft:
> * You could also think of storage levels as average storage levels over a given time interval which then would also be consistent with the definition of other variables

I like this idea from a technical point of view. But honestly, I see other problems that arise from a physical perspective, i.e. the extreme levels would be a challenge: If the storage level is zero, this would imply that there has been neither inflow nor outflow in the time step.

uvchik:
In my approach there was not time point before the defined time index. Right index means that the interval results will start with 1:00 if the index starts with 0:00.

But now I will change it. The index will be the start of the interval from now on.

Actually I change three things.

#### 1. Inner Core
First of all I improved the inner model core. Now a point variable has one index more than an interval variable. Which is just right and more consistent.

#### 2. Definition of the Index
For me this is also clearer than it was before. You have two ways to define the number and length of the intervals and you are not allowed to use both at the same time. So can use a DatetimeIndex or a numeric index which is straight forward.

#### 3. Preparing the Results
Because of (1) the content of the storage is no longer split in two variables (first time step and the rest) but together in one variable. This leads to different length of the index. We can solve this with adding a value or removing a value to bring all indexes in line. My new approach is to define this with a parameter.

As mentioned in my comment [above](https://github.com/oemof/oemof-solph/pull/811#issuecomment-995171746) there are two effects that were triggered with the definition of the `timemode` so I thought it is better to name both effects directly.
```python
EnergySystem(self, timeindex=None, timeincrement=None, infer_last_interval=True, **kwargs):
results(model, remove_last_time_point=None)
```
The default value for `infer_last_interval` is True and the default value for `remove_last_time_point` depends on whether  `infer_last_interval` is True or False. So we can keep the actual behaviour (True/True) but do not get the unwanted behaviour (False/True) because the default for the first False is False (False/False). You still have the choice to use (True/False) depending on what you want to do with the results.

We had the discussion of right/left index before, this has nothing to do with this PR but I think it is a good opportunity to solve it now.

> * You could also think of storage levels as average storage levels over a given time interval which then would also be consistent with the definition of other variables. Thus, this would remove the need for indexing the `storage_content` with the new `TIMEPOINTS` set and lead to it being indexed by `TIMESTEPS`. Also we wouldn't have to take care of an additional point (since actually there are none anymore).

I do not want to get rid of TIMEPOINTS, I really like them (s. above)

> * I've spoken to others and found out that they are actually doing this in such a way. Besides when you read other model documentations, this is what you'll come across, e.g. here https://www.diw.de/documents/publikationen/73/diw_01.c.558112.de/diw_datadoc_2017-088.pdf (page 12/13) or here https://neon.energy/emma-documentation.pdf (page 9)

I do not get it. I got the impression that they also use something like content(t)=content(t-1)-discharge(t)+charge(t). If you see t as an index for a time interval you still have the problem that you have one more interval for the content than for charge/discharge. I do not see how this would help.

@jokochems You solved your code comments above to quick. I added your error message and I talked to @gnn. He does not need to define an empty EnergySystem so I followed your suggestion and added an error message if both (timeindex/timeincrement) is None. Thank you for your suggestions :smile:

jokochems:
> > * You could also think of storage levels as average storage levels over a given time interval which then would also be consistent with the definition of other variables
> 
> I like this idea from a technical point of view. But honestly, I see other problems that arise from a physical perspective, i.e. the extreme levels would be a challenge: If the storage level is zero, this would imply that there has been neither inflow nor outflow in the time step.

@p-snft: 
I do not see why an empty storage and no in- and outflows would be an issue if you defined it over the timesteps (which are actually average values for time intervals). It can happen when it is a cheap option to deplete the storage. If you are referring to the `initial_storage_level`, you could just set it other than 0 to model some "realistic" starting value here. What happens before is out of scope for the optimization. Am I not seeing something here?

> > * You could also think of storage levels as average storage levels over a given time interval which then would also be consistent with the definition of other variables. Thus, this would remove the need for indexing the `storage_content` with the new `TIMEPOINTS` set and lead to it being indexed by `TIMESTEPS`. Also we wouldn't have to take care of an additional point (since actually there are none anymore).
> 
> I do not want to get rid of TIMEPOINTS, I really like them (s. above)

@uvchik: I understand that you think of discrete storage levels for snapshot timepoints. But I think, it is more common to have them indexed in the same way as other variables.

> 
> > * I've spoken to others and found out that they are actually doing this in such a way. Besides when you read other model documentations, this is what you'll come across, e.g. here https://www.diw.de/documents/publikationen/73/diw_01.c.558112.de/diw_datadoc_2017-088.pdf (page 12/13) or here https://neon.energy/emma-documentation.pdf (page 9)
> 
> I do not get it. I got the impression that they also use something like content(t)=content(t-1)-discharge(t)+charge(t). If you see t as an index for a time interval you still have the problem that you have one more index for the content than for charge/discharge. I do not see how this would help.

@uvchik: Yeah, that's true. You'll never get rid of the time interlinkage between t and t-1 (or t+1 and t) because that's how the storage works. Thus, for working with zero-based indexing, the 0th value has to be fixed (to the initial level) and you would define the constraint for your time set excluding the 0th element. Most of it would be pretty much what you have implemented, but the indexing of the variable is declared over `TIMESTEPS` instead of `TIMEPOINTS`.

I think, the constraint formulation would / could look like this then:

1. Define the `storage_content` variable in timesteps and fix it's value for the 0th step:

``` python
self.storage_content = Var(
    self.STORAGES, m.TIMESTEPS, bounds=_storage_content_bound_rule
)

# set the initial storage content
for n in group:
    if n.initial_storage_level is not None:
        self.storage_content[n, 0] = (
            n.initial_storage_level * n.nominal_storage_capacity
        )
        self.storage_content[n, 0].fix()
```

2. Constraint formulation:

``` python
def _storage_balance_rule(block):
    """
    Rule definition for the storage balance of every storage n and
    every timestep.
    """
    for n in self.STORAGES
        for t in m.TIMESTEPS:
            if t >= 1:
                expr = 0
                expr += block.storage_content[n, t]
                expr += (
                    -block.storage_content[n, t-1]
                    * (1 - n.loss_rate[t]) ** m.timeincrement[t]
                )
                expr += (
                    n.fixed_losses_relative[t]
                    * n.nominal_storage_capacity
                    * m.timeincrement[t]
                )
                expr += n.fixed_losses_absolute[t] * m.timeincrement[t]
                expr += (
                    -m.flow[i[n], n, t] * n.inflow_conversion_factor[t]
                ) * m.timeincrement[t]
                expr += (
                    m.flow[n, o[n], t] / n.outflow_conversion_factor[t]
                ) * m.timeincrement[t]
            self.balance.add((n, t), (expr == 0))

self.balance = Constraint(
	self.STORAGES, m.TIMESTEPS, noruleinit=True
)
self.balance_build = BuildAction(
    rule=_storage_balance_rule
)
```
So it is basically just simply excluding the 0th timestep and what happened before.
In that case, it would make sense to make the `initial_storage_level` a mandatory attribute since this then directly defines the value for the 0th timestep which you want to start with.

But long story short:
* If you have a strong favour of going towards thinking of storages values at discrete time points, I'd be okay with that.
* One can go either way. The most important thing seems to document (and ideally also visualize) the chosen approach.

p-snft:
> I do not see why an empty storage and no in- and outflows would be an issue if you defined it over the timesteps (which are actually average values for time intervals).

The point is that e.g. an `initial_value` of 0 would imply that there is not inflow in the first time step. If there was an inflow, this would result in a non.zero average storage content over the first time step. Thus, I'd opt to give energy contents at discrete points in time. Further, I would signify this in the examples by plotting Flows with `steps-post` and storage contents by lines that interpolate linearly between the points.

uvchik:
> * If you have a strong favour of going towards thinking of storages values at discrete time points, I'd be okay with that.

I do. Being more exact in this aspect makes it easier to understand your system in the end.

A left indexed makes it even clearer, so do fully agree now, that this is the better option.

> The point is that e.g. an `initial_value` of 0 would imply that there is not inflow in the first time step. If there was an inflow, this would result in a non.zero average storage content over the first time step. Thus, I'd opt to give energy contents at discrete points in time. Further, I would signify this in the examples by plotting Flows with `steps-post` and storage contents by lines that interpolate linearly between the points.

I planned to do it for myself to be sure that everything works as expected. It is very import to document it well no matter which way we use.

uvchik:
I removed the error if an empty EnergySystem is created, because of the following use case:
```python
energysystem = solph.EnergySystem()
energysystem.restore(filename="somthing.slph")
```

gnn:
Just a heads up: I wanted to chime in a while ago but didn't have the time to finish my answer. So please give me at least one more day to do so before merging.

src/oemof/solph/components/_generic_storage.py
```diff
@@ -421,70 +427,29 @@ def _storage_content_bound_rule(block, n, t):
...
         # set the initial storage content
+        # ToDo: More elegant code possible?
         for n in group:
             if n.initial_storage_level is not None:
-                self.init_content[n] = (
+                self.storage_content[n, 0] = (
                     n.initial_storage_level * n.nominal_storage_capacity
                 )
-                self.init_content[n].fix()
+                self.storage_content[n, 0].fix()
```
jokochems, line 440: I know, I had a different argumentation on this, but I spoke again to a colleague of mine:
* It seems not so uncommon to do something like it originally was, i.e. to define the first storage transition (from t=-1 to t= 0) separately. This allows to include some freedom to the model instead of just fixing the first variable value (for t=0) as I suggested in the discussion.
* You basically changed the definition of the `storage_content` from after inflows and outflows to before inflows and outflows of the respective timestep. Thus, the storage level for t=0 is now fixed, whereas you add |T|+1. When you cut this last time point in results processing, you loose an information actually determined by the model in favor of some information you specified externally.

Anyways, it remains a matter of definition. I also see not very much of a problem for modelers (like me) that think rather in the same intervals for all vars instead of disctincting average values for time intervals (for flows) and discrete time points (for storages). Either way, you have to drop one value if you think this way.

A side note: If you want to retain the original behaviour - which it seems we don't, actually, TIMEPOINTS[2] results would correspond with TIMESTEPS[1].

jokochems (commented):
I would be okay with how it is as of now, but I added some things to consider for the storage implementation and results extraction.

uvchik:
We will have an online video discussion on Friday, 7th at 10am. Please contact [me](https://www.uni-bremen.de/en/res/team/dr-ing-uwe-krien) to get the access link.

p-snft:
As there is still stuff to do, I converted this to a draft.

src/oemof/solph/_energy_system.py
```diff
@@ -30,14 +32,93 @@ class EnergySystem(es.EnergySystem):
...
         kwargs["groupings"] = GROUPINGS + kwargs.get("groupings", [])
 
-        super().__init__(**kwargs)
+        if not (
+            isinstance(timeindex, pd.DatetimeIndex)
+            or isinstance(timeindex, type(None))
+        ):
+            msg = (
+                "Parameter 'timeindex' has to be of type "
+                "pandas.datetimeindex or NoneType and not of type {0}"
```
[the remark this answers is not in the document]
jokochems: Still open

uvchik:
Now, a warning is raised if `infer_last_interval` is not defined, so people will be aware of the changes, even though the default behaviour does not changed by now but will change in v0.6.

The default behaviour of the processing has changed. This is API breaking but in most cases it will not effect the code e.g. a plot will just ignore the last `nan` value.

Furthermore, I added a function to create a datetime index (one year by default) easily. A typical EnergySystem could now look like this:

```python
from oemof.solph import EnergySystem, create_year_index

energysystem = EnergySystem(timeindex=create_year_index(2012), infer_last_interval=False,)
```

In the future it would be even simpler:
```python
energysystem = EnergySystem(timeindex=create_year_index(2012))
```

For a 15min interval a quarter of an hour can be defined as follows:
```python
energysystem = EnergySystem(timeindex=create_year_index(2012), interval=0.25)
```

A 2 hour interval can be defined as follows:
```python
energysystem = EnergySystem(timeindex=create_year_index(2012), interval=2)
```

It is also possible to reduce the number of time intervals e.g. for tests. The following code will create a datetime index with 3 intervals and 4 time points.
```python
energysystem = EnergySystem(timeindex=create_year_index(2012), number=3)
```

jokochems (approved):
It looks good to me. Thank you.

simnh:
I do not yet fully understand the whole consequences of the new structure - but I think the discussion is sound enough to merge once the others reviewers agree. I would only suggest renaming the function from `create_year_index` to `create_timeindex` or  just `create_index`   :)

docs/usage.rst
```diff
@@ -51,13 +51,29 @@ Set up an energy system
 
 In most cases an EnergySystem object is defined when we start to build up an energy system model. The EnergySystem object will be the main container for the model.
 
-To define an EnergySystem we need a Datetime index to define the time range and increment of our model. An easy way to this is to use the pandas time_range function.
-The following code example defines the year 2011 in hourly steps. See `pandas date_range guide <https://pandas.pydata.org/pandas-docs/stable/generated/pandas.date_range.html>`_ for more information.
+The model time is defined by the number of intervals and the length of intervals. The length of each interval does not have to be the same. This can be defined in two ways:
+
+1. Define the length of each interval in an array/Series where the number of the elements is the number of intervals.
+2. Define a `pandas.DatetimeIndex` with all time steps that encloses an interval. Be aware that you have to define 8761 time steps to get 8760 intervals.
```
p-snft, line 57: Maybe, use literals (a, b) instead of numbers to underline that this does not describe subsequent steps but are two options.

src/oemof/solph/_energy_system.py
```diff
@@ -30,14 +35,150 @@ class EnergySystem(es.EnergySystem):
...
+    """
+    if number is None:
+        if calendar.isleap(year):
+            hoy = 8784
+        else:
+            hoy = 8760
+        number = hoy / interval
+    return pd.date_range(
+        f"1/1/{year}", periods=number + 1, freq=f"{interval}H"
+    )
```
p-snft, line 184:
```suggestion
        number = round(hoy / interval)
    return pd.date_range(
        f"1/1/{year}", periods=number + 1, freq=f"{interval}H"
    )
```

`pd.date_range` expects an integer for the number of periods. We should round so that e.g. 0.99 is not truncated.

p-snft (approved):
Thanks a lot! I just have two tiny remarks, but no reasons to object.

uvchik:
@jokochems, @p-snft, @gnn  and @simnh Thank you for your feedback

I think we are ready to merge now :rocket:

uvchik:
An internal Attribute changed in Pyomo 6.4.1 so many tests failed. I added the new attribute name in a try/except  statement so that the old and the new name will work (see [24217c9](https://github.com/oemof/oemof-solph/pull/811/commits/24217c92d6c54c15f3033c01f6ec6369c527bc40))

As this PR is about to be merged, this fix will be available for all branches. I will also add this commit to the dev branch.

jokochems:
I'm in the process of updating #810 to the latest version of branch `v0.5.0`. This revision in particular brings along quite some difficulties, but I guess they can be resolved. Nonetheless, updating requires quite some effort.

jokochems:
I know, we had quite some intense discussions on this pull request and @uvchik really put a lot of effort into it which I really appreciate and also find good in terms of coding and some code simplifications in the `GenericStorage` class. Also, I see that when you take a technical perspective on energy systems, you might want to clearly distinguish on state of charge of a storage before and after some inflows.

Anyways, to me personally, from a user and dev perspective, I've experienced quite some difficulties with this feature (usage: in the course of updating a power market model from `v0.4.4` to `v0.5.0a1`; dev: in the course of deciding, how a multi-period investment model should deal with that). I like to think in terms of (hourly) trading intervals and snapshot values where storages aren't treated separately and I personally find feeding in 8761 values to simulate a regular year kind of counter-intuitive. Rethining it, I instead really liked the old definition of an initial storage value before the optimization, i.e. before inflows. What bothers me most is having to deal with nan values in your outcome. If I e.g. do iterative solving and do some calculations first (in my case: market values of renewable sources), I have to cut time series to drop the nan values. Before, I have to append a nan value to the bus balance duals to ensure that indices are not out of line.

I'd like to hear what other users think and what they desire. But maybe, we'll hear that once `v0.5.0` has been officially released.
I don't want to reopen this resp. the associated issue and have found some workarounds for myself, but maybe some awareness for potential future discussions isn't bad. Or maybe it is just me and everyone else finds things to be better and clearer now.

p-snft:
> I like to think in terms of (hourly) trading intervals and snapshot values where storages aren't treated separately and I personally find feeding in 8761 values to simulate a regular year kind of counter-intuitive.

Well, actually the 8761st value (for the storage) already belongs to the next year. In fact, with the old definition of giving the initial value before the observation period, the values for the storage were either off by one or you'd have to use right-indexed. intervals. (Then, your indexes should range from goes from 1st of January 01:00 to 1st of next January 00:00 instead of 1st of January 00:00 to 31st of December 23:00.)

PR #812, opened 2021-12-14
Add inicativity_costs to NonConvex

p-snft:
The newly added `inactivity_costs` explicitly model potential costs of non-operation. This might be applicable if degradation is faster when not operating a device. Previously, this could only be expressed implicitly, by giving negative values for the complementary `acitivity_costs`.

PS: This feature was originally proposed at openmod to model [startup_costs dependent on time since last run](https://forum.openmod.org/t/startup-costs-dependent-on-time-since-last-run/2937). When the costs rise linearly, this directly translates to `inacitivity_costs`.

### DIFF
--- a/docs/whatsnew/v0-5-0.rst
+++ b/docs/whatsnew/v0-5-0.rst
@@ -15,6 +15,7 @@ API changes
 New features
 ############
 
+* Add `inactivity_costs` as an option for `NonConvexFlow`.
 
 Documentation
 #############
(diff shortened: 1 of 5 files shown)


p-snft:
I don't get the number of coveralls. Over there, it says > 95% coverage. So, I consider this finished.

jokochems:
This looks all reasonable. Calculation of `1 - self.status` provides the inactivity status. The remainder is the same as for the activity costs. Test and lp file look nice as well.

jokochems (commented):
All reasonable. Go ahead and merge it. :-)

joroeder:
Generelly, I think it would also make sense to provide a short hint in the documentation or an example. Thereby, new features become more visible.

Issue #759, opened 2021-05-07
Start values could be provided to the solver


jokochems:
At the user & dev meeting in Spring 2021 it was brought up that **providing the solver with start values** may increase computation speed. This is especially true if you have to resolve a complicated model with minor modification often times (like in a sensitivity analysis) and probably most relevant for nonconvex (MILP) models. 

It seemed that from the developer side, there is no urgent need to have that feature though there is some general interest.

jokochems:
Without having a deeper look into it, this should be possible by passing down values to the _initialize_ attribute of pyomo variables, see the [pyomo docs on Variables](https://pyomo.readthedocs.io/en/stable/pyomo_modeling_components/Variables.html). This should actually be not so hard to do.

Finding appropriate start values and mapping them to your components and objects could potentially be harder to achieve. In my personal opinion, it wouldn't make so much sense to provide this feature without the option to extract suitable start values.

simnh:
Yes pyomo supports this. Finding values will indeed by not easy. The easiest way would be to use values from solved  similar models, where the pyomo model instance with results values is used. For example, if only the objective function is replaced, this could be done building the new objective, keeping the rest of the (solved model). However, if new components enter or new constraints due to other paramerisation it might be difficult. 

The main problem I see is that there is no direct link between the values / parameters inside components and variables / parameters inside the model. Therefore, in any way it would include working on the pyomo-model instance. 

On the other hand: it should be easy add a attribute to all flow variables, i.e. such as  "value" , which is then used when the model is build. that can be done for the investment variables etc as well...

We actually do this already for the components [(here](https://github.com/oemof/oemof-solph/blob/da6fc612ef7791372d7c476b146e658da77942ff/src/oemof/solph/models.py#L336-L344), where we fix the values (e.g. for modelling must-run or renewables). 

Without the `.fix()` call, I think that would be used as start value (if gurobi / cplex is used and the warmstart agrument is used.

jokochems:
I totally agree with you, @simnh. That is exactly what I had in mind in terms of difficulties. You need a remapping back from your model results to the pyomo variables. This could also be used to speed up rolling horizon computations because you wouldn't have to rebuild your model in each iteration.

> We actually do this already for the components [(here](https://github.com/oemof/oemof-solph/blob/da6fc612ef7791372d7c476b146e658da77942ff/src/oemof/solph/models.py#L336-L344), where we fix the values (e.g. for modelling must-run or renewables).
> 
[...]

PR #814, opened 2021-12-17
Introduce parameter "expected" for Flows

p-snft:
This value can be used as a hint to the optimiser to speed up
minimisation. The fature is actually a reintroduction:
Up to oemof v0.3, it was possible to have a time series "actual_value"
plus a boolean parameter "fixed". When fixed was false, it was just
taken as the initial value for optimisation of the energy system.
When compared to this previous implemenataion, having either fix=(series)
or expected=(series) might be easier to understand.

- [ ] Test speed-up for complex model
- [ ] (More) documentation
- [ ] Unit tests

Implements #759


p-snft:
For testing the speedup, I consider the following:

- Source with unlimited max and cost.
- Transformer with random conversion_factors.
- GenericStorage with very high capacity and random series for loss_rate.
- Sink with random (fix) demand 

With a sufficiently high number of time steps, this should be rather hard to optimised. So:

1. Optimise without giving starting values.
2. Redo 1. to make sure there is no effect due to caching, branch prediction and stuff like that.
3. Optimise using solution of 1. as starting values.
4. Redo 1 once again, same reasoning as in 2, should again be slower.

src/oemof/solph/_models.py
```diff
@@ -362,6 +362,13 @@ def _add_parent_block_variables(self):
                     elif (o, i) in self.UNIDIRECTIONAL_FLOWS:
                         for t in self.TIMESTEPS:
                             self.flow[o, i, t].setlb(0)
+
+                    if self.flows[o, i].value[self.TIMESTEPS[1]] is not None:
+                        for t in self.TIMESTEPS:
+                            self.flow[o, i, t].value = (
+                                self.flows[o, i].value[t]
+                                * self.flows[o, i].nominal_value
+                            )
```
p-snft, line 371: This statement does not seem to have any effect. Also, I did not see any hint in the documentation that assignment is possible. (However, it seems to work for the fixed entries.)

p-snft: I found examples that show the value being set that way in the documentation (https://pyomo.readthedocs.io/en/stable/working_models.html#changing-the-model-or-data-and-re-solving). Pyomo values are fixed, so thzis method seems to work. However, in the LP files, there is no starting value. Also, there is no run time difference.

As the [documentation for Pyomo variables](https://pyomo.readthedocs.io/en/stable/pyomo_modeling_components/Variables.html) says that initialisation "is particularly important for non-linear models", I feel that the starting value might be simply ignored for linear variables.

p-snft:
I learned that a parameter `warmstart=True` has to be used when solving the Promo model so that the starting values are used.

PR #815, opened 2021-12-22
Rename summed max/min

### LINKED ISSUE #790: Rename the FLow attributes `summed_max` and `summed_min`
The `FLow` attributes `summed_max` and `summed_min` can be used to define the minimum or maximum amount of energy of a Flow summarised over the whole period. Due to the Investment Mode these attributes are relative numbers and will be multiplied with the `nominal_value` or the `investment_variable` to get the absolute bound. 

```
maximum_energy = summed_max * nominal_value
maximum_energy = summed_max * investment_variable
```

In most cases the `nominal_value` is something like the installed capacity, therefore `summed_max` is something like the maximum full load time and the unit is a time unit e.g. `nominal_value`: [W] and `summed_max`: [h] the resulting maximum energy bound will have the unit: [Wh].

Related to [this forum post](https://forum.openmod.org/t/can-the-demand-be-optimized-in-oemof-solph-sink/2814)  I got the impression that the name `summed_max` is misleading and should be renamed to something like `max_nominal_time`. 

I do not really like `max_nominal_time` though.
@oemof-developer do you have any ideas? What do you think about this issue?

uvchik:
Resolves #790 

Renaming of Flow attributes:
`summed_max` to `max_capacity_factor`
`summed_min` to `min_capacity_factor`

Everything seems to work. Names are always a matter of taste but if you have strong objections leave a comment. For me it is definitely clearer than before.

Shall we just break the API or shall we add a `FutureWarning`?


src/oemof/solph/flows/_flow.py
```diff
@@ -73,10 +73,10 @@ class Flow(on.Edge):
           * `'costs``: numeric (scalar or None), the gradient cost per
             unit.
 
-    summed_max : numeric, :math:`f_{sum,max}`
+    max_capacity_factor : numeric, :math:`f_{sum,max}`
```
jokochems, line 76: Shouldn't we rename the formula element as well, e.g. to `cf_{max}`?

jokochems:
First of all, I like the renaming. It makes it easier to understand, what the attribute is good for.

> Shall we just break the API or shall we add a `FutureWarning`?
I don't know. I would raise some user feedback. Maybe it can be done the same way as when the storages attributes were renamed.

src/oemof/solph/flows/_flow.py
uvchik: Thank you, @jokochems  I forgot it.

jnnr (commented):
Sorry, but you asked for my opinion: Names may be a matter of taste, but if a term is widely used and has a clear definition, we should better follow it.  

Capacity factor is a unitless quantity between 0 and 1, relating the actual energy over a period of time to the maximum possible. (https://en.wikipedia.org/wiki/Capacity_factor)

The quantitiy that we are renaming here has a unit of time and can be between 0 and the defined period.

I see two options: Changing the equations such that `max_capacity_factor` represents the correct thing:
```
summed_energy <= max_capacity_factor * nominal_value * period_length
```
 Or choose another name.

uvchik:
Yes, @jnnr I asked for you opinion and I am happy to get your feedback. Normally our problem is a lack of comments and not to many.

You are right the unit is a time unit (hour by default but it is possible to change it to e.g. minutes) and the capacity factor is dimensionless. I have overseen this.

Therefore your suggestion `fullloadtime_max` or `full_load_time_max` (respectively `min`) might be a better option.

src/oemof/solph/flows/_flow.py
```diff
@@ -136,12 +136,26 @@ def __init__(self, **kwargs):
         # E.g. create the variable in the energy system and populate with
         # information afterwards when creating objects.
 
+        # --- BEGIN: The following code can be removed for versions > v0.6 ---
```
p-snft, line 139:
```suggestion
        # --- BEGIN: The following code can be removed for versions >= v0.6 ---
```

src/oemof/solph/flows/_flow.py
```diff
@@ -405,37 +421,45 @@ def _create(self, group=None):
...
+            for inp, out in self.MIN_CAPACITY_FACTOR_FLOWS:
                 lhs = sum(
                     m.flow[inp, out, ts] * m.timeincrement[ts]
                     for ts in m.TIMESTEPS
                 )
                 rhs = (
-                    m.flows[inp, out].summed_min
+                    m.flows[inp, out].min_capacity_factor
                     * m.flows[inp, out].nominal_value
                 )
```
p-snft, line 454: These lines make perfectly clear, what we speak about. LHS is energy (time integral over power), RHS is full load "hours" times nominal power.

p-snft (changes requested):
As far as I see, `full_load_time`would be a great name. Please change accordingly and the PR is fine by me.

uvchik:
I replaced the following attributes:

`summed_max` to `full_load_time_max`
`summed_min` to `full_load_time_min`

I hope I did not forget any occurrences. 

It is still possible to use the old names until v0.6. It is planned to remove the old names than, so I added a `FutureWarning` for the versions >= 0.5 and <0.6.

p-snft (commented):
I just added the change to the WhatsNew. The PR should be fine now.

p-snft (approved):
I just added the change to the WhatsNew. The PR should be fine now.

uvchik:
@jnnr , @jokochems Are you fine with this PR, too?

src/oemof/solph/flows/_flow.py
```diff
@@ -73,10 +73,10 @@ class Flow(on.Edge):
           * `'costs``: numeric (scalar or None), the gradient cost per
             unit.
 
-    summed_max : numeric, :math:`f_{sum,max}`
+    full_load_time_max : numeric, :math:`t_{full\_load,max}`
         Specific maximum value summed over all timesteps. Will be multiplied
         with the nominal_value to get the absolute limit.
```
jnnr, line 78:
```suggestion
        Upper bound on the summed flow expressed as the equivalent time that the flow would have to run at full capacity to yield the same sum. Will be multiplied with the nominal_value to get the absolute limit.
```

jnnr (approved):
Looks good! Got one suggestion for the docs which might help to understand the name better. If you agreed with the suggestion, you would need to adapt it for the min, too.

src/oemof/solph/flows/_flow.py
```diff
@@ -270,10 +284,12 @@ class FlowBlock(SimpleBlock):
...
-    SUMMED_MAX_FLOWS
-        A set of flows with the attribute :attr:`summed_max` being not None.
-    SUMMED_MIN_FLOWS
-        A set of flows with the attribute :attr:`summed_min` being not None.
+    full_load_time_max_FLOWS
+        A set of flows with the attribute :attr:`full_load_time_max` being not
+        None.
+    full_load_time_min_FLOWS
+        A set of flows with the attribute :attr:`full_load_time_min` being not
+        None.
```
jokochems, line 292: I'd suggest capitalization for Sets

jokochems (approved):
Looks good besides this minor capitalization issue in the docstrings.

uvchik:
Thank you @p-snft, @jnnr, @jokochems for your improvements.

PR #819, opened 2022-02-08
Fix/flow silently not fixed

### LINKED ISSUE #818: Flow's argument"fix" is silently ignored when no nominal_value is given
**Describe the bug**
Flow's argument "fix" is silently ignored when no nominal_value is given.

**Current behaviour**
```Python
solph.flows.Flow(fix=0.5)
```
leads to a flow with a non-fixed value without any feedback given.

**Expected behaviour**
As there is no `nominal_value` given, the Flow cannot be fixed. Thus, an error should be thrown.

p-snft:
Fixes #818.

Todo:
* [x] Unit test that provokes the error.


p-snft:
The errors (deviations from Black style) are in code that I did not touch.

tests/test_solph_network_classes.py
```diff
@@ -92,14 +92,55 @@ def test_flow_with_fix_and_min_max():
...
+    ):
+        solph.flows.Flow(summed_min=0.3)
+
+
 def test_min_max_values_for_bidirectional_flow():
-    a = solph.flows.Flow(bidirectional=True)  # use default values
-    b = solph.flows.Flow(bidirectional=True, min=-0.9, max=0.9)
+    a = solph.flows.Flow(
+        bidirectional=True,
+        nominal_value=1,
```
uvchik, line 134: This nominal_value in this Flow is `None` and not `1`, so you do not have to set it, because `None` is the default value.

tests/test_solph_network_classes.py
```diff
@@ -92,14 +92,55 @@ def test_flow_with_fix_and_min_max():
         solph.flows.Flow(fix=[1, 3], max=[0, 5], min=[4, 9])
 
 
+def test_infinite_values():
+    msg1 = "nominal_value must be a finite value"
+    msg2 = "max must be a finite value"
+    with pytest.raises(AssertionError, match=msg1):
+        solph.flows.Flow(nominal_value=float("+inf"))
+    with pytest.raises(AssertionError, match=msg2):
+        solph.flows.Flow(max=float("+inf"))
```
uvchik, line 101: I do not get the idea of an `AssertationError`. Shouldn't it be a `ValueError`?

src/oemof/solph/flows/_flow.py
```diff
@@ -250,6 +255,29 @@ def __init__(self, **kwargs):
...
+                    if kwargs.get(attr) is not None:
+                        warn(
+                            warn_msg_no_nominal_value.format(attr),
+                            debugging.SuspiciousUsageWarning,
+                        )
+            else:
+                assert math.isfinite(
+                    self.nominal_value
+                ), infinite_error_msg.format("nominal_value")
+        assert math.isfinite(self.max[0]), infinite_error_msg.format("max")
```
uvchik, line 279: Why is only max checked for a finite value and not min/full_load_time_max....
Why do you provoke an `AssertationError`, for me it is a `ValueError`, but I am not sure.

src/oemof/solph/flows/_flow.py
```diff
@@ -250,6 +255,29 @@ def __init__(self, **kwargs):
...
+                "nominal_value must be set as well.\n"
+                "Otherwise, it won't have any effect."
+            )
+            if self.nominal_value is None:
+                for attr in need_nominal_value:
+                    if kwargs.get(attr) is not None:
+                        warn(
+                            warn_msg_no_nominal_value.format(attr),
+                            debugging.SuspiciousUsageWarning,
+                        )
```
uvchik, line 274: Why is it a Warning an not an Error. The value is multiplied with the nominal value and e.g. 5 * None provokes an error. So it should be an Error.

p-snft: It should only be a problem if the `nominal_value` is still not set when the model is created. To my understanding, it is theoretically possible to leave it unset in the constructor and to set it later.

jokochems: I'd agree to what @p-snft stated, but I have no strong opinion on that. I think since in 99% of all cases, the `nominal_value` is set rightaway, throwing an Error would be fine. If you still want to update your `nominal_value` later, you are free to do so. Throwing an Error should be less prone to user errors.

jokochems (commented):
I guess, we are better off with throwing an error.

jokochems:
Thank you,  @p-snft. I really like the improvement towards user-friendlyness.

src/oemof/solph/flows/_flow.py
p-snft: Agreed. Raising an error is the better option for most users.

src/oemof/solph/flows/_flow.py
```diff
@@ -172,6 +168,15 @@ def __init__(self, **kwargs):
...
         }
+        need_nominal_value = [
+            "fix",
+            "full_load_time_max",
+            "full_load_time_min",
+            "max",
+            "min",
+            "summed_max",
+            "summed_min",
+        ]
```
uvchik, line 179: You could add a hint here that `summed_min/max` is deprecated, as a reminder, but this is just an idea.

uvchik (approved):
I agree with @jokochems, this is a good improvement. Thanks you @p-snft.

jokochems (approved):
Thank you! It looks very clean. From my side, there is no contradiction against going ahead and merging it.

PR #820, opened 2022-02-09
Add description for logo use

fwitte:


p-snft (approved):
Black errors are in code that was not touched here. Approved.

PR #760, opened 2021-05-11
Exclude pyomo version that causes a too long computing time

uvchik:
Hotfix. Exclude pyomo version `5.7.1`.

PR #821, opened 2022-02-11
Make solph work with pyomo >= 0.6

uvchik:
At the moment the use of Pyomo >= 0.6 leads to very high computing times ( up to 100 times longer). @gnn found out, that this is caused by the root logger but only if the logging level is DEBUG. The problem is located in the Model class.

1. Therefore I changed the default logger level in oemof.tools to INFO, because the logger from oemof.tools is used in almost every solph example. In some examples the logger is explicitly set to DEBUG, so changing the default value will not help.
2. As it is also possible to set the logger in a different way, I added a function to check the level of the logger in the Model class. Now an exception will be raised if the root logger level is set to DEBUG (default. WARNING). This exception can be avoided if the model is initialised using `debug=True`: `model = solph.Model(energysystem, debug=True)`. This can be done with a small model if the debugging mode is needed. If it was not intended to use the debug-level the level can be changed:
```python
import logging
logging.getLogger().setLevel(logging.INFO)
```
This will cause long computing times or an exception :
```python
from oemof.tools import logger
logger.define_logging(
    screen_level=logging.DEBUG,
    file_level=logging.WARNING,
)
```
If the new logger of oemof.tools is released, using the default values will fix the problem
```python
logger.define_logging()
# or
logger.define_logging(
    screen_level=logging.INFO,
    file_level=logging.WARNING,
)
```

* Related issues?
#760 
#766

* Things to do:
  * [x] Fix tests
  * [ ] Adapt examples, if the DEBUG level is explicitly set
  * [x] Write tests for the Exception
  * [ ] Change docstring of oemof.tools logger to make clear, that the root logger is touched by the function
  * [ ] Make a release of oemof.tools to make it easy to set the dependency from oemof.solph

In the long run we should find out the reason of the problem, because forbidding the logging level is more a workaround than a real solution of the problem. But it is good that it makes it possible to use the actual version of Pyomo.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -80,7 +80,7 @@ setup(
         "dill",
         "numpy",
         "pandas",
-        "pyomo >= 5.7.0, < 7.0",
+        "pyomo >= 6.0.0, < 7.0",
         "networkx",
         (
             "oemof.tools@https://github.com/oemof/oemof.tools/archive/feature/"
--- a/setup.py
+++ b/setup.py
@@ -82,8 +82,10 @@ setup(
         "pandas",
         "pyomo >= 5.7.0, < 7.0",
         "networkx",
-        ("oemof.tools@https://github.com/oemof/oemof.tools/archive/feature/"
-         "set-minimal-lggin-level.zip"),
+        (
+            "oemof.tools@https://github.com/oemof/oemof.tools/archive/feature/"
+            "set-minimal-lggin-level.zip"
+        ),
         "oemof.network",
     ],
     extras_require={
--- a/setup.py
+++ b/setup.py
@@ -82,7 +82,8 @@ setup(
         "pandas",
         "pyomo >= 5.7.0, < 7.0",
         "networkx",
-        "oemof.tools",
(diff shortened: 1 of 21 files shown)


uvchik:
In my tests this works as fast as before.

@p-snft I am not sure if there is some more potential to speed up the process, but at least it not works with the actual version of Pyomo as good as before.

setup.py
```diff
@@ -80,9 +80,12 @@ def read(*names, **kwargs):
         "dill",
         "numpy",
         "pandas",
-        "pyomo >= 5.7.0, < 5.7.3",
+        "pyomo >= 5.7.0, < 7.0",
```
p-snft, line 83:
```suggestion
        "pyomo >= 6.0.0, < 7.0",
```

The changes require `pyomo>=6`, when using 5.7, the following error occurs:

> ImportError: cannot import name 'ScalarBlock' from 'pyomo.core.base.block'

p-snft:
Thanks for solving the issue. Speed is fine for me, again.

p-snft (approved):

PR #822, opened 2022-02-17
Rename Transformer to Converter

### LINKED ISSUE #728: Rename Transformer
When people hear "transformer", they think of electrical devices. However, the `GenericTransformer` is neither meant to be (only) electrical nor bidirectional (as electrical transformers typically are). Thus, I got the suggestion (thx to @psychon) to rename it to `GenericConverter`. The term is more general and does not imply that back-conversion is possible.

p-snft:
When people hear "transformer", they think of electrical devices.
However, the Transformer is neither meant to be (only) electrical
nor bidirectional (as electrical transformers typically are).
Thus, I got the suggestion to rename it to Converter.
The term is more general and does not imply that back-conversion
is possible.

Todo:
- [x] Actual renaming
- [x] Create transitional wrapper for soft deprecation
- [x] Change examples
- [x] Check and fix documentation
- [x] Explain in WhatsNew

Fix #728


Bachibouzouk:
Wouldn't it make sense to do this PR at last before releasing v0.5 ? It seems this can be done with ctrl+f (or Strg+f for german keyboards) quite quiclky

p-snft:
It's actually not that easy, as “transformer” can be the correct term in some cases (where it does not refer to the class). I merged the current status of dev to be able to make some progress here.

p-snft:
We agreed that this might be moved to v0.5.1, because a transitional phase makes sense anyway.

p-snft:
The _relation_rule of the `OffsetConverterBlock` is currently not built. I am aware of that and trying to fix that. (The orresponding lines are also deleted from the LP files to find other issues.)

p-snft (commented):
I checked all the Python files, again. Also, I went through most of the lp files even if it's redundant. I'd really like to have a second person to review this but as this is mostly renaming things, I do not want to wait longer before at least having a RC for v0.5.1. So, I am going to merge this now.

PR #823, opened 2022-02-24
Fixed check for callables in processing.parameter_as_dict()

### LINKED ISSUE #707: Postprocessing: Methods appears in output of get_parameter_as_dict()
**Describe the bug**

The output of `processing.get_parameter_as_dict()` contains methods from oemof.tabular. The methods are e.g. the `build_solph_components` of the oemof.tabular facades. This is not intended - not what one expects as a parameter.

**To Reproduce**
1. Create an EnergySystem using oemof.tabular.
2. Call get_parameter_as_dict(es).
3. See the function objects `build_solph_components` in the dictionary.

**Expected behavior**
The function objects should not be part of the dictionary.

 - OS: Xubuntu
 - Python version 3.6

henhuy:
When processing results, function `processing.parameter_as_dict()` should return scalars and sequences of given ES. Within the function is a check for methods (aka a check for `callable()`) which shall not be included in parameter dict. Nevertheless, the check was implemented incorrect, resulting in methods to go into parameter dict. This PR fixes that.

Fixes #707 

Error was related to check outcome of `dir(component)` for callables. But as `dir()` returns list-of-str and str is not callable, methods weren't filtered out.

### DIFF
--- a/src/oemof/solph/processing.py
+++ b/src/oemof/solph/processing.py
@@ -267,9 +267,7 @@ def __separate_attrs(system, get_flows=False, exclude_none=True):
             "registry",
             "inputs",
             "outputs",
-            "register",
             "Label",
-            "from_object",
             "input",
             "output",
             "constraint_group",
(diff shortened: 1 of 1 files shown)


jnnr (approved):
Did not have the time to test it. By inspection this seems ok.

PR #825, opened 2022-02-24
Add option `exclude_attrs` to processing.parameter_as_dict() 

### LINKED ISSUE #824: Add ability to exclude attributes from processing results
This issue is related to #707

Problem description:
When using oemof_tabulars facades energysystem components are generated with multiple attributes which are not related to oemof.solph. Using `processing.parameter_as_dict()` of oemof.solph results in many unwanted parameters from related component (i.e. `subnodes` as written in #707). It would be nice to have an option to exclude attributes from parameter dict generation.

Proposal:
Add attribute `exclude_attrs=None` to `parameter_as_dict()` which can hold a list of unwanted attributes, which are then filtered out before preparing parameter dict results. Default results are untouched as `exclude_attrs` is optional.

henhuy:
This PR adds an option to exclude attributes from `processing.parameter_as_dict()` scan in.

Closes #824

### DIFF
--- a/src/oemof/solph/processing.py
+++ b/src/oemof/solph/processing.py
@@ -369,7 +369,8 @@ def __separate_attrs(
             if len(value) == 0 or value[0] is None:
                 del com["sequences"][ckey]
 
-    exclude_attrs = exclude_attrs or []
+    if exclude_attrs is None:
+        exclude_attrs = []
 
     # Check if system is es or om:
     if system.__class__.__name__ == "EnergySystem":
@@ -411,7 +412,8 @@ def parameter_as_dict(system, exclude_none=True, exclude_attrs=None):
     dict: Parameters for all nodes and flows
     """
 
-    exclude_attrs = exclude_attrs or []
+    if exclude_attrs is None:
+        exclude_attrs = []
 
     flow_data = __separate_attrs(
         system, exclude_attrs, get_flows=True, exclude_none=exclude_none
--- a/src/oemof/solph/processing.py
+++ b/src/oemof/solph/processing.py
@@ -243,7 +243,7 @@ def meta_results(om, undefined=False):
 
 
 def __separate_attrs(
-    system, exclude_attrs=None, get_flows=False, exclude_none=True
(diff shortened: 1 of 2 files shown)


henhuy:
Test fails due to black fails in unrelated files... 
What to do about this?

jnnr (commented):
Thanks for this feature! Looks good. Have not tested it. The fails of black in unrelated files doesn't have to be solved here.

src/oemof/solph/processing.py
```diff
@@ -251,7 +253,15 @@ def __separate_attrs(system, get_flows=False, exclude_none=True):
     {(node1, node2): {'scalars': {'attr1': scalar, 'attr2': 'text'},
     'sequences': {'attr1': iterable, 'attr2': iterable}}}
 
-    om : A solved oemof.solph.Model.
+    system:
+        A solved oemof.solph.Model or oemof.solph.Energysystem
+    exclude_attrs: List[str]
+        Optional list of additional attributes which shall be excluded from
+        parameter dict
```
p-snft, line 260: The documentation states that the argument is optional, but as no default (`None`?) is given, it is obligatory.

henhuy: You are right. I moved initialization of `exclude_attrs` into function `parameter_as_dict` where it is marked correctly as optional and forgot about init in `separate_attrs`. Will add it there too. Thanks

src/oemof/solph/processing.py
```diff
@@ -369,6 +369,8 @@ def remove_nones(com):
             if len(value) == 0 or value[0] is None:
                 del com["sequences"][ckey]
 
+    exclude_attrs = exclude_attrs or []
```
p-snft, line 372: In my opinion, a classical
```suggestion
    if exclude_attrs is None:
        exclude_attrs = []
```
would be more comprehensible.

henhuy: okay for me - I like this style, but adapted your suggestion (for both places)

p-snft (approved):
Fine by me.

PR #826, opened 2022-03-01
Add nonconvex investment flow

sasa821:
### Explanation:
This new class represents a new `Flow` which combines both `Investment` and `NonConvex` classes and provides the possibility to perform the capacity optimization of an asset considering min/max loads (i.e., percentages of the capacity which is being optimized), as well as the status of the operation. 

It must be noted since in the `NonConvexInvestmentFlow` a binary variable (`status` of the flow at each time) is multiplied by a continuous variable, `invest`, which is the capacity of the asset being optimized, the problem becomes **non-linear**. For some solvers such as **gurobi**, this kind of non-linearity is automatically linearized (probably using the big M method), but in other free solvers such as **cbc**, this causes an error in the optimization. For this reason, those nonlinearities are addressed by introducing a new variable called `new_param` and adding three new constraints.

### Example:
The following diagrams show a good comparison between the `Investment` and the `NonConvexInvestment` flows. These diagrams are duration curves of a diesel genset optimized for a mini-grid with the same input for both optimizations. As seen in the figures, assuming that the diesel genset has a minimum load of 20%, the `Investment` flow would result in an infeasible operation of this device for more than 40% of its annual operation. However, this would never happen using the `NonConvexInvestment` flow because it considers the minimum load of the genset during the optimization. 

Moreover, the optimal capacity using the `NonConvexInvestment` flow is significantly different from the one obtained from the `Investment` flow.

investment
non_convex_invest


src/oemof/solph/flows/_non_convex_invest_flow.py
```diff
@@ -0,0 +1,917 @@
...
+        Value of the investment variable, i.e. equivalent to the nominal
+        value of the flows after optimization.
+
+    Status variable (binary) `om.NonConvexInvestmentFlowBlock.status`:
+        Variable indicating if flow is >= 0 indexed by FLOWS
+
+    * :math: `new_param(i,o,t)` (non-negative real number)
+        new paramater representing the multiplication of
+        `P_{invest}` and status(i,o,t)`used for the constraints
+        on the minimum and maximum flow constraints.
```
p-snft, line 180: What about `P_{status}`?

p-snft:
Hi @sasa821, thanks for your contribution. As far as I see, the code does what it is supposed to do. To be included in the release, unit tests should be included. (Note that the test fails that checks for the error message when combining `NonConvex` and `Investment`, you already added a TODO note for that.)

In my opinion, the complexity (cf. Codacy check) is okay. Also, you introduce some code duplication. Both is supposed to be cleaned up later. (We currently have more duplication then necessary, it should not be hard to address the newly induced duplication when we fix the existing one.)

joroeder:
How about naming it ``NonlinearInvestmentFlow``? As it is already possible to make the InvestmentFlow nonconvex for applying investment thresholds, which is something different than here.

src/oemof/solph/flows/_non_convex_invest_flow.py
sasa821: > What about `P_{status}`?

I think `P_{status}` is NOT required here or anywhere else in this flow class. As long as the `P_{invest}` is larger than zero, `P_{status}` is always `True`, and when `P_{invest}` becomes zero, it means `P_{status}` would also be equal to zero. On the other side, by removing this integer variable from the problem, the computation speed will be faster while results won't change.

sasa821:
> Hi @sasa821, thanks for your contribution. As far as I see, the code does what it is supposed to do. To be included in the release, unit tests should be included. (Note that the test fails that checks for the error message when combining `NonConvex` and `Investment`, you already added a TODO note for that.)
> 
> In my opinion, the complexity (cf. Codacy check) is okay. Also, you introduce some code duplication. Both is supposed to be cleaned up later. (We currently have more duplication then necessary, it should not be hard to address the newly induced duplication when we fix the existing one.)

Hi @p-snft, thank you very much for your review. Could you assist me more with the inclusion of unit tests, please?

sasa821:
> How about naming it `NonlinearInvestmentFlow`? As it is already possible to make the InvestmentFlow nonconvex for applying investment thresholds, which is something different than here.

Thanks, @joroeder, for your comment. I used the name `NonConvexInvestment` because this new class enables the combination of `NonConvex` and `Investment` classes, which was not allowed so far in oemof. I understand your point, but I am not sure about `NonlinearInvestmentFlow`. There is non-linearity in this class, but it is caused by the multiplication of a binary variable and a continuous one. I believe the name `NonlinearInvestmentFlow` is not very descriptive.

src/oemof/solph/flows/_non_convex_invest_flow.py
p-snft: I wanted to suggest to (re)name the parameter `new_param` (to) `P_{status}`. As it is the product of the nominal power and the status, the name would make sense, imho.

p-snft:
> Hi @p-snft, thank you very much for your review. Could you assist me more with the inclusion of unit tests, please?

I'd be glad to help you. However, I am currently lacking time, so it will take some weeks until I can do so.

p-snft:
@sasa821: I'd have time now. Thus, I read through the full PR as it is right now. As using this module is restricted to special solvers, testing will be difficult. Also, I am considering if we should add a warning to the init function, so that nobody uses the feature with an unsupported solver.

Bachibouzouk:
> @sasa821: I'd have time now. Thus, I read through the full PR as it is right now. As using this module is restricted to special solvers, testing will be difficult. Also, I am considering if we should add a warning to the init function, so that nobody uses the feature with an unsupported solver.

@p-snft - I just tested it with cbc and it works, I am not sure I understand your point here?

sasa821:
>

p-snft:
I just opened a PR to your branch (@sasa821 ) with two examples for combinations of Invest/NonConvex. I did this just to have something at hand to test if the class works technically. (It does, thank you!) It would be nice if you could provide an example that reproduces the two plots you show in the first post of this PR.

After that, the next step would be adding the tests. Typically, the examples are a good starting points to create tests. This is why I contributed one for `NonConvexInvestFlow`.

Bachibouzouk:
> After that, the next step would be adding the tests. Typically, the examples are a good starting points to create tests. This is why I contributed one for `NonConvexInvestFlow`.

Thanks @p-snft - So examples are providable within oemof-solph package? I thought one had to store them in the oemof-examples repository (this is the reason why I told @sasa821 we should not commit his example file). Can one add tests where a simulation is run and we look at its output? We already added/modified a few tests in this PR, I am not sure if there is a "good" number of tests to be added ?

p-snft:
We used to have them separately. But as there are severe drawbacks (i.e. results are not tested automatically), this is about to change with v0.5.

When it comes to tests, of course the examples will be checked. This, however, is rather an integration test than a unit test. Thus, we need to add tests for individual functionality. In particular, there should be a constraint test that checks the generated LP file.

Bachibouzouk:
@p-snft - @sasa821 and I plan to wrap up this PR this coming wednesday would you be able/available to review it then? If not could you suggest another core dev of oemof?

p-snft:
I'd gladly do that. I can also offer to not consider problem with docs for solph as a whole (in the sense of https://github.com/oemof/oemof-solph/pull/826#pullrequestreview-1003970572) as blockers. Instead, we will rewrite those later.

PS: To be named when the code is cited in scientific publications, you should add your names to https://github.com/oemof/oemof-solph/blob/dev/CITATION.cff. (It was not mentioned in the contribution guidelines, I just fixed that.)

tests/test_scripts/test_solph/test_nonconvex_investment/test_nonconvex_investment.py
```diff
@@ -0,0 +1,328 @@
+# -*- coding: utf-8 -*-
+
+"""
+This test contains a NonConvexInvestFlow class.
+
+
+This file is part of project oemof (github.com/oemof/oemof). It's copyrighted
+by the contributors recorded in the version control history of the file,
+available from its original location
+oemof/tests/test_scripts/test_solph/test_nonconvex_investment/test_nonconvex_investment.py
```
p-snft, line 10: We moved away from this to explicitly naming authors in the "SPDX-FileCopyrightText".

src/oemof/solph/_exceptions.py
```diff
@@ -0,0 +1,6 @@
+class FlowOptionWarning(UserWarning):
+    pass
+
+
+class WrongOptionCombinationError(ValueError):
+    pass
```
p-snft, line 6: We have the other warnings in the tools package, to allow reusing them. It might make sense to do this for these exceptions as well.

(Just as a point for discussion, not a reason to reject this PR.)

tests/test_scripts/test_solph/test_nonconvex_investment/test_nonconvex_investment.py
Bachibouzouk: Do you have an example of the actual practice ?

src/oemof/solph/_exceptions.py
Bachibouzouk: I only saw one other warning `SuspiciousUsageWarning` in `oemof.tools`. We could define them there if this is needed. I personally find a bit cumbersome to have to create 2 PRs in 2 different packages, but I will abide :)

src/oemof/solph/flows/_non_convex_invest_flow.py
```diff
@@ -0,0 +1,860 @@
...
+        activity_costs=None,
+        inactivity_costs=None,
+        minimum_uptime=None,
+        minimum_downtime=None,
+        maximum_startups=None,
+        maximum_shutdowns=None,
+        initial_status=0,
+        positive_gradient=None,
+        negative_gradient=None,
+        **kwargs,
```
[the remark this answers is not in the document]
Bachibouzouk: As `NonConvexInvestmentFlow` inherits from `Flow` and `Flow` itself uses `**kwargs`, we then might as well define explicitely the argument in `Flow`, no? Couldn't this be tackled in a separate PR?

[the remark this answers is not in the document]
Bachibouzouk: I understand the benefit of explicit by getting rid of `**kwargs`, however wouldn't it make maintenance of inherited classes more cumbersome ? Moreover if typos are caught, then errors in the default values of the arguments in a child class are not caught (a solution for that would be to only pass the argument to the parent class if they don't have the default value of the child class, that way only the default value of the parent class rules.)

src/oemof/solph/flows/_non_convex_invest_flow.py
```diff
@@ -0,0 +1,860 @@
...
+        if self.investment is not None:
+            self.investment.nonconvex = True
+            self.investment._check_invest_attributes_maximum()
+            self.investment._check_invest_attributes()
+
+
+class NonConvexInvestFlowBlock(SimpleBlock):
+    r"""
+    **The following sets are created:** (-> see basic sets at
+    :class:`.Model` )
```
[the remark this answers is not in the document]
p-snft: I think some hints like "Same as in NonConvexFlow" would be nice to have, yes. In particular, when comparing the four flow classes, one shuld not have to open the doc pages in parallel.

tests/constraint_tests.py
```diff
@@ -178,6 +178,30 @@ def test_linear_transformer_invest(self):
...
+
+        solph.components.Transformer(
+            label="transformer_nonconvex_invest",
+            inputs={bfuel: solph.flows.Flow()},
+            outputs={
+                bel: solph.flows.NonConvexInvestFlow(
+                    nominal_value=None,
+                    variable_costs=25,
+                    min=0.2,
+                    max=0.8,
```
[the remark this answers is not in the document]
p-snft: Technically speaking, it is possible to have integer numbers >= 1 also for min/max. But i.e. 0.25 should also work. (It's just a hint. I thought it's nice to know. We also have numbers with ugly binary representation is the other tests. I just try to avoid this whenever I am aware that this might be a thing.)

src/oemof/solph/flows/_non_convex_invest_flow.py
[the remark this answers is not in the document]
p-snft: You are right, this can be done separately, as the Flow is not updated yet.

(Maybe, this PR is not the right place to discuss about the pros and cons of **kwags. We came to the conclusion to user friendlyness is the stronger argument in the current state of the package. So, all expected arguments are now made explicit. If you want additional arguments, you can still use them, but you have to hand them as a dict to the argument "options".)

tests/test_scripts/test_solph/test_nonconvex_investment/test_nonconvex_investment.py
p-snft: E.g. in https://github.com/oemof/oemof-solph/blob/daebfa4e2a75e80fb244f7b58a07c95bfb18d626/src/oemof/solph/_groupings.py#L15-L22

Bachibouzouk: Thanks, we will implement it then

tests/constraint_tests.py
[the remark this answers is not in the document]
Bachibouzouk: Ok then we will use 0.25 and 0.5 :)

src/oemof/solph/flows/_flow.py
```diff
@@ -231,10 +254,24 @@ def __init__(self, **kwargs):
...
+        # Checking for misuse of nonconvex and investment options of
+        # <class 'solph.flows.Flow'> instead of using
+        # <class 'solph.flows.NonConvexInvestFlow'>
+        if self.allow_nonconvex_investment is False:
+            if self.investment and self.nonconvex:
+                raise WrongOptionCombinationError(
+                    "Investment flows cannot be combined with "
+                    + "nonconvex flows using the class "
+                    + "<class 'solph.flows.Flow'>! Please consider using"
+                    + " <class 'solph.flows.NonConvexInvestFlow'>"
```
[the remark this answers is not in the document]
sasa821: > Is this syntax to highlight class names in error messages common? I'm not aware that we use it until now.

In the `test_storage_without_inputs` and `test_storage_without_outputs` tests the class names are highlighted in the warning message. @p-snft: Do you want us to keep the syntax as we already did or to remove it?

examples/nonconvex_invest_flow_examples/house_with_nonconvex_investment.py
```diff
@@ -0,0 +1,149 @@
...
+    - fireplace: wood firing, has a minimum heat and
+                 will burn for a minimum time if lit.
+    - boiler: gas firing, more flexible but still
+              with minimal load and also with
+              higher cost than wood firing
+
+
+Installation requirements
+-------------------------
+This example requires the version v0.5.x of oemof. Install by:
```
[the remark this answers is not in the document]
Bachibouzouk: Done in [3f55df9](https://github.com/oemof/oemof-solph/pull/826/commits/3f55df92d7b867dca3e9b72d0814a4708d35246a)

src/oemof/solph/flows/_flow.py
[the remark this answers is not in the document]
Bachibouzouk: Done in [8f7f537](https://github.com/oemof/oemof-solph/pull/826/commits/8f7f537ca183d090f48332314a84b272a6091393)

src/oemof/solph/flows/_non_convex_invest_flow.py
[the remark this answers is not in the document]
Bachibouzouk: done in [a0cfc72](https://github.com/oemof/oemof-solph/pull/826/commits/a0cfc724ce26726adadf24c63a1685ee2a8f53d3)

tests/lp_files/flow_nonconvex_invest_bounded_transformer.lp
```diff
@@ -0,0 +1,161 @@
...
+-1 flow(transformer_nonconvex_invest_electricityBus_0)
+= 0
+
+c_e_TransformerBlock_relation(transformer_nonconvex_invest_fuelBus_electricityBus_1)_:
++0.33000000000000002 flow(fuelBus_transformer_nonconvex_invest_1)
+-1 flow(transformer_nonconvex_invest_electricityBus_1)
+= 0
+
+c_e_TransformerBlock_relation(transformer_nonconvex_invest_fuelBus_electricityBus_2)_:
++0.33000000000000002 flow(fuelBus_transformer_nonconvex_invest_2)
```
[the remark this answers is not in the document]
Bachibouzouk: Done in [bba5854](https://github.com/oemof/oemof-solph/pull/826/commits/bba5854f1043305b860e04809d16467e3efd64ce)

tests/lp_files/flow_nonconvex_invest_bounded_transformer.lp
```diff
@@ -0,0 +1,161 @@
...
++1 NonConvexInvestFlowBlock_invest(transformer_nonconvex_invest_electricityBus)
+>= 0
+
+c_u_NonConvexInvestFlowBlock_maximum_rule(transformer_nonconvex_invest_electricityBus)_:
++1 NonConvexInvestFlowBlock_invest(transformer_nonconvex_invest_electricityBus)
+-1234 NonConvexInvestFlowBlock_invest_status(transformer_nonconvex_invest_electricityBus)
+<= 0
+
+c_u_NonConvexInvestFlowBlock_min(transformer_nonconvex_invest_electricityBus_0)_:
++0.20000000000000001 NonConvexInvestFlowBlock_invest_non_convex(transformer_nonconvex_invest_electricityBus_0)
```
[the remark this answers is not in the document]
Bachibouzouk: Done in [bba5854](https://github.com/oemof/oemof-solph/pull/826/commits/bba5854f1043305b860e04809d16467e3efd64ce)

tests/test_scripts/test_solph/test_nonconvex_investment/test_nonconvex_investment.py
Bachibouzouk: Done in [927eb02](https://github.com/oemof/oemof-solph/pull/826/commits/927eb022490b37f6e757fb75b5ac1e37a4b73fcf)

Bachibouzouk:
@uvchik - I just saw the PR close automatically because the oemof:v0.5 branch was deleted, is it then moved to another branch ? What should we then do with the non_convex branch which we wanted to merge into oemof-solph?

p-snft:
Changing the target branch is only allowed for @sasa821, I guess.

(I just reopened the v0.5 branch to be able to reopen this PR.)

Bachibouzouk:
Thanks @p-snft, now the target branch is correct, should someone else review this PR? Or merge it?

p-snft:
I asked for a second reveiw in the dev chat. For non-trivial stuff we want to have more then one review if possible.

src/oemof/solph/flows/_non_convex_invest_flow.py
```diff
@@ -0,0 +1,849 @@
...
+        for wm in other_warning_messages:
+            warnings.warn(wm)
+
+        # If an <class 'solph.Investment'> instance is used then its
+        # 'nonconvex' attribute should be set to True and the checks
+        # should be run again
+        if self.investment is not None:
+            self.investment.nonconvex = True
+            self.investment._check_invest_attributes_maximum()
+            self.investment._check_invest_attributes()
```
joroeder, line 144: I do not understand why these checks are repeated and the `self.investment.nonconvex` is set True. This attribute was intended to add a investment with fixed and capacity dependent investment costs (see [https://oemof-solph.readthedocs.io/en/latest/reference/oemof.solph.html#oemof.solph.flows._investment_flow.InvestmentFlowBlock](https://oemof-solph.readthedocs.io/en/latest/reference/oemof.solph.html#oemof.solph.flows._investment_flow.InvestmentFlowBlock))

Thus, I would rather recommend to remove this lines:

```suggestion
```

You could put a warning if either `investment.nonconvex` is True or `investment.offset` is given.

Actually, an investment attriubte is required for the NonConvexInvestFlow.

But maybe it makes sense to resolve the `Investment` class in the flow classes -- wasn't this decided on some dev meeting? Then, all this kind of issues become cleaner.

joroeder: In the end, this kind of combination of a nonconvex Investment (having a minimum investment threshold with an fixed offset value) and your NonConvexInvestFlow is not implemented, and the user should know about that.

joroeder (changes requested):
First of all, I think that is really nice contribution. Thanks to all people that were involved.

I would appreciate a feedback on the comments in the files.

General remarks (that should not hinder the merge):
 * I am not that deep into grouping, so I cannot say if everything is fine there.
 * Documentation should undergo an overall revision anyway, but this is a separate issue -- so, very nice contribution!
 * Local build of the API documentation did not work - is it a problem of this PR or with the dev branch? (or at my machine?)
 * Do we need the __init__.py in the examples? The others do not have an init either, so remove it?
 * The repository will become very large if we include many examples with a dataset of 8760 time steps -- Is this wanted?

src/oemof/solph/flows/_non_convex_invest_flow.py
Bachibouzouk: This was to be able to raise an AttributeError if the `solph.Investment` instance passed as init argument of `solph.flows.NonConvexInvestFlow` does not have `maximum` provided, the test test_nonconvex_investment_without_maximum_raises_warning was written to make sure an error is triggered. Said otherwise there would be nothing which prevents the user to provide a convex `solph.Investment`. This would not make sense, so we force the `nonconvex` of the `solph.Investment` instance to `True` and rerun the checks that would have only run if the `solph.Investment` instance would have had `nonconvex` set to `False`. The test [test_nonconvex_investment_without_maximum_raises_warning](https://github.com/sasa821/oemof-solph/blob/cc747bd6a770b23020b795374762e96e89417db1/tests/test_warnings.py#L157) we added would fail if you comment those lines.

tests/test_scripts/test_solph/test_nonconvex_investment/diesel_genset_data.csv
```diff
@@ -0,0 +1,8761 @@
+Demand,SolarGen
```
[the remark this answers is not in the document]
Bachibouzouk: I agree that we need to trim it to the minimum time required to be able to see the difference with the non_convex without invest problem

Bachibouzouk:
> * Do we need the **init**.py in the examples? The others do not have an init either, so remove it?

This was probably left from a failed attempt to reuse the example in the tests (so one would need to be able to import the example)

src/oemof/solph/flows/_non_convex_invest_flow.py
joroeder: ah okay, I see the point, the `maximum` is required for the NonConvexInvestFlow and you want to use the check. However, in my opinion, there should be some adaptions, the usage of the Investment attributes should be consistent with the nonconvex option of the InvestmentFlow, and therefore (maybe until a better solution is found), the checks should be adapted specifically to the NonConvexInvestFlow:
 * the `Investment` class/attribute needs to be there at all (otherwise, the NonConvexInvestFlow does not work)
 * `ep_costs` needs to be given
 * `maximum` needs to be given
 * `offset` must not be given (as it does not have an effect)
 * `existing` must not be given (as it does not have an effect)
 * `nonconvex` must be `False` (otherwise, the meaning of the attributes of the Investment class is not consistent)
 * `minimum` is optional (and sets a "hard" minimum which means that the investment cannot become zero)

Bachibouzouk: I agree with you, this would actually be easy to implement once the `get rid of "**kwargs"` todo has been implemented (we agreed above to tackle that in a subsequent PR), I suggest we add your comment as a TODO inside the code

PR #827, opened 2022-03-04
Convert Link assertion into SuspiciousUsageWarning

jnnr:
The __init__ of the Link component asserts that 2 inputs, outputs and conversion factors are passed. However, some users might want to initialize without passing those, setting them later. This is what oemof.tabular tries to do - and fails. The PR aims to fix that by converting the assertion into a SuspiciousUsageWarning which can be silenced.

Related: https://github.com/oemof/oemof-tabular/pull/21#issuecomment-1058331661

### DIFF
--- a/src/oemof/solph/components/experimental/_link.py
+++ b/src/oemof/solph/components/experimental/_link.py
@@ -92,7 +92,11 @@ class Link(on.Transformer):
             + "disable the SuspiciousUsageWarning globally."
         )
 
-        if len(self.inputs) != 2 or len(self.outputs) != 2 or len(self.conversion_factors) != 2:
+        if (
+            len(self.inputs) != 2
+            or len(self.outputs) != 2
+            or len(self.conversion_factors) != 2
+        ):
             warn(msg, debugging.SuspiciousUsageWarning)
 
     def constraint_group(self):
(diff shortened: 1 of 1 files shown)


src/oemof/solph/components/experimental/_link.py
```diff
@@ -81,14 +83,16 @@ def __init__(self, *args, **kwargs):
...
-        wrong_args_message = (
-            "Component `Link` must have exactly"
-            + "2 inputs, 2 outputs, and 2"
-            + "conversion factors connecting these."
+        msg = (
+            "Component `Link` should have exactly "
+            + "2 inputs, 2 outputs, and 2 "
+            + "conversion factors connecting these. "
+            + "If this is intended and you know what you are doing you can "
+            + "disable the SuspiciousUsageWarning globally."
```
p-snft, line 91: The message does not state that the constructor is called deviating from this specification.

jnnr: Thanks for your review! Hope [4f058e4](https://github.com/oemof/oemof-solph/pull/827/commits/4f058e49d8ec33173b5c097d9507de0e7f68cdcc) helps.

jnnr:
Do you have other remarks or recommendations, @p-snft?

p-snft (approved):

PR #834, opened 2022-03-25
Refactored assertion error in Link component into suspicious warning

henhuy:
This PR aims to merge PR #827, which was merged into v0.5, into dev as well.
This PR only turns errors into a warning - should be fine to merge.

### DIFF
--- a/src/oemof/solph/custom/link.py
+++ b/src/oemof/solph/custom/link.py
@@ -16,7 +16,9 @@ SPDX-FileCopyrightText: jnnr
 SPDX-License-Identifier: MIT
 
 """
+from warnings import warn
 
+from oemof.tools import debugging
 from oemof.network import network as on
 from pyomo.core import Binary
 from pyomo.core import Set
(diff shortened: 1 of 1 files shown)


henhuy:
As @jnnr mentioned in chat, we want to make this PR availalbe in dev branch. Should have no conflicts.
@p-snft PR #827 holding the same code changes is already merged to v0.5 - ths I think there is no need to merge dev into v0.5 afterwards (also as structure has changed in v0.5 and `link.py` has been refactored/moved).

p-snft:
> As @jnnr mentioned in chat, we want to make this PR availalbe in dev branch. Should have no conflicts. @p-snft PR #827 holding the same code changes is already merged to v0.5 - ths I think there is no need to merge dev into v0.5 afterwards (also as structure has changed in v0.5 and `link.py` has been refactored/moved).

Maybe, this would be the point in time where a v0.4 branch would be useful. At some point, the branches dev and v0.5 will have to be merged. This v0.4 branch would be the location for continued maintenance and backporting of fixes from v0.5.

henhuy:
So you suggest renaming this PR into v0.4?
Or shall I branch a v0.4 from dev and merge this PR into it?

p-snft:
The clean way would be to form `v0.4` from the current `dev` and to merge this PR into the new branch. As, however, there are no (other) changes to `dev`, the result is the same when you just rename `fix/link_inputs` to `v0.4`. (You will just not receive formal reviews as there is no PR.)

PS: Pleas do not branch from the tagged versions. That would just create extra work.

PR #836, opened 2022-04-05
Features/network 0.5

p-snft:
Follows changes between oemof.network v0.4 and v0.5:
* [x] https://github.com/oemof/oemof-network/issues/3, https://github.com/oemof/oemof-network/pull/14
* [x] https://github.com/oemof/oemof-network/issues/16 (marked as deprecated for a long time)

The second point has been implemented in #867. I merged this branch here to be able to test the combination of solph v0.5 and network v0.5.

PS: This PR is against feature/restructure_flow, as I included all changes from that branch (cf. #867). It would be cleaner to merge #856 first and merge this into dev afterwards.


p-snft:
I decided to already merge the changes from #877 to be able to solve merge conflicts.

setup.py
```diff
@@ -78,12 +78,12 @@ def read(*names, **kwargs):
     install_requires=[
         "blinker",
         "dill",
-        "numpy",
+        "numpy < 1.24",
```
jokochems, line 81: You are working on making it support never numpy versions, right?

jokochems (approved):
If I see it correctly, the main change this brings along is that it removes `oemof.network.network.Node.registry`, thus requires you to explicitly call `energysystem.add(component)` for all your components which I find a good thing as it is sort of a control routine for yourself. I did not do a very thorough review of the remainder since most of it is contained in my review on #877. Thank you for this feature.

setup.py
p-snft: In my opinion, this should be done before the release of v0.5. However, as the problem was just discovered here, it should be fixed in another PR.

Issue #837, opened 2022-04-11
Add option to prevent direction constraints in Link component


henhuy:
There has been an update in the LinkBlock component, which adds directional flow constraints.
Unfortunately, this feature breaks Link creation in oemof_tabular. 
I suggest to add an option `limit_direction` (which defaults to `True`) in Link component, by which direction constraints can be disabled.

p-snft:
As the component is experimental anyway, I don't see any reason not to change it quickly. PR is merged, I will close this. Thanks!

PR #838, opened 2022-04-11
Added option to (un-)limit directions in Link component

henhuy:
Related issue: #837 

This PR adds an option `limit_direction` to Link component in order to suppress creation of flow constraints regarding link direction.
As option defaults to `True`, this should not effect current setup, but only enables other devs to disable mentioned feature.


p-snft (approved):
Reads reasonable. (As the component is experimental, I will not start a discussion to improve the naming.)

PR #839, opened 2022-04-26
Require oemof.network v0.4

p-snft:
Set stable release of oemof.network as requirement. As the stable release emits a FutureWarning when the deprecated Node/Entity.registry is used, the warning has to be silenced when testing for other warnings.

Note that out unit tests use the registry, but other software does probably not.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -86,7 +86,7 @@ setup(
         "pyomo >= 5.7.0, < 5.7.3",
         "networkx",
         "oemof.tools",
-        "oemof.network",
+        "oemof.network >= 0.4.0, < 0.5.0",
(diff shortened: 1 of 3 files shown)


uvchik:
Why is it necessary to restrict oemof.network to versions < 0.5?

p-snft:
I'd suggest to remove the auto-registry in a second step. First, it's important to make sure solph will work when network is developed further. There is no known incompatibility, but v0.5 will break the API by convention. (Increasing the limit if there are just irrelevant changes to the API will be a quick fix.)

uvchik (approved):

PR #842, opened 2022-05-06
Add dual example and remove warnings

uvchik:
Using the dual variable produces two warnings that make people unsure if everything works fine.

This PR will remove these two warnings and another deprecated warning. Furthermore, an example is added two show how to add dual variables, how to fetch them from the results dictionary and how to interpret them.


examples/dual_variable_example/dual_variable_example.py
```diff
@@ -0,0 +1,287 @@
...
+# ********** PART 1 - Define and optimise the energy system ******************
+# ****************************************************************************
+
+###############################################################################
+# imports
+###############################################################################
+
+import logging
+import os
+import pprint as pp
```
p-snft, line 38:
```suggestion
import pprint as pp
```
All three imports are unused. However, pprint might be used if matplotlib is really optional. (Maybe, it shouldn't be optional. The plot seems to be essential.)

examples/dual_variable_example/dual_variable_example.py
```diff
@@ -0,0 +1,287 @@
+# -*- coding: utf-8 -*-
+
+"""
+General description
+-------------------
+
+A basic example to show how to get the dual variables from the system. Try
+to understand the plot.
```
p-snft, line 8: It would be nice to have an explanation for the plot. At the moment, it's really hard to understand.

p-snft (changes requested):

examples/dual_variable_example/dual_variable_example.py
```diff
@@ -0,0 +1,287 @@
...
+solver_verbose = False  # show/hide solver output
+
+# initiate the logger (see the API docs for more information)
+logger.define_logging()
+
+date_time_index = pd.date_range(
+    "1/1/2012", periods=number_of_time_steps, freq="H"
+)
+
+energysystem = EnergySystem(timeindex=date_time_index)
```
p-snft, line 64:
```suggestion
energysystem = EnergySystem(timeindex=date_time_index, infer_last_interval=True)
```
We should make sure the new default behaviour is adhered to.

examples/dual_variable_example/dual_variable_example.py
```diff
@@ -0,0 +1,287 @@
...
+        label="rgas",
+        outputs={bus_gas: flows.Flow(variable_costs=38)},
+    )
+)
+
+# create fixed source object representing pv power plants
+energysystem.add(
+    cmp.Source(
+        label="pv",
+        outputs={bus_elec: flows.Flow(fix=pv, nominal_value=700)},
```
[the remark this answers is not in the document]
uvchik: This is a matter of taste. I prefer to use fix+excess because I will find the excess in the excess variable. But it is also possible to subtract the resulting `PV_out` variable from the given max value to get the excess.

examples/dual_variable_example/dual_variable_example.py
uvchik: It was my plan to make matplotlib mandatory. With #850 it will be possible to install `pip install oemof.solph[examples]` which will install matplotlib. So I added this already to the header.

examples/dual_variable_example/dual_variable_example.py
```diff
@@ -0,0 +1,287 @@
...
+)
+
+my_flows = pd.concat(
+    [flows_to_bus, flows_from_bus, storage, duals],
+    keys=["to_bus", "from_bus", "content", "duals"],
+    axis=1,
+)
+
+my_flows.plot()
+plt.show()
```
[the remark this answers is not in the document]
uvchik: As matplotlib is a common library I will keep things simple.

examples/dual_variable_example/dual_variable_example.py
```diff
@@ -0,0 +1,287 @@
...
+)
+
+energysystem.add(storage)
+
+##########################################################################
+# Optimise the energy system and plot the results
+##########################################################################
+
+# initialise the operational model
+model = Model(energysystem)
```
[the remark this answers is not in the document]
uvchik: As this example is meant for v0.5.x we do not need this.

examples/dual_variable_example/dual_variable_example.py
uvchik: For me it is not to complicated. It is always good to have more docs but I think it is pretty self explaining. But for sure you have to understand how dual variables work, but this is not our task.

examples/dual_variable_example/dual_variable_example.py
uvchik: :+1:

examples/dual_variable_example/dual_variable_example.py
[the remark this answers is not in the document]
p-snft: This is why I put it as a question. You don't need to change this.

examples/dual_variable_example/dual_variable_example.py
```diff
@@ -0,0 +1,280 @@
...
+
+from oemof.solph import EnergySystem
+from oemof.solph import Model
+from oemof.solph import buses
+from oemof.solph import components as cmp
+from oemof.solph import flows
+from oemof.solph import processing
+
+solver = "cbc"  # 'glpk', 'gurobi',....
+debug = False  # Set number_of_timesteps to 3 to get a readable lp-file.
```
p-snft, line 44:
```suggestion
```
Unused, should be deleted.

examples/dual_variable_example/dual_variable_example.py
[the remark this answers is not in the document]
uvchik: I just published a new release of oemof.tools. You can just update oemof.tools and everything should work as expected.

uvchik:
Thank you @p-snft for your extensive review. I think everything should be fine now.

p-snft (approved):
With the release of oemof.tools, it now works.

PR #844, opened 2022-05-16
Fix attribute that has been renamed with Pyomo 6.4.1

uvchik:
With the try/except statement it will also work for older versions of Pyomo.


uvchik:
Sorry, I forgot that this fix is not necessary as the dev branch is restricted to Pyomo < 0.6.0

PR #845, opened 2022-05-16
Revise docstring of flow class

uvchik:
Related to #772.

Docstring for the Flow class (without InvestmentFlow, NonConvexFlow)

### DIFF
--- a/docs/reference/oemof.solph.rst
+++ b/docs/reference/oemof.solph.rst
@@ -25,25 +25,20 @@ oemof.solph.buses.Bus
     :undoc-members:
     :show-inheritance:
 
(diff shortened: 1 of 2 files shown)


p-snft (approved):
I wonder, if we should define a global list of symbols somewhere. In particular, it would be helpful if one could directly see the physical meaning of a symbol.

Here, I stumbled over `f_{min}` and `F_{min}`. The latter is a time, and the choice of symbols implies that it is the integral of the other. This is not correct but pretty close. Also, we might want to use `\times` instead of `\cdot`.

However, this is probably a topic to be addressed later, in a separate PR.

PR #848, opened 2022-05-19
[docs]: Add info on available tox environments

fwitte:
Basically as said in the title.


p-snft (approved):
I think Coveralls is wrong (again).

PR #849, opened 2022-05-20
Fix pandas FutureWarnings

uvchik:
pandas.DataFrame.append is deprecated

### DIFF
--- a/src/oemof/solph/processing.py
+++ b/src/oemof/solph/processing.py
@@ -209,15 +209,8 @@ def results(model, remove_last_time_point=False):
             df_dict[k] = df_dict[k].pivot(
                 columns="variable_name", values="value"
             )
-            # Add empty row at the end
-            df_dict[k] = pd.concat(
-                [
-                    df_dict[k],
-                    pd.DataFrame(data=["nan"], columns=df_dict[k].columns),
-                ],
-                ignore_index=True,
-                axis=0,
-            )
+            # Add empty row at the end df.loc[df.iloc[-1].name + 1,:] = np.nan
+            df_dict[k].loc[df_dict[k].iloc[-1].name + 1, :] = float("nan")
             set_result_index(df_dict, k, result_index)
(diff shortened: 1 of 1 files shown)


simnh:
Using the oemof storage invest example, I get an error due to different shapes...Somehow, for the storage with investment, the 
df_dict looks like this: 

```
(Pdb) df_dict[k]
variable_name  flow  invest
timestep                   
0               0.0     0.0
1               0.0     NaN
2               0.0     NaN
3               0.0     NaN
4               0.0     NaN
...             ...     ...
163             0.0     NaN
164             0.0     NaN
165             0.0     NaN
166             0.0     NaN
167             0.0     NaN
```

uvchik:
Thank you @simnh for the feedback.

I used a different approach to replace the deprecated `append` method and checked it with 17 different examples. I hope it is stable now.

src/oemof/solph/processing.py
```diff
@@ -210,7 +210,7 @@ def results(model, remove_last_time_point=False):
                 columns="variable_name", values="value"
             )
             # Add empty row at the end
-            df_dict[k] = df_dict[k].append(pd.Series(), ignore_index=True)
+            df_dict[k].loc[df_dict[k].iloc[-1].name + 1, :] = float("nan")
```
p-snft, line 213: Two considerations:
1. As pandas seems to use numpy.nan internally, it might be a good idea to directly do the same.
2. You add 1 to the name of the last entry and then create the entry using that name. It might help to understand the code better if there was a comment on what "name +1" means in this context.

uvchik: 1. I do not see the advantage, but if you like it better it is okay for me.
2. I shortened the expression and extended the comment.

uvchik: 1. I do not see the advantage, but if you like it better I can change it.
2. I shortened the code and extended the comment. It should be clearer now.

p-snft (approved):
Thanks for fixing this.

PR #850, opened 2022-05-20
Add examples from oemof-examples

uvchik:
We decided to move the solph examples to the solph repository.

### DIFF
--- a/examples/time_step_example/time_index_example.py
+++ b/examples/time_step_example/time_index_example.py
@@ -23,20 +23,15 @@ Installation requirements
 
 This example requires oemof.solph, install by:
 
-    pip install oemof.solph
+    pip install oemof.solph[examples]
 """
-import pandas as pd
 from oemof import solph
-
-try:
-    import matplotlib.pyplot as plt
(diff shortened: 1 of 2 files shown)


jnnr:
I had a look at examples/electrical/transhipment, which failed

```
  File "oemof-solph/src/oemof/solph/components/experimental/_link.py", line 215, in _flow1_rule
    + m.flow[i, link, t]
  File "pyomo/core/expr/numvalue.pyx", line 736, in pyomo.core.expr.numvalue.NumericValue.__mul__
  File "pyomo/core/expr/numeric_expr.pyx", line 1875, in pyomo.core.expr.numeric_expr._generate_mul_expression
AttributeError: 'NoneType' object has no attribute 'is_variable_type'

```

This is probably related to https://github.com/oemof/oemof-solph/issues/837, where we introduced `limit_direction` to switch off some constraints in Link. When I do the same in the transhipment example, the error does not occur any more. Doing this, the example is fixed, but Link together with `limit_direction=True` seems to be broken.

PR #852, opened 2022-06-02
Feature/constraint flow idle

jnnr:
Adds a new constraint to the solph.constraints module which allows to define an "idle time" for a flow (flow2) depending on another flow (flow1), i.e. a number of timesteps that flow1 has to be inactive before flow2 can be active. 

It can be created by `set_idle_time(model, f1, f2, n) `

It involves a product of two binary variables and thus leads to a quadratic program. Pyomo thinks it is a valid lp file, though. I have read that some solvers do automatically convert this into a linear formulation. If any reviewer comes up with a simple way to set this up using linear terms only, your suggestion is welcome.

**Update** Tested it with cbc and gurobi. Cbc refuses to run because of the quadratic constraint, gurobi manages to solve it.

**Update**: I rewrote the constraint to make it linear. Now, there are separate conditions for the different timesteps. To be precise, for each timestep `t` the status of the dependent flow (which has to be idle for n timesteps after the first flow was active) is compared with the status of the independent flow in timestep `t,...,t-n`.

TODOs:
* [x] Update mathematical formula.
* [ ] Allow to set status of flow1 for timestep t=-1
* [ ] Add a note in Whatsnew.

### DIFF
--- a/examples/idle_time_example/idle_time_example.py
+++ b/examples/idle_time_example/idle_time_example.py
@@ -1,0 +1,119 @@
+# -*- coding: utf-8 -*-
+
+"""
+
+"""
+
+__copyright__ = "oemof developer group"
+__license__ = "GPLv3"
+
+import os
+import pandas as pd
+import oemof.solph as solph
+
+import matplotlib.pyplot as plt
+
+
+solver = "cbc"
+
+# Create an energy system and optimize the dispatch at least costs.
+# ####################### initialize and provide data #####################
+
+datetimeindex = pd.date_range("1/1/2016", periods=10, freq="H")
+energysystem = solph.EnergySystem(timeindex=datetimeindex)
+
+# ######################### create energysystem components ################
+
+bus = solph.buses.Bus(label="Bus")
+sink = solph.components.Sink(
+    label="Sink",
+    inputs={
+        bus: solph.flows.Flow(
+            nominal_value=1,
+            variable_costs=-1,
+            min=0.2,
+            nonconvex=solph.NonConvex(),
+        )
+    },
+)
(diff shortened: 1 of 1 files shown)


jnnr:
To linearize the constraints, could express the products of two status variables x and y as a new variable z and set

```
z <= x  # if x is 0, x*y has to be 0
z <= y  # if y is 0, x*y has to be 0
z > =x + y - 1  # z is one if both x and y are 1
```

src/oemof/solph/constraints/set_idle_time.py
```diff
@@ -0,0 +1,90 @@
...
+    Parameters
+    ----------
+    model : oemof.solph.Model
+        Model to which the constraint is added.
+    f1 : tuple
+        First flow tuple.
+    f2 : tuple
+        Second flow tuple. Has to be inactive for a defined number of
+        timesteps after first flow was active.
+    n : int
```
p-snft, line 34: This constraint would be more versatile if the idile time would be given as a time. If non-equidistant time steps are used, their number is not really mmeaningful.

p-snft (changes requested):
I like the constraint, however, some improvements are needed:

* What's about time before the initial time step? Can you modify the constraint so that the status at (before) that time can be given?
* An example has to be added anyway.
* Giving time instead of steps would be great, but it's not a must, imho.

src/oemof/solph/constraints/set_idle_time.py
jnnr: Good idea! However, I see it as an optional future development that can be addressed when it is needed.

PR #854, opened 2022-06-21
v0.5 into dev

uvchik:
The `dev` branch is oudated. Nevertheless, there a still some commits there. To avoid annoying merge conflicts I would close `v0.5` in favour to `dev` and go on with an API breaking dev.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -5,7 +5,6 @@ on:
     branches:
       - master
       - dev
-      - v0.5
   pull_request:
 
 jobs:
--- a/.github/workflows/packaging.yml
+++ b/.github/workflows/packaging.yml
@@ -3,7 +3,7 @@ name: packaging
 on:
   # Make sure packaging process is not broken
   push:
-    branches: [master, dev, v0.5]
+    branches: [master, dev]
   pull_request:
   # Make a package for release
   release:
--- a/.github/workflows/tox_checks.yml
+++ b/.github/workflows/tox_checks.yml
@@ -6,7 +6,6 @@ on:
     branches:
       - master
       - dev
-      - v0.5
   pull_request:
 
   workflow_dispatch:
--- a/.github/workflows/tox_pytests.yml
+++ b/.github/workflows/tox_pytests.yml
@@ -5,7 +5,6 @@ on:
     branches:
       - master
       - dev
-      - v0.5
   pull_request:
 
   workflow_dispatch:


.github/workflows/tox_checks.yml
```diff
@@ -6,10 +6,9 @@ on:
     branches:
       - master
       - dev
+      - v0.5
```
p-snft, line 9: Shall we leave this for a future stable v0.5 branch?

.github/workflows/tox_pytests.yml
```diff
@@ -5,10 +5,9 @@ on:
     branches:
       - master
       - dev
+      - v0.5
```
p-snft, line 8: Shall we leave this for a future stable v0.5 branch?

.github/workflows/packaging.yml
```diff
@@ -3,7 +3,7 @@ name: packaging
 on:
   # Make sure packaging process is not broken
   push:
-    branches: [master, dev]
+    branches: [master, dev, v0.5]
```
p-snft, line 6: Shall we leave this for a future stable v0.5 branch?

.github/workflows/lint.yml
```diff
@@ -5,10 +5,8 @@ on:
     branches:
       - master
       - dev
+      - v0.5
```
p-snft, line 8: Shall we leave this for a future stable v0.5 branch?

p-snft (approved):
There is one remaining question (in four places), but I give my approval. The reason is simple: I'd really suggest to (again) have a stable branch for fixes and an unstable dev branch in the future. This would be a logical choice, i.e. as we also have branches called [v0.3](https://github.com/oemof/oemof-solph/tree/v0.3) and [v0.4](https://github.com/oemof/oemof-solph/tree/v0.4), which are also (old)stable. So, the current PR might merged unchanged. (As the changes will never see the old stable branches, it makes no sense adding them here.)

uvchik:
I removed `v0.5` from the branch checks and tests.

We kept the dev branch API stable to make it easier to release minor releases with new features. Otherwise you have to create v0.5.1, v0.5.2 ... branches for each minor release to release new features if there is an API breaking change in the dev branch.

In v0.3 we had a complex but not API breaking feature in the `dev` branch. Afterwards there was an API breaking change in the `dev` branch and some other changes. Now we wanted to release the new feature so we had to unravel the code to move non-API breaking changes into a new v0.3.1 branch and leave the rest in the `dev` branch. Therefore we decided to go the other way and for me it makes things easier. For me it is okay to change the `dev` branch to unstable, when we are heading to a new major release.

Furthermore , in my opinion we should not break the API without a warning, so API breaking changes can be done right before a release by removing old API warnings and the according code.

p-snft:
I had to reopen the branch v0.5 to be able to rebase remaining PRs.

PR #856, opened 2022-06-29
Feature/restucture flow

p-snft:
Notes: We need to either include #799 or this makes it obsolete.


Bachibouzouk:
Discussion with @p-snft , @joroeder , @sasa821 : there should not be InvestFlow, NonConvexFlow, NonConvexInvestFlow but just one Flow class and depending on the argument the user provides, the Flow can be nonconvex and/or with invest mode

p-snft:
Because NonConvex and Investment are no longer mutually exclusive, we may rethink the structure of Flow, NonConvexFlow and InvestmentFlow, because there is NonConvexInvestmentFlow. For the API, it might be more intuitive to have Invest and NonConvex as options. (In my memory we wanted to get rid of the options because they were confusing because they are exclusive.)

Bachibouzouk:
To define a nonconvex and investment flow

```
solph.flows.Flow(
    nominal_value=None,
    variable_costs=...,
    min=<number in (0,1)>,
    max=<number in [min,1)>,
    investment=solph.Investment(
        ep_costs=...,
        maximum=..., # need to have maximum provided here
    )
)
```
Another alternative
```
solph.flows.Flow(
    nominal_value=None,
    variable_costs=...,
    min=<number in (0,1)>,
    max=<number in [min,1)>,
    ep_costs=...,
    investment_maximum=..., # need to have maximum provided here
    )
)
```
To me it should be possible to not explicitly provide `nominal_value`
```
solph.flows.Flow(
    variable_costs=...,
    min=<number in (0,1)>,
    max=<number in [min,1)>,
    ep_costs=...,
    investment_maximum=..., # need to have maximum provided here
    )
)
```

jokochems:
Thanks for opening this discussion.

In general, I like the idea of having one single Flow class.
I don't like putting arguments into a flat argument structure. I think, it is much more intuitive and less error prone to use a nested structure, i.e. if you want an investment to happen, in my view, the most intuitive form is something like

```python
solph.flows.Flow(
    nominal_value=None,
    variable_costs=...,
    min=<number in (0,1)>,
    max=<number in [min,1)>,
    investment=solph.Investment(
        ep_costs=...,
        maximum=..., # need to have maximum provided here
    )
)
```

jokochems:
Now referring to the above questions:

> ```python
> solph.flows.Flow(
>     nominal_value=<number>,
>     variable_costs=...,
>     min=[0 <= float <= max],  # Should a minimum imply NonConvex? (Q1)
>                               # Until now, setting a min without NonConvex means always on.
>     max=[min <= float],  # Currently, max can be equal to min and can also exceed 1.
>                          # Should it stay like that? (Q2)
>     fix=[min <= float <= max],  # Should it be possible to have a series like [0, 0, .5, None, .2],
>                                 # where there is not fixed value for step 4? (Q3)
>                                 # If yes: Should it be possible to define min, max, and fix at the same time? (Q4)
> )
> ```

You are thinking very much in terms of modelling single technical units, @p-snft. I think, this is not necessarily the case and some of the changes would take some flexibility from the framework. I myself have use cases for what you want to exclude in Q1 and Q2. 

- Q1: I am **strongly** against letting `min` imposing a nonconvex flow. Think of modelling a power plant cluster with a minimum load boundary for it.

- Q2: I don't see a large gain from forcing `max` not to exceed 1. Think of an exogenous capacity expansion over time. But of course, you could scale that down to have 1 as maximum value. So I am against the adjustment, but not quite as strong as for Q1. Setting `max` equal to `min` can lead to some numerical issues and the solver not finding a solution, though. Maybe including some sort of numerical tolerance would be an idea here.

- Q3: I don't have a use case for including None values into a time series, i.e. only single steps being optimized. If it should be supported, I'd suggest to raise a user warning pointing to the time step missing. If it should not be supported (which is probably my preference), I think we need to discuss whether to raise an error or to include interpolation for missing values.
- Q4: Good point! I think, in most cases, this rather brings contradictions than benefits and you'd have to match missing values for the `fix` arguments with the `min` and `max` argument. We should find out whether there are users / proponents for allowing None values within the time series used for `fix` (Q3).

p-snft:
For Q1 and Q2, we have the same opinion.

For Q3 and Q4, I know that there are people who are setting min=max for single time steps, as you currently cannot set fix and min/max at the same time. However, there are known/fixed values just for a very limited number of time steps. (Another workaround would use two flows, one fix and one with min/max.)

joroeder:
My opinion and comments:

Q1: I am against that the `min` attribute leads to a nonconvex flow (Note that there is also the working case of bidirectional flows with `min=-1`).

Q2: Yes. It is practical that the maximum power can exceed the nominal_value with the `max` attribute, e.g. for heat pumps, when the nominal_value refers to a design sink and source temperature, and if the sink temperature is lower, the power output might be higher, and so on ...

Q3 & Q4: Interesting! For some E-mobility charging use cases this might be of interest. Is there anything against? If Q3 is possible, in my opinion, Q4 should also be possible, then these attributes allow powerful combinations of all these three constraints. And wouldn't it be easy to catch infeasible combinations?

uvchik:
I agree with @jokochems regarding the nested structure. I would add two attributes, one for the Investment object and one for the NonConvex object. Than it is easy to decide what mode should be used: Investment, NonConvex, NonConvexInvestment or SimpleFlow/DispatchFlow (I think we should rename the Block-Class from Flow to DispatchFlow or something like this. Than it would be clear that we have one Network class `Flow()` with four different Blocks.

Q1: I am against implicit switches.
Q2: I would not change that.
Q3: For me defining `min` and `max` to get a range in some time steps and a fixed value in other time steps (`min`=`max`) is not a workaround but straight forward. Defining a fixed time series with some `None` values is "hidden magic" for me. If a fixed value is None than the `min`/`max` value is valid and if the fixed value is not None the `min`/`max` value is not valid seems more complicated to me. If `fix` is set `min`/`max` is ignored!
Q4: No. As `min`/`max` could be contrastive to `fix` an error sould be raised or at least a Warning that they will be ignored.

joroeder:
Regarding the general structure, intuitively, the nested structure makes also more sense to me. I mean what would be the alternatives? Maybe, we could collect the alternatives with the pros and cons, however, there are already three opinions (@jokochems @uvchik @joroeder ) in favor of nested structure. So, maybe this is not necessary. But still, I think such an overview could help. I make an attempt:

Option 1:

* nested structure (as it is in v0.4.x, and first alternative of above [Bachibouzouk comment](https://github.com/oemof/oemof-solph/pull/856#issuecomment-1169967658))

Option 2:

* the user only sees the `Flow` class (alternatives 2, 3 and 4 of [Bachibouzouk comment](https://github.com/oemof/oemof-solph/pull/856#issuecomment-1169967658)) In the background, and depending on the given attributes, oemof.solph selects the appropriate Block in the background. This means that the whole API of the Flow would look something like this:

```
solph.flows.Flow(
    nominal_value=...,
    max=...,
    min=....,
    fix=...,
    positive_gradient=...,
    negative_gradient=...,
    full_load_time_max=...,
    full_load_time_min=...,
    variable_costs=...,
    integer=...,
    investment_ep_costs=...,
    investment_maximum=...,
    investment_minimum=...,
    investment_existing=...,  # could be combined with nominal_value
    investment_nonconvex=...,
    investment_offset=...,
    nonconvex_startup_costs=...,
    nonconvex_shutdown_costs=...,
    nonconvex_activity_costs=...,
    nonconvex_inactivity_costs=...,
    nonconvex_minimum_uptime=...,
    nonconvex_minimum_downtime=...,
    nonconvex_maximum_startups=...,
    nonconvex_maximum_shutdowns=...,
    nonconvex_initial_status=...,
    nonconvex_positive_gradient=...,    # duplicated
    nonconvex_negative_gradient=...,   # duplicated
    )
)
```

...of course, there might be some duplication, and of course, the `investment` and `nonconvex` prefix of the attribute names could be left out. This is just to give you an idea.

Option 3:

* the user has to select one of the classes each time: `Flow`, `InvestmentFlow`, `NonConvexFlow`, `NonConvexInvestFlow`, and each class allows only the attributes that are possible in this class. To my understanding, this way was originally intended for 0.5.x

p-snft:
> Hello @p-snft! Thanks for updating this PR. We checked the lines you've touched for [PEP 8](https://www.python.org/dev/peps/pep-0008) issues, and found:
> 
> * In the file [`src/oemof/solph/flows/_flow.py`](https://github.com/oemof/oemof-solph/blob/ca004d87fb846e40595d2e42fa34b8e941317af1/src/oemof/solph/flows/_flow.py):
> 
> > [Line 95:80](https://github.com/oemof/oemof-solph/blob/ca004d87fb846e40595d2e42fa34b8e941317af1/src/oemof/solph/flows/_flow.py#L95): [E501](https://duckduckgo.com/?q=pep8%20E501) line too long (91 > 79 characters)
> > [Line 97:80](https://github.com/oemof/oemof-solph/blob/ca004d87fb846e40595d2e42fa34b8e941317af1/src/oemof/solph/flows/_flow.py#L97): [E501](https://duckduckgo.com/?q=pep8%20E501) line too long (82 > 79 characters)
> 
> ##### Comment last updated at 2022-08-17 13:03:08 UTC

Both lines contain (long) references to submodules. How to cope with these?

CITATION.cff
```diff
@@ -97,4 +97,11 @@ authors:
     alias: "@stefansc1"
   -
     alias: "@jakob-wo"
+    given-names: jakob-wo
```
p-snft, line 100:
```suggestion
```

Bachibouzouk:
The nested structure seems to be the preferred way. I would also vote in favor of it. What I like is that the user doesn't have to know in advance that they have to use `Flow/DispatchFlow`, `InvestmentFlow`, `NonConvexFlow`, `NonConvexInvestFlow`. They just use `Flow` and depending on the options provided, `oemof-solph` which one of the four options it can belong to (are there possible settings which are not mutually exclusive, i.e. belonging to `InvestmentFlow`, `NonConvexFlow` at the same time ?).

src/oemof/solph/flows/_non_convex_invest_flow_block.py
```diff
@@ -0,0 +1,639 @@
...
+                for g in group
+                if g[2].nonconvex.positive_gradient["ub"][0] is not None
+            ]
+        )
+
+        # New nonconvex-investment-related set defines in the
+        # <class 'oemof.solph.flows.NonconvexInvestFlow'> class.
+        self.NON_CONVEX_INVEST_FLOWS = Set(
+            initialize=[(g[0], g[1]) for g in group]
+        )
```
p-snft, line 364: This name for a `Set` is already used in the `InvestmentFlowBlock`  to have non convex investments (i.e. minimal size that can be installed). If it's not possible to have a non-convex investment when also the flow is non-convex, this has to be docuumented and caught if users still try to do it. In any way, the name has to be different.

p-snft:
IMHO, the decision is clear by now. Thus I started with the implementation and removed the tags "feedback wanted" and "beginner".

I am currently removing code duplications by creating shared factory functions. I think it would be useful to also remove duplications in the documentation by moving the documentation from the blocks to the factory functions. When this is done, maybe some restructuring would make sense: If a variable or set is needed (just) for a certain constraint, it should be created with that one. Currently, we have a classic C-like layout where all variables are created first and populated later. Tis is rather hard to follow.

PS: For some reason, my local tox does not complain about Black issues, sorry.

src/oemof/solph/flows/_non_convex_invest_flow_block.py
p-snft: I [propose to rename](https://github.com/oemof/oemof-solph/pull/856/commits/509d44e6897ebac5896a738c153c4d80905fecc8) the new type of flows to `InvestNonConvexFlow`. The names will hopefully make clear that one is a non-convex investment into a flow while the other is an investment in a non-convex flow.

docs/whatsnew/v0-5-0.rst
```diff
@@ -7,7 +7,7 @@ API changes
 
 * Parts of the energy system graph are now clearly structured into `buses`, `components`, and `flows`.
 * Public and private API are be more distinguished now. ('_' signifies private, public API is defined in init files.)
-* Started API for distinct `Flow`, `InvestmentFlow`, and `NonConvexFlow`.
+* Started API for distinct `Flow`, `InvestmentFlow`, `NonConvexFlow`, and `NonConvexInvestFlow`.
```
p-snft, line 10: This needs to be adjusted.

src/oemof/solph/flows/_simple_flow_block.py
```diff
@@ -0,0 +1,324 @@
...
+    :math:`(i, o)`, which is omitted in the following for the sake of
+    convenience. The creation of some variables depend on the values of
+    *Flow* attributes. The following variables are created:
+
+    * :math:`P(t)`
+        Actual flow value (created in :class:`~oemof.solph._models.Model`).
+        The variable is bound to: :math:`f_{min}(t) \cdot P_{nom} \ge P(t) \le f_{max}(t) \cdot P_{nom}`.
+
+        If `Flow.fix` is not None the variable is bound to
+        :math:`P(t) = f_{fix}`.
```
p-snft, line 44:
```
                 :math:`P(t) = f_{fix} \cdot P_{nom}`.
```

p-snft:
Before I continue, I'd like to have some feedback on optimised code and duplication. I removed some specialised constraint formulations. This way, one formulation can be used for different kinds of flows. To do so, I introduce a variable even if something is constant (i.e. `nominal_value` of a `Flow` without investment). In a second step, I either consider it in the cost function (investment) or constraint it to a fixed value (no investment). This way, the LP files get longer, but in total, the number of different equations is reduced and the code becomes shorter.

To have an impression of the impact, I ran the "Generic Investment Limit" example a thousand times. (I picked it because it does not need the investment into non-convex flows. The only change I made is that I replaced the automatic node registration by explicit registration.) Turns out, the harmonised constraints are faster:

commit | feature level                   | average runtime (full script)
--------|--------------------------|----------------------------
1d373aa2a91fad1a70a79c4e74dcb30fba1f427b | harmonised constraints | 72.9 ms +/-  7.2 ms
08db3a5d0f2b2b3c9be84231deecd71a5e55cf9c | old constraints              |	81.4 ms +/- 9.6 ms
e95fcc3de5ad8ae79ba5704b8c145cfe15dfbd4b | dev				     | 85.1 ms +/- 9.7 ms

src/oemof/solph/flows/_non_convex_invest_flow_block.py
Bachibouzouk: I would understand `InvestNonConvexFlow` as `Invest + NonConvexFlow` and not `NonConvexInvest + Flow`. I would be more in favor of normal flow with non-convex investment being named `NonConvexInvestFlow` and non-convex flow with normal investment being named `InvestmentNonConvexFlow`.

p-snft: @Bachibouzouk : Isn't this is the way it is now in the current PR?

p-snft:
Almost done. Docs and examples still need to be checked.

src/oemof/solph/flows/_non_convex_invest_flow_block.py
Bachibouzouk: You renamed `NonConvexInvestFlow` to `InvestNonConvexFlow` to describe a non-convex investment into a flow, I think the name `NonConvexInvestFlow` is a better name for a non-convex investment into a flow

joroeder:
I am wondering, if we introduce the subpacke `flows/_shared`, we should probably shift the code from the module `_options` into the two new modules `investment.py` and `non_convex.py`. ' I think it perfectly fits in there, as `_options` contains only the two classes `Investment` and `Nonconvex`.

joroeder:
There will be still dublication of the block docstrings for attributes/constraints that can be used in all blocks, e.g. `min`, etc., right? I mean that is not a problem. The alternative would be to move the documentation of the mathematical constraints "one level lower" to the single methods for creating each constraint, but this would not be very clear but diffuse.

p-snft:
> I am wondering, if we introduce the subpackage `flows/_shared`, we should probably shift the code from the module `_options` into the two new modules `investment.py` and `non_convex.py`. ' I think it perfectly fits in there, as `_options` contains only the two classes `Investment` and `Nonconvex`.

I see two reasons against that: 
1. There is investment into storages.
2. The current solution having "_shared" is not ideal and might not persist. Currently, only the non-convex flow blocks really share those factories, while it turned out to be more difficult for the investment. Thus, I am considering to derive `InvestNonConvexFlowBlock` from`NonConvexFlowBlock`. (I realised that when I started to clean the documentation. Thus, I am planning to try this solution during the next couple of days.)

> There will be still duplication of the block docstrings for attributes/constraints that can be used in all blocks, e.g. min, etc., right? I mean that is not a problem. The alternative would be to move the documentation of the mathematical constraints "one level lower" to the single methods for creating each constraint, but this would not be very clear but diffuse.

To tackle that, and to also move the documentation closer to the code it describes, I am considering to have the documentation in (private?) member functions that are used to create the variables, sets, and constraints. I added one example implementing that in one of the last commits: https://github.com/oemof/oemof-solph/pull/856/commits/eeed2aeb513c63f40beb33743e8e0438b9f6f52b#diff-ccd2682cb6d677f000ce744bdf52fa15c007372eba109ae2f31de90604dc139a

joroeder:
> 
> I see two reasons against that:
> 
> 1. There is investment into storages.
> 
> 2. The current solution having "_shared" is not ideal and might not persist. Currently, only the non-convex flow blocks really share those factories, while it turned out to be more difficult for the investment. Thus, I am considering to derive `InvestNonConvexFlowBlock` from`NonConvexFlowBlock`. (I realised that when I started to clean the documentation. Thus, I am planning to try this solution during the next couple of days.)

Sounds reasonable to me. And yes, there is also the storage, and maybe in future some other components with investment.

> 
> 
> > There will be still duplication of the block docstrings for attributes/constraints that can be used in all blocks, e.g. min, etc., right? I mean that is not a problem. The alternative would be to move the documentation of the mathematical constraints "one level lower" to the single methods for creating each constraint, but this would not be very clear but diffuse.
> 
> To tackle that, and to also move the documentation closer to the code it describes, I am considering to have the documentation in (private?) member functions that are used to create the variables, sets, and constraints. I added one example (partly) implementing that in one of the last commits: [eeed2ae#diff-ccd2682cb6d677f000ce744bdf52fa15c007372eba109ae2f31de90604dc139a](https://github.com/oemof/oemof-solph/commit/eeed2aeb513c63f40beb33743e8e0438b9f6f52b#diff-ccd2682cb6d677f000ce744bdf52fa15c007372eba109ae2f31de90604dc139a)

This would be one way to go and probably we should try that based on one or two examples (thank your for the examples so far), and then check, if we are satisfied with that solution. Because I think it would contradict the previous strategy of how the block docstrings should look like (e.g. having the most important content, e.g. without details info about the sets or groups, but the variables, parameters and constraints, e.g. as here: [https://oemof-solph.readthedocs.io/en/v0.4.4/reference/oemof.solph.html#oemof.solph.blocks.investment_flow.InvestmentFlow](https://oemof-solph.readthedocs.io/en/v0.4.4/reference/oemof.solph.html#oemof.solph.blocks.investment_flow.InvestmentFlow)). So, I fear that in the end this closer-to-the-code strategy will be less informative and more complicated to read for the user, but for sure the way to go from a software-engineering perspective.

p-snft:
The style is not 100 % customisable any more, but the order of the documented functions can be customised. I continued with the [doc of InvestNonConvexFlow](https://oemof-solph.readthedocs.io/en/feature-restucture_flow/reference/oemof.solph.flow.html#module-oemof.solph.flows._invest_non_convex_flow_block) as an example.

src/oemof/solph/flows/_invest_non_convex_flow_block.py
```diff
@@ -0,0 +1,474 @@
...
+class InvestNonConvexFlowBlock(ScalarBlock):
+    r"""
+    **The following sets are created similar to the
+    <class 'oemof.solph.flows.NonConvexFlow'> class:**
+    (-> see basic sets at :class:`.Model` )
+
+    .. document private functions
+    .. automethod:: _create_variables
+    .. automethod:: _create_sets
+
```
joroeder, line 41:
```suggestion

    .. automethod:: oemof.solph.flows._non_convex_flow_block::NonConvexFlowBlock._create_sets

    .. autofunction:: oemof.solph.flows._shared.investment.minimum_investment_constraint

```

joroeder:
> The style is not 100 % customisable any more, but the order of the documented functions can be customised. I continued with the [doc of InvestNonConvexFlow](https://oemof-solph.readthedocs.io/en/feature-restucture_flow/reference/oemof.solph.flow.html#module-oemof.solph.flows._invest_non_convex_flow_block) as an example.

I also think if the `invest_non_convex_flow_block` is done, the other docstrings are straight forward.

src/oemof/solph/flows/_invest_non_convex_flow_block.py
```diff
@@ -373,8 +372,8 @@ def _max_invest_rule(block, i, o):
             self.INVEST_NON_CONVEX_FLOWS, rule=_max_invest_rule
         )
 
-        self.min = nc.minimum_flow_constraint(self)
-        self.max = nc.maximum_flow_constraint(self)
+        self.min = self._minimum_flow_constraint()
+        self.max = self._maximum_flow_constraint()
 
         # z = x * y, where x is a binary variable (in our case `status`),
         # y is a continuous variable (in our case `invest`), and z denotes
```
sasa821, line 379: I like the idea of renaming the new parameter to `status_nominal`. It is more descriptive! But the old name stil exists in some parts. Here, it should be:
```
... and z denotes the new parameter `status_nominal`.
```

Similarly, for the following functions,
1. `_linearization_rule_invest_non_convex_one`
2. `_linearization_rule_invest_non_convex_two`
3. `_linearization_rule_invest_non_convex_three` 

the docstrings can be updated as follows:
```
Rule definition for the linearization of the `status_nominal`.
```

sasa821 (commented):
To me it looks very good. Thank you!

p-snft:
While refactoring, I found the following description in the documentation.
```Python
    If `nonconvex.positive_gradient["costs"]` is set by the user:
        .. math::
            \sum_{i, o \in POSITIVE_GRADIENT_FLOWS} \sum_t
            positive_gradient(i, o, t) \cdot positive\_gradient\_costs(i, o)

    If `nonconvex.negative_gradient["costs"]` is set by the user:
        .. math::
            \sum_{i, o \in NEGATIVE_GRADIENT_FLOWS} \sum_t
            negative_gradient(i, o, t) \cdot negative\_gradient\_costs(i, o)
```
However, I did not find the according implementation. I have the feeling, it was never done. If you do not object, I will just delete this part of the documentation.

joroeder:
I remember that @uvchik talked about deleting these constraints last year or so, as they were not implemented fully/probably/correctly. Probably, this is a left-over. So, these lines should be deleted.

src/oemof/solph/flows/_invest_non_convex_flow_block.py
p-snft: I refactored the according code and changed the documentation. Do you like it?

https://oemof-solph.readthedocs.io/en/feature-restucture_flow/reference/oemof.solph.flow.html#oemof.solph.flows._invest_non_convex_flow_block.InvestNonConvexFlowBlock._linearised_investment_constraints

p-snft:
The documentation is not fully coherent and the "convex" flows are not refactored, yet. However, these are changes "under the hood" inside the corresponding classes. Thus, I'd like to merge this to be able to work on other stuff without creating conflicts. Is this okay for you, @oemof-developer?

joroeder:
I also think that is a good idea. To finish all the restructuring, e.g. especially make all docstrings consistent, this could be done via separate PRs. And then other features can directly branch from and into the current dev, and do not need to wait until the last detail of this restructuring is done.

@p-snft : Could you open an issue with todos and the open points you see to complete the flow restructuring?

p-snft:
@joroeder: Sure, I'll collect a Todo list. By the way, is this an official positive review for the PR?

docs/usage.rst
```diff
@@ -1003,25 +1007,127 @@ you have to do is to invoke a class instance inside your Flow() - declaration:
...
+.. code-block:: python
+
+    b_diesel = solph.buses.Bus(label='diesel')
+    b_el = solph.buses.Bus(label='electricity')
+
+    solph.components.Transformer(
+        label='diesel_genset',
+        inputs={b_diesel: solph.flows.Flow()},
+        outputs={
+            b_el: solph.flows.NonConvexInvestFlow(
```
Bachibouzouk, line 1075:
```suggestion
            b_el: solph.flows.NonConvexInvestmentFlow(
```

examples/nonconvex_invest_flow_examples/house_with_nonconvex_investment.py
```diff
@@ -0,0 +1,149 @@
...
+)
+
+# For one year, the equivalent periodical costs (epc) of an
+# investment are equal to the annuity.
+epc = economics.annuity(5000, 20, 0.05)
+
+fireplace = solph.components.Source(
+    label="fireplace",
+    outputs={
+        b_heat: solph.flows.NonConvexInvestFlow(
```
Bachibouzouk, line 83:
```suggestion
        b_heat: solph.flows.NonConvexInvestmentFlow(
```

src/oemof/solph/components/_offset_transformer.py
```diff
@@ -78,7 +78,7 @@ def __init__(self, *args, **kwargs):
             for k, v in self.inputs.items():
                 if not v.nonconvex:
                     raise TypeError(
-                        "Input flows must be of type NonConvexFlow!"
+                        "Input flows must be have NonConvex attribute!"
```
Bachibouzouk, line 81:
```suggestion
                        "Input flows must have NonConvex attribute!"
```

src/oemof/solph/flows/_flow.py
```diff
@@ -87,11 +83,24 @@ class Flow(on.Edge):
...
+        will be used instead of
+        :class:`~oemof.solph.flows._simple_flow_block.SimpleFlowBlock`.
 
     Notes
     -----
-    See :py:class:`~oemof.solph.flows._flow.FlowBlock` for the variables,
-    constraints and objective parts, that are created for a FLow object.
+    See :py:class:`~oemof.solph.flows._simple_flow.SimpleFlowBlock`
+    for the variables, constraints and objective parts, that are created for
+    a FLow object.
```
Bachibouzouk, line 103:
```suggestion
    a Flow object.
```

src/oemof/solph/flows/_flow.py
```diff
@@ -87,11 +83,24 @@ class Flow(on.Edge):
         :attr:`fix`.
     integer : boolean
         Set True to bound the flow values to integers.
+    investment : :class:`Investment <oemof.solph.options.Investment>`
+        Object indicating if a nominal_value of the flow is determined by
+        the optimization problem. Note: This will refer all attributes to an
+        investment variable instead of to the nominal_value. The nominal_value
```
Bachibouzouk, line 89:
```suggestion
        investment variable rather than to the nominal_value. The nominal_value
```

src/oemof/solph/flows/_non_convex_flow_block.py
```diff
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
```
Bachibouzouk, line 1: This file here is difficult to review in details

src/oemof/solph/flows/_simple_flow_block.py
```diff
@@ -0,0 +1,327 @@
+# -*- coding: utf-8 -*-
```
Bachibouzouk, line 1: I assume this was mostly copy-pasted from `src/oemof/solph/flows/_flow.py`

tests/constraint_tests.py
```diff
@@ -178,6 +178,31 @@ def test_linear_transformer_invest(self):
...
+    def test_nonconvex_invest_transformer(self):
+        """Non-convex invest flow with offset, without minimum."""
+        bfuel = solph.buses.Bus(label="fuelBus")
+        bel = solph.buses.Bus(label="electricityBus")
+
+        solph.components.Transformer(
+            label="transformer_nonconvex_invest",
+            inputs={bfuel: solph.flows.Flow()},
+            outputs={
+                bel: solph.flows.Flow(
```
Bachibouzouk, line 190: Shouldn't it be `NonConvexInvestmentFlow` ?
```suggestion
                bel: solph.flows.NonConvexInvestmentFlow(
```

Bachibouzouk: Maybe I will retract this comment as I advance in the review (I am making it now because I don't know if I would have time to finish reviewing all files tonight)

tests/test_components.py
```diff
@@ -227,7 +227,7 @@ def test_generic_storage_too_many_outputs():
 def test_offsettransformer_wrong_flow_type():
     """No NonConvexFlow for Inflow defined."""
     with pytest.raises(
-        TypeError, match=r"Input flows must be of type NonConvexFlow!"
+        TypeError, match=r"Input flows must be have NonConvex attribute!"
```
Bachibouzouk, line 230:
```suggestion
        TypeError, match=r"Input flows must have NonConvex attribute!"
```

Bachibouzouk (commented):
Great work restructuring here!

Were `examples/nonconvex_invest_flow_examples/`
-  `diesel_genset_nonconvex_investment.py` 
- `house_with_nonconvex_investment.py`
- `house_without_nonconvex_investment.py`
 ran? 

I would search for occurences of `NonConvexInvestFlow` and replace them by `NonConvexInvestmentFlow`

tests/test_warnings.py
```diff
@@ -133,3 +133,21 @@ def test_storage_without_inputs(warning_fixture):
...
+
+
+def test_nonconvex_investment_without_maximum_raises_warning(warning_fixture):
+    """
+    <class 'solph.flows.NonConvexInvestFlow'> without specifying
+    the maximum attribute of the <class 'solph.Investment'>
+    """
+
+    with pytest.raises(AttributeError):
+        solph.flows.NonConvexInvestFlow(
```
Bachibouzouk, line 145:
```suggestion
        solph.flows.Flow(
```

tests/test_warnings.py
```diff
@@ -133,3 +133,21 @@ def test_storage_without_inputs(warning_fixture):
...
+            nominal_value=None,
+            variable_costs=25,
+            min=0.2,
+            max=0.8,
+            investment=solph.Investment(
+                ep_costs=500,  # no maximum is provided here
+            ),
```
Bachibouzouk, line 152:
```suggestion
            ),
            nonconvex=solph.NonConvex(),
```

tests/test_warnings.py
```diff
@@ -133,3 +133,21 @@ def test_storage_without_inputs(warning_fixture):
         )
         assert len(w) == 1
         assert msg in str(w[-1].message)
+
+
+def test_nonconvex_investment_without_maximum_raises_warning(warning_fixture):
+    """
+    <class 'solph.flows.NonConvexInvestFlow'> without specifying
```
Bachibouzouk, line 140:
```suggestion
    <class 'solph.flows.Flow'> without specifying
```

jokochems:
I'm not going to go into details with a code review and at the moment, unfortunately, my time for oemof development is very limited, but I had a quick look and really like the structural improvements, this PR brings along. I'm a bit worried about integrating #828, though, but I'll try to cut some time in October to resolve merge conflicts and copy & paste the changes brought along. ;-)

p-snft:
> Were `examples/nonconvex_invest_flow_examples/`
> -  `diesel_genset_nonconvex_investment.py` 
> - `house_with_nonconvex_investment.py`
> - `house_without_nonconvex_investment.py`
>  ran? 
> 
> I would search for occurences of `NonConvexInvestFlow` and replace them by `NonConvexInvestmentFlow`
> 
> I see that  the class `InvestNonConvexFlowBlock` is often referred to as `NonConvexInvestFlowBlock` in docstrings and comments
> 

Thanks for your feedback. Adding to these points, I also forgot to align the naming if the example with the new wording.

src/oemof/solph/flows/_non_convex_flow_block.py
p-snft: It should be just refactored in the sense that code has been grouped in functions.

joroeder (approved):
When going through the changes, I updated the links in docs. (which is also part of #863, and made some minor suggestion for improvement.

Do we need an `__init__.py` in the examples. I think this is not necessary and we should remove the file.

Besides, I do not see anything against merging this PR.

docs/usage.rst
```diff
@@ -1003,25 +1007,127 @@ you have to do is to invoke a class instance inside your Flow() - declaration:
...
+    \begin{align}
+        &invest\_non\_convex \le status \times invest_{max} \\
+        &invest\_non\_convex \le invest \\
+        &invest\_non\_convex \ge invest - (1-status) \times invest_{max}
+    \end{align}
+
+In the above equations, :math:`status` is the binary variable denoting the operation status of the flow at each time,
+:math:`invest_{max}` is the maximum capacity of the component being optimized, which is a required fixed input parameter
+that must be given by providing the `maximum` parameter in the :py:mod:`~oemof.solph.options` of the
+:py:class:`~oemof.solph.flows.NonConvexInvestmentFlow` class (see example below), and :math:`invest\_non\_convex` is an attribute
```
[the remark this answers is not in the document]
sasa821: I think @p-snft changed the name of this new variable to `status_nominal`, which is more descriptive, in my opinion.

p-snft:
As far as I see, there is no objection and one positive review.As some PRs already build on the changes presented here, I will now merge. Remaining clean-up (of the documentation and adjustments to the other flow classes) is addressed in #863.

PR #857, opened 2022-08-18
Fix/reformat documentation

p-snft:
This PR introduces a slight change to the test pipeline
so that it will fail on problems in the documentation.
To make this possible, all kinds of problems with the
documentation are fixed. These include:
* Configuration of Sphinx
* Internal references not being updated
* LaTeX errors
* Formating of inline code examples


joroeder:
Everything looks fine 👍

joroeder (approved):

PR #859, opened 2022-09-14
Maximum uptime constraint

joroeder:
This PR adds a maximum uptime constraint for NonConvex flows.

The motivation is a scheduling problem for industrial processes, if you have e.g. a cyclic industrial process (block profile) with a certain time between each process and a fixed length of the process. With a `maximum_uptime` attribute and in combination with the existing `minimum_downtime` and other attributes, it is possible to model this problem within oemof.solph. See the example.

The API is an additional attribute at the NonConvex class:

```
Flow(
    nonconvex=NonConvex(
        maximum_uptime=4,
    )
)
```

The constraint itself is based on the rule that the sum of `maximum_uptime` +1 consecutive binary decision variables of the flow need to be smaller or equal the maximum operation time `maximum_uptime`.

If there is nothing against the implementation of this constraint, I would continue to implement docstrings, tests, etc. I would a appreciate a short feedback on that.

p.s.: This constraint would allow the scheduling of block profiles. We would be also interested to schedule fixed but fluctuating profiles. Any ideas?


src/oemof/solph/flows/_non_convex_flow.py
```diff
@@ -373,6 +373,13 @@ def _create(self, group=None):
                 if g[2].nonconvex.minimum_uptime is not None
             ]
         )
+        self.MAXUPTIMEFLOWS = Set(
```
Bachibouzouk, line 376: Must be placed close to https://github.com/oemof/oemof-solph/blob/b6dc604527fb865ee67744534b2ad789bb670a64/src/oemof/solph/flows/_non_convex_flow_block.py#L231 as the _non_convex_flow.py was refactored into _non_convex_flow_block.py

Bachibouzouk:
Closing due to inactivity and because there are no constraint proposed in the code yet

PR #866, opened 2022-09-26
Fix doc of GenericStorage

p-snft:
The length of the time steps is given in hours. Thus, the old explanation was at least misleading.


p-snft:
See [non-equidistant time step example](https://github.com/oemof/oemof-solph/blob/7f56f81c84b62be8b2372f6955e2ea73d0a831a3/examples/time_step_example/non_equidistant_time_step_example.py) for reference.

p-snft:
This is an almost trivial change, I will just merge.

PR #867, opened 2022-09-27
Revision/remove node registry

p-snft:
The feature to automatically add Nodes (strictly speaking Entities) of an oemof.network to an energy system is deprecated. Thus, it should not be used in the tests. This PR implements that.

PS: This PR is against feature/restructure_flow, as I branched from that branch. It would be cleaner to merge #856 first and merge this into dev afterwards.


p-snft:
Now included in #836.

Issue #861, opened 2022-09-16
OffsetTransformer with nonconvex and investment options


sasa821:
Currently, the `OffsetTransformer` can be used **ONLY** when its `nominal_value` is already known, which means, it is not possible to find the optimal capacity of a transformer with a given efficiency curve (instead of a single efficiency).  

Nevertheless, after #856 is finalized, it will be possible to have this functionality. But the `OffsetTransformer` needs to be adjusted because, in the current version, the `nonconvex` option MUST be defined for the input flow. This will cause problems if the optimal capacity of the component is unknown. In this situation, the `nonconvex` option and the `investment` option must be defined for the output flow. The reason is that the `ep_cost` in the investment option should apply to the outflow of assets, and it is impossible to map this value for the input flow. Moreover, we cannot define `investment` in the outflow and `nonconvex` in the inflow because, in this case, since the `nominal_value` would be unknown, both these options must be used in the same flow.

I am making some changes in the `OffsetTransformer` and will soon open a pull request about the proposed changes, but it would be nice to discuss it here, if there is disagreement on this issue.

sasa821:
@p-snft I would like to finalize the `OffsetTransformer` according to the new flow structure. Is it OK if I open a new PR in #856, or is it better to wait until #856 is merged into the `dev`?

p-snft:
@sasa821: You can open a PR against `/feature/restructure_flow`. I did the same with #836. (If you think #856 is ready to be merged, state it over there and I will do so. I fixed the example you asked about and now I'm just waiting for a positive review.)

sasa821:
> There was a reason putting the offset on the inflow, but I don't remember exactly. It may be that this matters only in the case, if `min=0`, than there could be outflow at zero inflow, or something like that ..
> 
> Nonetheless, there should be nothing against opening a PR. It would be really nice in combination with an investment. It makes for sure sense, if you take the new structure of the flows as basis. If it was really just the case I described above, you could then also issue a warning. Or you make it more general somehow, that you can set the offset arbitrarily to the in- or outflow, that would be cool anyway.

I believe it would make more sense to define the `NonConvex` and `Investment` options, both **ONLY** on the output flow. Otherwise, the implementation of the required adjustments in the `OffsetTransformer` would become unnecessarily complicated. It would also be confusing for the users to sometimes use the input flow and sometimes use the output flow as the basis for modeling the efficiency curve. On the other hand, I still do not understand why the outflow should be non-zero if the inflow is zero.

[...]

PR #868, opened 2022-10-14
Feature/restucture flow -- upgrade OffsetTransformer

sasa821:
### Explanation:

Based on the new structure of flows according to #826 and #856, the model for the `OffsetTransformer` can be improved significantly. So far, this component could **ONLY** take the `NonConvex` attribute. This means it was assumed that the component had already been optimized and the nominal capacity was known. However, the component’s **real** efficiency curve and its (probably) less-efficient operation in part-load could not be considered in the optimization until now. 

Thanks to the new structure of the `Flow`, which enables the combination of `NonConvex` and `Investment` attributes, these parameters can now be part of the optimization. The following example shows that the results obtained from the previous optimization (applied to the `Transformer`) and the new one (applied to the `OffsetTransformer`) are considerably different.

Another change suggested in this PR is the definition of the `NonConvex` and `Investment` attributes. In the previous version of the `OffsetTransformer`, the `NonConvex` attribute had to be defined for the **input flow**, but in the new model proposed in this PR, it must be defined for the **output flow** (see #861  for more information).

### Example:
As an example, the efficiency and load duration curves of a diesel genset are shown, which are obtained from different optimization methods. As explained before, previously, it was only possible to use the `Transformer` for obtaining the optimal size of a component, but restructuring the flow provided the possibility to use the `OffsetTransformer` for obtaining the optimal capacity.

offset-transformer

As seen in the figures, the diesel genset has a minimum and maximum load of 20 and 100%. The `Transformer` component takes a _constant efficiency_ as an input (here 33%), but the `OffsetTransformer` component deals with the _efficiency curve_ of this component, which is obtained using the minimum and maximum loads and efficiencies. 

It can be seen that the optimal size of the diesel genset when the `OffsetTransformer` is used is around **8% smaller**. The reason is that since this component is not operating efficiently at low loads, instead of having a larger capacity and operating at low loads, the optimizer tries to reduce the capacity so that it does not have to operate inefficiently at very low loads. This can be seen in the diagrams very clearly. On the right side, where the `Transformer` is used to model the diesel genset, since the efficiency is assumed constant, in approximately 15% of the annual operation of this component, it operates at the minimum load. However, as seen on the left side, the `OffsetTransformer` does not allow the diesel genset to operate at the minimum load.


sasa821:
> Will you directly incorporate the changes suggested in #822 (and change the name to `OffsetConverter`)? Or shall I wait with the other PR until you are done?

Is it OK to change the base of this PR from [dev](https://github.com/oemof/oemof-solph/tree/dev) to [revision/rename_transformer_to_converter](https://github.com/oemof/oemof-solph/tree/revision/rename_transformer_to_converter) and then proceed with the renaming?

p-snft:
> Is it OK to change the base of this PR from [dev](https://github.com/oemof/oemof-solph/tree/dev) to [revision/rename_transformer_to_converter](https://github.com/oemof/oemof-solph/tree/revision/rename_transformer_to_converter) and then proceed with the renaming?

You want to merge this into [revision/rename_transformer_to_converter](https://github.com/oemof/oemof-solph/tree/revision/rename_transformer_to_converter) before I continue with it? Sounds reasonable. Thanks for the suggestion.

sasa821:
This PR is closed here and will be finalized in #895 to incorporate the changes suggested in #822.

PR #869, opened 2022-10-14
Features/schedule profiles

joroeder:
    * Scheduling of fixed load profiles, e.g. you have an industrial process with a fixed load profile, and you are wondering how to schedule this process in time.
  * Where is it located?
    * NonConvex option and NonConvex Flow Block
  * How does the API look?
  
  ```
  nonconvex=NonConvex(
            minimum_downtime=3,
            minimum_uptime=4,
            maximum_uptime=4,
            profile=[2, 3, 4, 5],
            startup_costs=10,
            # shutdown_costs=2,
        ),
  ```

* Related issues?
  * #859

  * This is a first draft, it is still under development
  * It makes use of the binary start-up variable, so at the moment, `startup_costs` must not be none. Roughly, the binary startup variable is multiplied with the fixed profile for each timestep.
  * At the moment, the profile must not begin or end with values of zero. This needs to be considered via a minimum_downtime.
  * Furthermore, the first process cannot start before the lenght of the profile (The starting phase is not implemented or the constraint need to be reformulated to make this possible)

  * I am wondering if anybody did something before or is interesting is such a feature and could support development, e.g. also by testing and using this feature.

### DIFF
--- a/examples/min_max_uptime_example/min_max_uptime_example.py
+++ b/examples/min_max_uptime_example/min_max_uptime_example.py
@@ -2,29 +2,39 @@
 """
 Example that shows how to use a combination of minimum and maximum uptime.
 
+The model consists of a volatile renewable energy source with
+a fixed profile (Source: 'renewable_energy'), a cyclic industrial process
+modelled as a Sink with a NonConvex flow with negative variable costs as
+revenues, and an excess for surplus of renewable energy.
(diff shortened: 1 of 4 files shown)


Bachibouzouk:
Closing because of inactivity. The feature is not uninteressant though

PR #877, opened 2022-11-09
Make kwargs explicit

### LINKED ISSUE #729: How can typos be caught?
This started with my PR #727. Before implementing another suggestion, I feel like there should first be a rough consensus about what the solution should be.

First, the problem: People make typos. When someone writes `Flow(variable_cost=42)`, the missing s (did you spot it?) is silently ignored and one gets a wrong result. Thus, the suggestion is to catch these typos. Somehow.

The problem is complicated by people wanting to be able to add arbitrary properties to objects. So, `print(Flow(my_cool_property=42).my_cool_property)` should produce `42`.

So, here are some random thoughts about possible aspects of a solution:

- When should typos be checked?
  - At each single object construction
  - When the `Model` instance is constructured from the energy system
  - When solving a model
  - With an explicit call to `check_for_unknown_properties()`
- How should the checking work?
  - Instead of having `Flow.__init__(**kwargs):`, only the known properties are available as named arguments. New properties could be added via a special `custom_properties` argument. (Of course, the name for this argument is open for bike-shedding)
  - There could be an internal list of known properties. Everything not known is "bad". Users could add own properties.
    - ...in a global list of known properties
    - ...as an argument somewhere somehow to the check itself
- What should the result of the check be?
  - Exception
  - SuspiciousUsageWarning
  - a list/dict/... of `$SOMETHING` that is returned to the caller
  - Nothing (the current state of things: no typo checks done)

Uhm... that's it. These are all the possibilities I could come up with. Feel free to think about this and add more options. If you have the powers to do so, feel free to edit this comment so that there is "a single list of everything".

No idea how the usual oemof decision making works. Depending on mood, time, and the phase of the moon, I might or might not be able to send a PR for whatever version is preferred.

joroeder:
now, with better branch name according to @gnn ;-)

Based on multiple discussions, see Issue: ...

* [x] Flow #881 
* [x] GenericStorage #878 
* [x] Transformer #891  
* [x] Sink https://github.com/oemof/oemof-solph/pull/885
* [x] Source https://github.com/oemof/oemof-solph/pull/886
* [x] Investment
* [x] Nonconvex
* [x] GenericCHP
* [x] Extraction_turbine
* [x] Offset_transformer

Fixes #729

### DIFF
--- a/src/oemof/solph/_helpers.py
+++ b/src/oemof/solph/_helpers.py
@@ -35,6 +35,7 @@ def check_node_object_for_missing_attribute(obj, attribute):
 
 
 def warn_if_missing_attribute(obj, attribute):
+    """Raises warning if attribute is missing for given object"""
     msg = (
         "Attribute <{0}> is missing in Node <{1}> of {2}.\n"
         "If this is intended and you know what you are doing you can"
--- a/src/oemof/solph/_options.py
+++ b/src/oemof/solph/_options.py
@@ -187,8 +187,6 @@ class NonConvex:
         for attribute, value in custom_attributes.items():
             setattr(self, attribute, value)
 
-        self._max_up_down = None
-
     @property
     def max_up_down(self):
         """Return maximum of minimum_uptime and minimum_downtime.
--- a/src/oemof/solph/components/_generic_chp.py
+++ b/src/oemof/solph/components/_generic_chp.py
@@ -79,8 +79,8 @@ class GenericCHP(network.Transformer):
         Dictionary with key-value-pair of `oemof.Bus` and `oemof.Flow` object
         for the heat output. Related parameters like `Q_CW_min` are passed as
         attributes of the `oemof.Flow` object.
-    Beta : list of numerical values
-        Beta values in same dimension as all other parameters (length of
+    beta : list of numerical values
+        beta values in same dimension as all other parameters (length of
         optimization period).
     back_pressure : boolean
         Flag to use back-pressure characteristics. Set to `True` and
@@ -111,7 +111,7 @@ class GenericCHP(network.Transformer):
     ...        })},
     ...    heat_output={bth: solph.flows.Flow(
     ...        custom_attributes={"Q_CW_min": [10.552]})},
-    ...    Beta=[0.122], back_pressure=False)
+    ...    beta=[0.122], back_pressure=False)
     >>> type(ccet)
     <class 'oemof.solph.components._generic_chp.GenericCHP'>
     """
@@ -121,7 +121,7 @@ class GenericCHP(network.Transformer):
         fuel_input,
         electrical_output,
         heat_output,
-        Beta,
+        beta,
         back_pressure,
         label=None,
         custom_attributes=None,
@@ -133,7 +133,7 @@ class GenericCHP(network.Transformer):
         self.fuel_input = fuel_input
         self.electrical_output = electrical_output
         self.heat_output = heat_output
-        self.Beta = sequence(Beta)
+        self.beta = sequence(beta)
         self.back_pressure = back_pressure
         self._alphas = None
 
@@ -290,7 +290,7 @@ class GenericCHPBlock(ScalarBlock):
     :math:`Y`                       `Y[n,t]`                V    status variable on/off
     :math:`\alpha_0`                `n.alphas[0][n,t]`      P    coefficient describing efficiency
     :math:`\alpha_1`                `n.alphas[1][n,t]`      P    coefficient describing efficiency
-    :math:`\beta`                   `Beta[n,t]`             P    power loss index
+    :math:`\beta`                   `beta[n,t]`             P    power loss index
     :math:`\eta_{el,min,woDH}`      `Eta_el_min_woDH[n,t]`  P    el. eff. at min. fuel flow w/o distr. heating
     :math:`\eta_{el,max,woDH}`      `Eta_el_max_woDH[n,t]`  P    el. eff. at max. fuel flow w/o distr. heating
     =============================== ======================= ==== =============================================
@@ -384,7 +384,7 @@ class GenericCHPBlock(ScalarBlock):
             expr = 0
             expr += -self.H_F[n, t]
             expr += n.alphas[0][t] * self.Y[n, t]
-            expr += n.alphas[1][t] * (self.P[n, t] + n.Beta[t] * self.Q[n, t])
+            expr += n.alphas[1][t] * (self.P[n, t] + n.beta[t] * self.Q[n, t])
             return expr == 0
 
         self.H_F_2 = Constraint(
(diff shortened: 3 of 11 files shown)


jnnr:
Will we drop *args and **kwargs entirely, in the end?

p-snft:
> @gnn: Will we be able to drop *args and **kwargs entirely, in the end? Answer: yes. Kwargs will be replaced by `options`

We (@joroeder and me) decided that `custom_attributes` is a more speaking name, as all keys given in this dict will be (custom) attributes of the class. We implemented this suggestion for the class `Flow`.

p-snft:
I am puzzled why I cannot make the other arguments of NonConvex explicit. For some reason the tests fail if I try.

jokochems:
> I am puzzled why I cannot make the other arguments of NonConvex explicit. For some reason the tests fail if I try. To my understanding, the following two code snippets should should create code that does the same thing:
> 
> ```python
> class MyClass
>     def __init__(
>         self,
>         foobar=None,
>         **kwargs,
>     ):
>         self.foobar = foobar
> ```
> 
> ```python
> class MyClass
>     def __init__(
>         self,
>         **kwargs,
>     ):
>         setattr(self, "foobar", kwargs.get("foobar"))
> ```
> 
> However, the creation of Pyomo constraints fails in the first case.

I agree that both should be the same. Does the problem still persist or have you found a fix? It seems that all tests except for Codacity and coverage are working fine. Was it only a problem on your local machine?

p-snft:
I did not commit because checks where failing locally. Probably, it was just a typo somewhere, that I did not find. I re-did exactly the same thing today (at least I feel like I did) and now everything works.

p-snft (approved):
The remaining errors are actually not new and I think they should not be fixed in this PR.

src/oemof/solph/_options.py
```diff
@@ -144,85 +145,56 @@ class NonConvex:
...
-        self._max_up_down = max_up_down
-
-    @property
-    def max_up_down(self):
-        """Compute or return the _max_up_down attribute."""
-        if self._max_up_down is None:
-            self._calculate_max_up_down()
 
-        return self._max_up_down
+        return max(self.minimum_uptime, self.minimum_downtime)
```
jokochems, line 200: The check has become obsolete since you are setting the default values to 0, right?
We don't want to allow the user for actively setting either one to None, do we?

src/oemof/solph/components/_generic_chp.py
```diff
@@ -101,36 +101,49 @@ class GenericCHP(network.Transformer):
...
-        self.electrical_output = kwargs.get("electrical_output")
-        self.heat_output = kwargs.get("heat_output")
-        self.Beta = solph_sequence(kwargs.get("Beta"))
-        self.back_pressure = kwargs.get("back_pressure")
+    def __init__(
+        self,
+        fuel_input,
+        electrical_output,
+        heat_output,
+        Beta,
```
jokochems, line 124: Weird that this is highlighted differently. It's not reserved, though, is it?

src/oemof/solph/components/_sink.py
```diff
@@ -13,18 +13,57 @@
...
+    >>> bel = solph.buses.Bus(label='electricity')
+
+    >>> electricity_export = solph.components.Sink(
+    ...    label='el_export',
+    ...    inputs={bel: solph.flows.Flow()})
+
+
+    Notes
+    -----
+    It is theoretically possible to use the Sink object with multiple outputs.
```
jokochems, line 45: It should be multiple inputs here, not outputs.

src/oemof/solph/_options.py
```diff
@@ -60,18 +60,19 @@ def __init__(
         existing=0,
         nonconvex=False,
         offset=0,
-        **kwargs,
+        custom_attributes=None,
     ):
-
+        if custom_attributes is None:
+            custom_attributes = {}
```
jokochems, line 66: I would phrase 
```python3 
if not custom_attributes:
    ...
```

src/oemof/solph/flows/_flow.py
```diff
@@ -133,148 +145,104 @@ def __init__(self, **kwargs):
...
-        need_nominal_value = [
-            "fix",
-            "full_load_time_max",
-            "full_load_time_min",
-            "max",
-            "min",
-            # --- BEGIN: To be removed for versions >= v0.6 ---
-            "summed_max",
-            "summed_min",
-            # --- END ---
```
jokochems, line 170: Why is this missing below? It shall be removed from v0.6 onwards.

src/oemof/solph/flows/experimental/_electrical_line.py
```diff
@@ -57,10 +57,28 @@ class ElectricalLine(Flow):
...
+            integer=kwargs.get("integer", False),
+            bidirectional=kwargs.get("bidirectiona", False),
+            investment=kwargs.get("investment"),
+            nonconvex=kwargs.get("nonconvex"),
+            custom_attributes=kwargs.get("costom_attributes"),
+        )
         self.reactance = solph_sequence(kwargs.get("reactance", 0.00001))
 
+        self.input = kwargs.get("input")
+        self.output = kwargs.get("output")
```
jokochems, line 80: This does not seem consistent with what the PR aims to do. But since it is an experimental component, I don't find it too bad.

jokochems (commented):
Thanks everyone for the strucutral improvements and sorry for the late response to this Pull Request. I left some comments, but nothing really important, though.

src/oemof/solph/flows/_flow.py
p-snft: I don't really get this comment, sorry.

src/oemof/solph/flows/experimental/_electrical_line.py
p-snft: I think, we did not adapt this. Also, kwargs can be handy for prototyping which might still be needed here.

src/oemof/solph/components/_sink.py
p-snft: Thanks, fixed.

src/oemof/solph/components/_generic_chp.py
p-snft: This is because the parameter is capitalised. According to our style guide, it should be lower case. I will adapt this.

src/oemof/solph/_options.py
p-snft: We still need the longer time of the two. However, I forgot to delete _max_up_down which is no longer used.

src/oemof/solph/_helpers.py
```diff
@@ -29,12 +31,16 @@ def check_node_object_for_missing_attribute(obj, attribute):
...
-        )
-        warn(
-            msg.format(attribute, obj.label, type(obj)),
-            debugging.SuspiciousUsageWarning,
-        )
+        warn_if_missing_attribute(obj, attribute)
+
+
+def warn_if_missing_attribute(obj, attribute):
+    msg = (
```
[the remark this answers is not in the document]
p-snft: Good point.

p-snft:
I accidentally merged some of this in #836, sorry.

PR #883, opened 2022-11-10
make kwargs explicit for component Transformer

DFuh:
remove kwargs from network transformer class to avoid misspelled kwargs


joroeder:
... continued in #891

PR #885, opened 2022-11-10
Make Sink kwargs explicit

jnnr:
Makes the arguments label and inputs for Sink explicit.

Open question: Can we drop the `check_node_object_for_missing_attribute`? Answer: Will have to be re-worked as soon as the check_for_missing_attributes function is refactored.


p-snft:
Despite the fact that "kwargs" is still exiting, I will merge this to continue in the other branch.

PR #886, opened 2022-11-10
Make kwargs explicit for Source component

srhbrnds:
This PR makes kwargs explicit for component Source.

### DIFF
--- a/src/oemof/solph/components/_source.py
+++ b/src/oemof/solph/components/_source.py
@@ -21,12 +21,12 @@ from oemof.solph._helpers import check_node_object_for_missing_attribute
 class Source(on.Source):
     """An object with one output flow."""
 
-    def __init__(self, *args, label=None, outputs=None, **kwargs):
+    def __init__(self, label=None, outputs=None, **kwargs):
 
         if outputs is None:
             outputs = {}
 
-        super().__init__(*args, label=label, outputs=outputs, **kwargs)
+        super().__init__(label=label, outputs=outputs, **kwargs)
         check_node_object_for_missing_attribute(self, "outputs")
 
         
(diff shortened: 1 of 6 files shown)


src/oemof/solph/components/_source.py
```diff
@@ -19,10 +19,42 @@
 
 
 class Source(on.Source):
-    """An object with one output flow."""
+    """A node object which is designed for one output flow.
```
jnnr, line 22:
```suggestion
    """A component which is designed for one output flow.
```

src/oemof/solph/components/_source.py
```diff
@@ -19,10 +19,42 @@
...
+    >>> bel = solph.buses.Bus(label='electricity')
+
+    >>> pv_plant = solph.components.Source(
+    ...    label='pp_pv',
+    ...    outputs={bel: solph.flows.Flow()})
+
+    >>> type(pv_plant)
+    <class 'oemof.solph.components._source.Source'>
+
+    >>> str(pv_plant.outputs()])
```
joroeder, line 44:
```suggestion
    >>> str(pv_plant.outputs())
```

src/oemof/solph/components/_source.py
jnnr: Just a suggestion to make it more specific

src/oemof/solph/components/_source.py
```diff
@@ -19,10 +19,42 @@
...
+    It is theoretically possible to use the Source object with multiple outputs.
+    However, we strongly recommend using multiple Source objects instead.
+    """
+
+    def __init__(self, label=None, outputs=None):
+        if outputs is None:
+            outputs = {}
+
+        super().__init__(label=label, outputs=outputs)
         check_node_object_for_missing_attribute(self, "outputs")
```
srhbrnds, line 58: Function needs to be replaced by refactored function that raises error message and should be triggered in if-loop (line 55)

srhbrnds:
- [ ] Run tox to check if tests are corrected @srhbrnds

Issue #723, opened 2020-09-15
Mainline "link"


p-snft:
For quite some time, there has been a "Link" component in "custom" that is meant for lossy transport of electricity between two buses (typically locations). With #720, the component has been enhanced to allow exclusive Flow in either one or the other direction.

Next, I want to mainline the component. The following is missing:

* [x] Documentation
* [x] Tests

It turned out that it might be helpful to have two approaches for different purposes:
1. `Binary`: This one has a binary constraint. The constraint `limit_active_flow_count` might be used, but a component might be easier.
2. `Signed`: This one would solve faster but allows flows at both sides. A `SignedFlow` might be used instead of the `Link`.

p-snft:
The documentation should carefuly explain why and under which conditions the link works. (Including the formulas.)

jnnr:
#754 Introduced two new constraints that force a direction to the links. This introduces a binary variable that represents the direction. This can lead to significant extra computation time. In some applications, this may be justified and useful. In others, it may be not desired. To avoid loop flows, small `variable_cost` can be set instead.

To conclude: The 2 extra equaions should be optional, e.g with a flag `unidirectional=True/False`.

Any ideas how to implement this in a clean way?

p-snft:
In https://github.com/oemof/oemof-solph/pull/838, @henhuy added the flag `limit_direction`. I agreed to not slow things down with fixing tabular and as the component is still experimental. Actually, I like `unidirectional` better.

henhuy:
Hey @minhki95,

this is exactly the same error, which led to implementation of #838 . With this implementation you should be able to add parameter `limit_direction=False` to your Link component. But maybe your model should work also with current setting?! not sure...

minhki95:
Thanks @henhuy it worked! So, what exactly is the reason that the run fails, if `limit_direction=True` ?
Or is it just in my case that it does? As far as I have understood if `limit_direction=False` does not add the 1st and 2nd inflow constraint to the program right?

p-snft:
> I saw that the Link component as it is right now is not working, which is also confirmed by an example that was working in version v0.4.4. I would be glad to help out in whatever capacity it is possible.

So, it worked in v0.4.4 but it is currently not working? I had a quick look and I assume that the problem is caused by the updated Pyomo version we use for v0.5. As there are no tests for the Link, yet, nobody noticed that it currently depends on the old version.

jnnr:
[...]

PR #888, opened 2022-11-10
Establish Link

### LINKED ISSUE #887: Move link from 'experimental' to established components
Link is useful, e.g. in oemof.tabular. Therefore, it should be part of the established components.

What has to be provided to do so?

- good docstrings
- tests
- working example
- higher subjective code quality requirements
- fix Link in `limit_direction=True` mode
- ...?

jnnr:
Addresses #723 by

* adding a constraint test with a checked lp-file
* removing non-functional binary variables and constraints
* [x] docstring
* [x] Fix tests
* [x] Add Changelog


jnnr:
@p-snft: Were shall I base this so that it will end up in the release 0.5.0? This is important for oemof.tabular's next release 0.4.0.

jnnr:
Seems that the CI constraint test for link fails with an empty lp-file. Could not reproduce this on my computer.

p-snft:
@jnnr: Didn't you include this in v0.4.5 already?

jnnr:
> @jnnr: Didn't you include this in v0.4.5 already?

Removing the obsolete constraints, yes. But not the rest.

p-snft:
I see.

nailend:
Hey @p-snft I took over on this PR and finished the last bits. Could you give me some advice on whats missing to include this in the next release? As @jnnr already mentioned, this PR is apparently necessary for our next oemof.tabular release. 
the
Would you be so kind to give me a review on this PR?
I adjusted the LP file, to the new multi-period notation. I just don't know why the representation of the constant changed in the LP-file (see my last [commit](https://github.com/oemof/oemof-solph/pull/888/commits/09867067ef4d4f2ef45cb8298e1d0541a8c620ad). Can can validate this change?

p-snft (changes requested):
I think there are three things missing:
1. Make arguments explicit. (We dropped kwargs, so that e.g. typos are easily caught.)
2. Test if warning is emitted when number of Flows is not matched.
3. Test if the ValueError in the Block class is risen.

src/oemof/solph/components/_link.py
```diff
@@ -71,14 +78,30 @@ class Link(on.Transformer):
...
+    ):
+        if inputs is None:
+            warn_if_missing_attribute(self, "inputs")
+            inputs = {}
+        if outputs is None:
+            warn_if_missing_attribute(self, "outputs")
+            outputs = {}
+        if conversion_factors is None:
+            warn_if_missing_attribute(self, "conversion_factors")
+            conversion_factors = {}
```
p-snft, line 96: Wouldn't it be easier to just require the arguments instead of having None as default? Whoever wants None can still give it manually.

nailend: I totally agree, especially as they are necessary in inputs. Thanks for the hint :)

src/oemof/solph/components/_link.py
```diff
@@ -74,14 +78,30 @@ class Link(on.Transformer):
...
```
p-snft, line 96:
```suggestion
        if inputs is None:
            inputs = {}
        if outputs is None:
            outputs = {}
        if conversion_factors is None:
            conversion_factors = {}
```
As this won't happen accidentally, I would not warn. Whoever sets it to Null on purpose, should be know what they are doing.

p-snft (commented):
For the value error: I you cannot trigger the error for a test, remove the code that is raising it.

p-snft (approved):
Looks good. Thank you!

nailend:
> Looks good. Thank you!

Thanks for you kind help!

PR #895, opened 2022-11-23
Upgrade OffsetConverter

sasa821:
### Explanation:

Based on the new structure of flows according to #826 and #856, the model for the `OffsetConverter` can be improved significantly. So far, this component could **ONLY** take the `NonConvex` attribute. This means it was assumed that the component had already been optimized and the nominal capacity was known. However, the component’s **real** efficiency curve and its (probably) less-efficient operation in part-load could not be considered in the optimization until now. 

Thanks to the new structure of the `Flow`, which enables the combination of `NonConvex` and `Investment` attributes, these parameters can now be part of the optimization. The following example shows that the results obtained from the previous optimization (applied to the `Converter`) and the new one (applied to the `OffsetConverter`) are considerably different.

Another change suggested in this PR is the definition of the `NonConvex` and `Investment` attributes. In the previous version of the `OffsetConverter`, the `NonConvex` attribute had to be defined for the **input flow**, but in the new model proposed in this PR, it must be defined for the **output flow** (see #861  for more information).

### Example:
As an example, the efficiency and load duration curves of a diesel genset are shown, which are obtained from different optimization methods. As explained before, previously, it was only possible to use the `Converter` for obtaining the optimal size of a component, but restructuring the flow provided the possibility to use the `OffsetConverter` for obtaining the optimal capacity.

offset-converter

As seen in the figures, the diesel genset has a minimum and maximum load of 20 and 100%. The `Converter` component takes a _constant efficiency_ as an input (here 33%), but the `OffsetConverter` component deals with the _efficiency curve_ of this component, which is obtained using the minimum and maximum loads and efficiencies. 

It can be seen that the optimal size of the diesel genset when the `OffsetConverter` is used is around **8% smaller**. The reason is that since this component is not operating efficiently at low loads, instead of having a larger capacity and operating at low loads, the optimizer tries to reduce the capacity so that it does not have to operate inefficiently at very low loads. This can be seen in the diagrams very clearly. On the right side, where the `Converter` is used to model the diesel genset, since the efficiency is assumed constant, in approximately 15% of the annual operation of this component, it operates at the minimum load. However, as seen on the left side, the `OffsetConverter` does not allow the diesel genset to operate at the minimum load.


sasa821:
Can anyone please help me with the `tox` errors? I have no idea how to fix them 🤔

fwitte:
Thanks for the super nice explanations! I am hacking the same Problem over at https://github.com/oemof/tespy. I have only found, that a recent pytest version is incompatible with tox in specific setups. See https://github.com/pytest-dev/pytest/issues/10420

fwitte:
#897 solves the issue. You can wait for that branch to be merged in dev and then merge dev in your branch :)

fwitte:
> https://github.com/oemof/oemof-solph/pull/897 solves the issue. You can wait for that branch to be merged in dev and then merge dev in your branch :)

Correction, the changes had already been implemented in dev, I just overlooked it. So merging the current dev should work already ;).

Bachibouzouk:
@p-snft - so as you mentionned in https://github.com/oemof/oemof-solph/pull/900#issuecomment-1401541680, we could make a prerelease with this PR ? Is there a place where the (pre) release protocol is documented? I looked across the README, CONTRIBUTING files as well as in the RTD and searched for "release" in the whole repo but could not find a protocol.

Should this PR still point toward `oemof:revision/rename_transformer_to_converter` ? I have time to allocate to push this work of @sasa821 into a pre-release, let me know how I can help there !

I plan to merge the latest stand of `dev` into the branch first and make sure the tests pass

p-snft:
There is no formal process to have a pre-release but I think it is common sense that they should not be inconsistent.

I think, this is easily achieved if we crate releases just from the version branch (e.g. `v0.5`) with the exception that alpha/dev releases can also be on dev. The exception can be necessary if the latest version branch is in feature freeze, so a pre-release for the version after that can be made.

The alternative would be to either have more version branches, like `v0.5.1` and `v0.5.2` (better `release/v0.5.1` to not be confused with the tags). But I think that's more than we currently need.

PS: Renaming Transformer to Converter is planned for v0.5.1, so the PR can keep it's target.

PR #904, opened 2023-01-07
Allow SinkDSM's delay_time to be an iterable for approach 'DSM'

### LINKED ISSUE #903: Allow for distinct delay times in SinkDSM component
For the `DLR` approach of the SinkDSM component, the `delay_time` attribute is passed as an integer value and converted to a range object - ultimately ending with a list of feasible delay times.

Here is an example:
```python3
import oemof.solph as solph
solph.components.experimental.SinkDSM(
     ...
    delay_time=10
)
```

Results in feasible delay times: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10].

Now while in principle, this is exactly what you want, it comes at the expense of introducing new variables and balancing variables as well as constraints for the different delay times within that list.

Again, in 99% of the cases, this is not really a problem. But if you want to optimise very large-scale energy systems and your model becomes very complex, this becomes a problem e.g. in terms of memory usage.

One easy enhancement to address this, would be to limit the feasible shifting times, e.g. for the example case only allow shifting times of [1, 2, 5, 10] or so.

This requires to allow for passing a list to the `delay_time` attribute. I'll self assign and add this feature in a pull request.

jokochems:
This enables to directly pass an iterable (usually) a list to the `delay_time` attribute of the experimental `SinkDSM`.
Instead of passing an integer, this allows to exclude some delay times which might be desired in order to limit model complexity which grows with longer delay times, see #903. 

Closes #903

### DIFF
--- a/src/oemof/solph/components/experimental/_sink_dsm.py
+++ b/src/oemof/solph/components/experimental/_sink_dsm.py
@@ -118,14 +118,15 @@ class SinkDSM(Sink):
         None.
         It's the interval in which between :math:`DSM_{t}^{up}` and
         :math:`DSM_{t}^{down}` have to be compensated.
-    delay_time: int
+    delay_time: int or iterable
         Only used when :attr:`~approach` is set to 'DIW' or 'DLR'. Otherwise,
-        can be None.
+        can be None. Iterable only allowed in case approach 'DLR' is used.
         Length of symmetrical time windows around :math:`t` in which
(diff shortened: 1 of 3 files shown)


jokochems:
I'll merge this pull request into the feature branch #828. The failing tests in my view do not relate to the changes introduced here, but are more general matters that need attention in the course of updating and finalizing #828. The newly introduced test passes and the feature works as expected.

Issue #906, opened 2023-01-18
GenericStorage is missing `custom_attributes` attribute


jnnr:
In #877, all components got `custom_attributes`, apart from GenericStorage.

p-snft:
That's because all custom attributes where also ignored before. So, if you mistyped something, nothing happened, even no attribute was added.

jnnr:
I don't see the difference. Before making kwargs explicit, GenericStorage and e.g. Sink just passed args and kwargs to the next class in the hierarchy. Now, kwargs have been dropped and for those cases where you want to set other attributes, you pass them via custom_attributes. How is GenericStorage different to e.g. Sink?

p-snft:
Well, `GenericStorage` did not pass it. Maybe that was not intentional, I just meant to not blame whoever did the explicit kwargs for `GenericStorage`. We can sure change that.

uvchik:
@p-snft is right. It was possible to add an Attribute to the `Flow` class but not to Nodes as `Transformer` or `GenericStorage`. I do not know why and I doubt that this was intended.

To make it easier to add additional constraints it should be possible to attributes to components as well but as it was not possible in previous versions it is a new feature so we can add it when it is finished in any version we want. In my opinion it does not block the v0.5.0 release.

Anyhow it could be more consistent in the announcements. "Now it is possible to add `custom_ attributes` to all Flow and Components" sounds better than "Now it is possible to add `custom_ attributes` to all Flow and most Components".

But I am fine with v0.5.1. Just poke me at the matrix chat, if you need a quick review.

p-snft:
I agree that we should be consistent. Current status is:

- `custom_attributes` exist for `Flow` and all components except ``GenericStorage`.
- `custom_attributes` add class attributes for instances of the `Flow`
- `custom_attributes` are ignored (literally nothing happens) for all components

For the flow, the keyword is (and always was) evaluated within solph.

I see two options:
1. Store the attributes is in `oemof.network`. This way, handling for Flow and Components could be done using a single implementation. We need to update network for that, having `custom_attributes` in solph is a preparation for that.
2. Implement the handling inside the solph classes.

I'd opt for the first option and accept that there will be a useless keyword for some time.

jnnr:
Thanks for giving this some thoughts!

Regarding the options: Currently, the handling for Flow uses option 2): set the attributes in the solph class. 

[...]

PR #908, opened 2023-01-24
Add custom_attributes to GenericStorage

p-snft:
Adds custom_attributes (dictionary) to GenericStorage in the same way, other components also have them.

Implements #906

### DIFF
--- a/src/oemof/solph/components/_generic_storage.py
+++ b/src/oemof/solph/components/_generic_storage.py
@@ -150,12 +150,20 @@ class GenericStorage(network.Node):
         fixed_losses_absolute=0,
         inflow_conversion_factor=1,
         outflow_conversion_factor=1,
+        custom_attributes=None,
     ):
         if inputs is None:
             inputs = {}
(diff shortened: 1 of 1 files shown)


p-snft:
@jnnr: As far as I see, `custom_attributes` is ignored for all components. Could you double-check?

jnnr:
> @jnnr: As far as I see, `custom_attributes` is ignored for all components. Could you double-check?

Ignored in the sense that nothing happens, or that they cannot be found among the attributes of an instance?

p-snft:
`GenericStorage is derived from `network.Node` and passes kwargs to this paren class. However, `network.Node` just considers the kwargs "label", "inputs", and "outputs", every other entry to kwargs s just dropped. The solph casses `Sink`, `Source`, and `Transformer` are derived from the corresponding classes in network but those just pass things to `network.Node`.

PS: I checked those things for v0.4, to make sure it was not just removed.

jnnr (approved):

PR #909, opened 2023-01-24
Merge v0.4 in v0.5

jnnr:


p-snft:
Unfortunately, I do not see a way to do the merge using this workflow, as it includes putting v0.5 stuff into the v0.4 branch. I will do it offline.

PR #910, opened 2023-01-24
Revision/upstream v04 v05

p-snft:
Upstream changes made for v0.4 to v0.5

Replaces #909

PR #911, opened 2023-01-24
Revision/upstream v04 v05

p-snft:
Upstream changes from v0.4 to v0.5.

Replaces #909, #910


p-snft:
Checks are failing because Codacy does not run at all. As the changes are coming from a stable release, they should also be in the next one. Could anyone double-check that I did not do anything stupid?

p-snft:
Made obsolete by merging dev into v0.5.

PR #912, opened 2023-01-30
Enable modelling of cellular grid structures

lensum:
## What it is about

I'm trying to enable the modelling (and subsequent optimization) of cellular grid structures in oemof-solph. The cellular grid is prominently supported/proposed by the VDE-ETG. It is similar to Multi-Microgrid structures, where the microgrids are allowed to exchange energy with each other, but in a cellular grid each cell can contain sub-cells, which can contain sub-cells themselves (and so on). So there are more "levels" possible than in a multi-microgrid structure.

As the structure of a multi-microgrid / low-level cellular grid offers itself to be disaggregated, I'd like to enable a method to solve the resulting problem in a distributed way (e.g. Dantzig-Wolfe). As of yet, this is not part of the pull request and I have little knowledge and no experience with distributed optimization.

## What functionality does it implement?

The overall functionality is described above, but I will try and go into detail a little more by explaining the classes I added. The usage is also displayed in the new example called `cellular`.

### Model

**Edit 2023-05-18**: The model is now capable to handle cellular structures. Minor changes were necessary, mainly in the way how the `flows` and `nodes` are collected from the energysystem. A dict can now be passed to the `Model`-object under the `energysystem` keyword. This is the indication, that a cellular structure is to be optimized. **Hint**: There is still one large pyomo Model build for the whole cellular energy system. Not one per cell. This will change when and if I get the Dantzig-Wolfe reformulation working.

**Update 2023-06-22**: As pointed out in [the comments below](https://github.com/oemof/oemof-solph/pull/912#pullrequestreview-1462513906), a `list` will be used instead of a `dict` to pass the cells to the `Model`-class.

## Other comments and questions

- I have not yet written any tests and documentation is consistently lacking. 
- I am woefully unexperienced when it comes to open source collaboration, but I'm trying my best. Please make me aware of any misbehaviour of mine.
- Lastly: Feedback is highly welcome! I started this with an "stop thinking and make it work" attitude, which might not be the best approach (**Edit 2023-05-18**: Indeed, it was not :D ). It was the only viable option, as I don't have much knowledge about e.g. `oemof-network` and limited time on my hands. It might very well be, that there are simpler, better and more user-friendly ways to implement this. If you know them, I appreciate you pointing them out to me.

## Related Issues

I know that https://github.com/oemof/oemof-solph/pull/840 was working on something similar, but to the best of my knowledge the development stopped.


lensum:
In case someone follows this PR, I want to provide a short status report.

The example given in examples\cellular\cellular.py is working, as far as I have tested. More extensive testing still needs to be done and I anticipate a few problems:

1) Right now, **the implementation does not allow simultaneous energy im- and export for cells**. I assumed this to be a necessary constraint, initially, however when looking at multi-level cell structures, it became clear to me that this will not work (i.e.: if `cell_b` and `cell_c` are both connected as "children" to `cell_a` and energy shall be moved from `cell_b` to `cell_c`, `cell_a` needs to simultaneously import and export energy). This should be easily fixable within the `CellConnectorBlock` class. This will probably also affect issue 2)
2) The implementation of **`max_power` in the `CellConnector` class is redundant**, as the flows connecting the `CellConnector` with the buses also have the ability to restrict the maximum power via `max`-Parameter. This could lead to inconsistent parametriziation and should be alleviated by changing the `_max_input_rule` and `_max_output_rule` in the `CellConnectorBlock` class.
3) The **usage of CellConnector objects is way too complicated**. Two `CellConnector` objects and four `Flow`s is just too much hassle. There has got to be an easier way to do this. My next attempt will be to somehow connect the cells with automatically generated `CellConnector` objects, which will be part of the upmost cell. The specifics are still hazy to me as well, but I hope to reduce the number of `CellConnector` objects with this measure to one per connection, instead of the current two and the necessary Flows maybe to two (maybe with the `bidirectional` functionality, which I didn't use so far).

Also, there are a lot of more or less current #TODO in the code, which I hope to resolve soon.

If anyone has any questions or recommendations, feel free to reach out! I plan on atteding the upcoming oemof-meeting, so discussion in person is also possible.

p-snft:
Hi @lensum,

I'm not really into the topic, so please please forgive me my lack of understanding. I have a few questions:

- What are the advantages of modelling a grid as a cellular structure vs. having everything in one energy system without cells? Won't the results be the same?
- If flows can go into both directions at the same time, what is the difference between a `CellConnector` and a `Bus`? If the former has a limited capacity, then why it is not the `Link`?

If this is mostly about data analysis, we might just think about having something like balancing groups in post-processing.

lensum:
Hey again @p-snft ,
thank you very much for your feedback! I sat down and implemented the changes and it significantly reduced my "footprint" on the code, which I appreciate. It's a much better status to report on on the upcoming meeting.

I also updated the initial text. If I should just delete the old text instead (for a cleaner look), feel free to point that out to me.

src/oemof/solph/_models.py
```diff
@@ -110,7 +114,13 @@ def __init__(self, energysystem, **kwargs):
         # ########################  Arguments #################################
 
         self.name = kwargs.get("name", type(self).__name__)
-        self.es = energysystem
+        self.iscellular = isinstance(energysystem, dict)
+
+        if self.iscellular:
+            self.es = list(energysystem.keys())[0]
+            self.ec = energysystem[self.es]
```
jokochems, line 121: I think since you cast to a list anyways and the dict has only one key, you might make it a list rightaways and declare the 0th element to be your "master" energy system.

src/oemof/solph/_models.py
```diff
@@ -129,6 +139,18 @@ def __init__(self, energysystem, **kwargs):
...
+                for io, f in cell.flows().items():
+                    if io not in self.flows.keys():
+                        self.flows.update({io: f})
+                    else:
+                        msg = (
+                            "Two flows with identical input-output are tried "
+                            "to be added to the Model. Second flow is skipped."
+                            "\n{}"
+                        )
+                        raise Warning(msg.format({io: f}))
```
jokochems, line 153: I'm not so sure if raising an error is enough or whether it should be an error.

src/oemof/solph/_models.py
```diff
@@ -352,14 +392,34 @@ def __init__(self, energysystem, discount_rate=None, **kwargs):
...
+                for node in cell.nodes:
+                    if node not in self.nodes:
+                        self.nodes.append(node)
+                    else:
+                        msg = (
+                            "Two nodes of identical name are tried to be added"
+                            " to `CellularModel.nodes`. Second node is"
+                            " skipped.\n{}"
+                        )
+                        raise Warning(msg.format(node))
```
jokochems, line 417: Same as above: Ìs it only a warning or even an error here?

src/oemof/solph/_models.py
```diff
@@ -446,29 +506,38 @@ def _add_parent_block_variables(self):
         for o, i in self.FLOWS:
             if self.flows[o, i].nominal_value is not None:
                 if self.flows[o, i].fix[self.TIMESTEPS.at(1)] is not None:
+                    # if nominal_value and fix is given,
+                    # set flow to fix * nominal value
```
jokochems, line 510: I think removing these comments again would imrove code quality.

src/oemof/solph/_models.py
```diff
@@ -352,14 +392,34 @@ def __init__(self, energysystem, discount_rate=None, **kwargs):
...
 
     def _add_parent_block_sets(self):
         """Add all basic sets to the model, i.e. NODES, TIMESTEPS and FLOWS.
         Also create sets PERIODS and TIMEINDEX used for multi-period models.
         """
-        # set with all nodes
-        self.NODES = po.Set(initialize=[n for n in self.es.nodes])
+        if self.iscellular:
+            # collect all nodes from the child cells
+            self.nodes = self.es.nodes
```
jokochems, line 406: I guess, you could move this line above the if condition (line 403) and then you would not need an else statement, would you?

jokochems (commented):
Hi @lensum,

sorry, was meaning to look at this feature earlier and really liked your presentation at the user & dev meeting as well as the idea to allow for a decomposition of oemof.solph models by slightly re-arranging / ordering things.

Please find some comments of mine to dedicated code sections. Overall,
* I think, you could facilitate even a bit more (by using a list instead a dict for the energy_system) and
* I'm a bit sceptical towards adding a facade without any additional functionality inside oemof.solph.

Also, I think, simply dropping the sections that are crossed out in the original post would facilitate its readability.

src/oemof/solph/_models.py
lensum: Good point. I'm changing it to be a ValueError.

src/oemof/solph/_models.py
lensum: Yes, you are correct. Changed that!

src/oemof/solph/components/experimental/_energy_cell.py
```diff
@@ -0,0 +1,14 @@
...
+class EnergyCell(EnergySystem):
+    """
+    This is a facade for an EnergySystem. It does not provide any additional
+    functionality and it is used purely to prevent misconceptions.
+
+    This also means, that every energy system can be regarded as an energy cell.
+    """
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
```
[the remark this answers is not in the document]
lensum: It's removed in the latest commits!

lensum:
Hey @jokochems,

I implemented most of the suggested changes. Thanks a ton! I'm now working on satisfying the checks before I add tests.
Would it be okay to ask you for a review, once I'm done?

lensum:
`coveralls` complains about decreased coverage, but the lines that aren't covered weren't written by me, as far as I can see. Should I provide tests for those lines regardless?

lensum:
All [requirements for a pull request known to me](https://oemof-solph.readthedocs.io/en/latest/contributing.html#pull-request-guidelines) are fulfilled, therefore I'm removing the "Draft" status.

Also, I took the liberty to change the title of the documentation from "Welcome to oemof's documentation" to "Welcome to oemof solph's documentation". I did this because I seem to remember that the term "oemof" should no longer be used synonymously with "solph" and this rather prominent display bothered be somehow.

src/oemof/solph/_models.py
p-snft: I agree that comments should not just retell self-explanatory code. If it does not help explaining it, remove the comment.

lensum: Okay, the latest commit removes these comments.

jokochems:
> Hey @jokochems,
> 
> I implemented most of the suggested changes. Thanks a ton! I'm now working on satisfying the checks before I add tests. Would it be okay to ask you for a review, once I'm done?

Sure, I can have a look. But I'll leave the final decisions to @p-snft.

docs/index.rst
```diff
@@ -3,9 +3,9 @@
    You can adapt this file completely to your liking, but it should at least
    contain the root `toctree` directive.
 
-=================================
-Welcome to oemof's documentation!
-=================================
+=======================================
+Welcome to oemof solph's documentation!
```
jokochems, line 7: Here it should be oemof-solph, right? So the hyphen needs to be added.
In general, I like your idea of making it clear in the docs that this is oemof-solph, not oemof.

docs/usage.rst
```diff
@@ -1265,6 +1265,119 @@ Besides the `invest` variable, new variables are introduced as well. These are:
       monthly periods, but you would need to be very careful in parameterizing your energy system and your model and also,
       this would mean monthly discounting (if applicable) as well as specifying your plants lifetimes in months.
 
+Modelling cellular energy systems
```
jokochems, line 1268: You could make it more general by adding " and modularizing energy system models".

docs/usage.rst
```diff
@@ -1265,6 +1265,119 @@ Besides the `invest` variable, new variables are introduced as well. These are:
...
+---------------------------------
+
+The cellular approach is a concept proposed by the VDE-ETG. It is related to smart-grids and multi-microgrid systems but extends
```
jokochems, line 1271: You could quote the VDE publication that introduces and explains the concept.

docs/usage.rst
```diff
@@ -1265,6 +1265,119 @@ Besides the `invest` variable, new variables are introduced as well. These are:
...
+both. The idea is to group the components of an energy system into a hierarchically aggregating structure of cells. For example,
+the sources, sinks, storages and converters of a household could be a single cell. Then a group of physically neighboring 
+households could form another cell, made up of household-cells. This behaviour can be scaled up. The real game-changer in the
```
jokochems, line 1274: better: "consisting of" instead of "made up of"

docs/usage.rst
```diff
@@ -1265,6 +1265,119 @@ Besides the `invest` variable, new variables are introduced as well. These are:
...
+
+The cellular approach is a concept proposed by the VDE-ETG. It is related to smart-grids and multi-microgrid systems but extends
+both. The idea is to group the components of an energy system into a hierarchically aggregating structure of cells. For example,
+the sources, sinks, storages and converters of a household could be a single cell. Then a group of physically neighboring 
+households could form another cell, made up of household-cells. This behaviour can be scaled up. The real game-changer in the 
+cellular approach is the way the cells are operated, which will not be covered here. Here, we focus on the way such cellular
+energy systems can be modeled.
+
+So far, the implementation in solph is just a neat way to group different parts of a larger energy system into cells. Further 
+functionality is expected to make its way into solph in the future, e.g. a distributed optimization procedure based on
```
jokochems, line 1279: You could mention that the structural updates are a prerequisite for most (or even any?) decomposition algorithm. I think, Bender's decomposition for instance would also be allowed for given the restructuring.

docs/usage.rst
```diff
@@ -1265,6 +1265,119 @@ Besides the `invest` variable, new variables are introduced as well. These are:
...
+
+Now we can go on and add components to the energy cells just like we do with regular energy systems.
+
+.. code-block:: python
+
+    bus_el_es = buses.Bus(label="bus_el_es")
+    es.add(bus_el_es)
+
+    bus_el_ec_1 = buses.Bus(label="bus_el_ec_1")
+    sink_el_ec_1 = cmp.Sink(
```
jokochems, line 1308: I think there is no standard import alias, is there? Thus I would replace `cmp` by `solph.components` and also place the import statement `import oemof.solph as solph` above to make it clear.

src/oemof/solph/_models.py
```diff
@@ -352,14 +372,35 @@ def __init__(self, energysystem, discount_rate=None, **kwargs):
...
+            and energysystem[0].periods is not None
+        ):
+            self.discount_rate = 0.02
+            msg = (
+                f"By default, a discount_rate of {self.discount_rate} "
+                f"is used for a multi-period model. "
+                f"If you want to use another value, "
+                f"you have to specify the `discount_rate` attribute."
+            )
+            warnings.warn(msg, debugging.SuspiciousUsageWarning)
```
jokochems, line 386: If I'm not mistaken, this is the same as above. Couldn't we add a tiny method for this that we call twice to avoid code duplication?

examples/cellular/cellular.py
```diff
@@ -0,0 +1,304 @@
+"""
+General description
+-------------------
+
+Cellular energy systems are proposed by the VDE-ETG. They are, as the name
+implies, energy systems that are made up of cells. Each cell can contain
```
jokochems, line 6: See above "consist of"

jokochems (approved):
Hello @lensum,

to me everything looks good. I think, the tiny decrease in coverage can be ignored.
I left some minor comments in the docs with suggestions for slight rephrasing or additions. Also, I think introducing a new method in `_models.Model` could help to avoid code duplication. 

Overall, these are all optional aspects. I would be fine with it as it currently is and leave it up to @p-snft to decide.

Thanks a lot! Looking forward to your upcoming decomposition activities.

docs/index.rst
```diff
@@ -3,9 +3,9 @@
    You can adapt this file completely to your liking, but it should at least
    contain the root `toctree` directive.
 
-=================================
-Welcome to oemof's documentation!
-=================================
+=======================================
+Welcome to oemof-solph's documentation!
```
p-snft, line 7: Just a note on development style: Changes like this should be made in an independent branch to be able to merge them before there is a decision about the feature branch. (And I have to confess that I am not always strict although I know better.)

examples/cellular/cellular.py
```diff
@@ -0,0 +1,304 @@
...
+    )
+
+    ec_1.add(source_el_ec_1)
+    ec_2.add(source_el_ec_2)
+    ec_1_1.add(source_el_ec_1_1)
+    ec_1_2.add(source_el_ec_1_2)
+    ec_2_1.add(source_el_ec_2_1)
+    ec_2_2.add(source_el_ec_2_2)
+
+    connector_el_ec_1 = buses.Bus(
```
p-snft, line 195: It might be cleaner to use the ’Link’ here, as it does not allow the flow to return into its original cell. (It still can if the loop is closed in the other cell.)

src/oemof/solph/_models.py
```diff
@@ -44,8 +45,11 @@ class BaseModel(po.ConcreteModel):
 
     Parameters
     ----------
-    energysystem : EnergySystem object
-        Object that holds the nodes of an oemof energy system graph
+    energysystem : EnergySystem object or list of EnergySystem objects
+        Object that holds the nodes of an oemof energy system graph.
+        If a list is passed, a cellular structure is assumed and the
+        first element needs to be the upmost energy cell (structurally
+        containing all other cells).
```
p-snft, line 52: Maybe, it would be better to allow to add an `EnergySystem` to an existing `EnergySystem`. For convenience, the one to be added could be a derived version called `EnergyCell`. When you add stuff to these, the labels will be automatically prefixed by the label of the cell. So, it was easier to create cells without having to think about global names.

p-snft (commented):
I like the Idea but I think there is room for improvement. As we introduce new options in the API, I feel like these should be discussed first.

docs/index.rst
lensum: Yeah, I thought the same while making this change and then went with it anyways. I opened a separate PR for it (#952).

src/oemof/solph/_models.py
lensum: That was my initial attempt as well. However, @gnn warned me at the oemof meeting 2022.11 that this is basically the implementation of sub-graphs, which (for reasons I must admit I have forgotten) is a rather complicated thing. Also, I struggled with the recursive looping through the `EnergySystem`/`EnergyCell` instances to get all the nodes and flows. I vaguely remember that the difficulty was to loop through them in such a way, that when adding a flow which connects a "parent" and a "child" cell, the component of the child cell needs to be known to the `Model` instance already. For some reason, I couldn't find a way to make that work without creating errors. However, that was before I switched to the `Bus` class for connections so that might not be an issue anymore.

Anyways, I don't think that it's such a hassle to think of global names to label the `EnergySystem` instances. I think the example shows nicely how it can be done. Also, I do really like the simplicity of this presented approach. Yes, it requires a tiny bit of abstraction when modelling cellular energy systems, but in exchange there aren't many changes to the code and the maintenance should be rather simple. On the other hand, I don't really see a benefit from the nested `EnergySystem`/`EnergyCell` approach.

p-snft: I see. What if we mark the option to have a list as experimental? (E.g., I'd like to have the structure including global names for automatic generation of cells/ sub-graphs in MTRESS and thus have to look into that anyway.)

lensum: I'd be fine with that! If/when the sub-graph functionality makes its way into solph the cellular feature could also be reworked to exploit that functionality.

p-snft: I don't think anybody would give a list accidentally, so it should be enough to make a statement in the documentation.

lensum: It has a `(experimental)` hint attached now.

src/oemof/solph/_models.py
```diff
@@ -45,7 +45,7 @@ class BaseModel(po.ConcreteModel):
 
     Parameters
     ----------
-    energysystem : EnergySystem object or list of EnergySystem objects
+    energysystem : EnergySystem object or list of EnergySystem objects (experimental)
```
p-snft:
```suggestion
    energysystem : EnergySystem object or (experimental) list of EnergySystem objects
```

Maybe this way, to highlight that just the second option is experimental?

lensum: I'm fine with that! It's easier to understand, indeed. I needed to shorten the line, so I deleted the `of EnergySystem objects` after `list`. It now reads `energysystem : EnergySystem object or list (experimental)` and I would change it to `energysystem : EnergySystem object or (experimental) list`. The explanation that the list needs to hold `EnergySystem` objects moved into the text below. 
Is that okay with you as well?

I apologize for the cumbersome discription.

p-snft: I see. The line `energysystem : EnergySystem object or (experimental) list` sounds reasonable. Maybe also add the keyword in the description, so it reads:

```
    energysystem : EnergySystem object or (experimental) list
        Object that holds the nodes of an oemof energy system graph.
        Experimental: If a list  is passed, the list needs to hold
        EnergySystem objects and a cellular structure is assumed.
        In this case, the first element needs to be the upmost energy cell
        (structurally containing all other cells).
```

lensum: Good idea! The latest commit includes this change.

PR #913, opened 2023-02-02
Remove upper bound for Numpy

### LINKED ISSUE #905: Numpy >= 1.24 is not supported
With numpy, 1.24 the way we use tuples as indexes is no longer supported. I don't know the details, but some unit-tests fail.

# How to reproduce

* Install solph with numpy >= 1.24
* Run the unit tests.
* Some will fail.

p-snft:
Fixes #905

I have the feeling that the problems have been solved in Numpy/Pandas.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -64,7 +64,11 @@ nitpicky = False
 
 exclude_patterns = ["_build", "whatsnew/*"]
 
-linkcheck_ignore = [r"https://requires.io/.*", r"https://matrix.to/*"] + (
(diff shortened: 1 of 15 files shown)


p-snft:
Don't know why one black fails but the other passes. Also, nothing was changed in the code.

uvchik (approved):
Looks fine to me. I agree that the bug was fixed upstream. At least everything worked as expected. I just fixed some style issues and removed the `DeprecatedWarning` from Pyomo but nothing essential.

PR #914, opened 2023-02-02
Revision/remove nose

p-snft:
Support for tests written for nose is deprecated in pytest[1]. Thus, we should go fully pytest and remove the legacy nose stuff. In particular, many features of nose just "save a few characters", i.e by typing `eq_(a, b)` instead of `assert a == b`. 

1: https://docs.pytest.org/en/stable/deprecations.html#support-for-tests-written-for-nose

### DIFF
--- a/tests/constraint_tests.py
+++ b/tests/constraint_tests.py
@@ -121,7 +121,7 @@ class TestsConstraint:
                 expected = normalize_to_positive_results(expected)
                 generated = normalize_to_positive_results(generated)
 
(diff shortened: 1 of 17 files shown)


p-snft:
Same Black errors we got everywhere (fixed in #913). Removal of Nose is ready for review.

uvchik (approved):
I did not check every single line but..
... the coverage has increased
... all tests are fine
... nothing but test have been changed

So I think this can be safely merged.

PR #916, opened 2023-02-03
v0.5

p-snft:
Release v0.5.0


p-snft:
That particular test is expected to fail because it references to the released version already.

Issue #894, opened 2022-11-15
Discontinue kwarg "investment"


p-snft:
At the moment, we have two mutually exclusive keyword arguments, `nominal_value` and `investment`. IMHO, it would be more intuitive to just give the investment object to the `nominal_value`.

PR #917, opened 2023-02-07
Revision/investment api

p-snft:
We had two mutually exclusive keyword arguments `nominal_value` and `investment`, that both define the size (capacity or storage capacity). With this API change, `nominal_value` (of a `Flow`) and `nominal_storage_capacity` (`GenericStorage`) accept an `Investment` object, and the argument `investment` is marked as deprecated.

Implements #894


jokochems (approved):
I introduced a minor editorial change to the Whatsnew file, but besides am happy to approve it. Thank you for improving the structure by removing a potential source of errors

p-snft:
Tests fail because of changed order of lines in LP files that were not touched. I will merge anyway.

PR #919, opened 2023-02-09
Let GenericStorage inherit from network.Component

jnnr:
GenericStorage is a component. Its definition resides in the module components. However, it inherits directly from Node.

This PR fixes that by letting GenericStorage inherit from network.Component.

I checked: All other components do inherit in some way from Component.


p-snft (approved):
Great, we're fixing all those old inconsistencies now.

Issue #926, opened 2023-03-06
Add SCIP solver to open source solver


ckaldemeyer:
Hi everybody,

I just wanted to mention that the [SCIP solver](http://scipopt.org) is licensed under the Apache 2 license from version `8.0.3` onwards and can be used to solve `LP/MIP` problems.

My experiences are so far very good in terms of usability and performance when compared to `GLPK` and `CBC` and in the newer versions, it can be used easily from Pyomo using AMPL bindungs, which I could successfully test under Linux and Windows 10. See my related question here: http://listserv.zib.de/pipermail/scip/2023-March/004601.html

If you think, this provides some value and your tests are positive too, I would offer to add some documentation on how to install it on Linux und Windows.

Best wishes,
Cord

ckaldemeyer:
Here you can find a performance benchmark (compared to Gurobi 8, CBC and GLPK with SCIP version 7): https://www.gurobi.com/wp-content/uploads/Gurobi_8_Performance_Benchmarks.pdf?x58432

p-snft:
I did not test it, yet, as I did not know that it has a usable license now. Do you think it's possible to have an hands-on at the next user meeting? (Currently, I doubt that my callendar allows to do this without tutoring or an explicit explanation.)

ckaldemeyer:
I won't be on the next user meeting as I am on parental leave at the moment. Wouldn't it be sufficient to test the current examples and optionally compare the results?

ckaldemeyer:
https://github.com/oemof/oemof-solph/issues/855 is closely connected and might be integrated in the same PR -> see PR

PR #927, opened 2023-03-13
Add SCIP open source solver

ckaldemeyer:
*   Describe your pull request as transparent as possible
    * This PR adds instructions to install the [SCIP solver](https://scipopt.org) on Windows and Linux machines
    * `SCIP` is able to solve LP and MIP models and seems to be overall faster than `CBC` and `glpk`
    * Moreover, it adds the solver to the installation test routines
*   Related issues?
  * https://github.com/oemof/oemof-solph/issues/926
*   Share your knowledge: Insights/Remarks
  * So far, my experiences using the solver are very good
  * Nonetheless, it would be nice, if we could do some more tests
  * These should include different operating systems such as Windows, Linux and OSX
*   Other comments and questions
*   For new features: Remember the documentation!


ckaldemeyer:
I would suggest to document the installation here if we agree, that we want the solver to be included: https://oemof-solph.readthedocs.io/en/latest/readme.html#installing-a-solver 

My first idea would be to split the documentation into different sections for each solver. What do you think?

ckaldemeyer:
I just noticed that the links to the CBC downloads for Windows from the AMPL site (e. g. https://ampl.com/dl/open/cbc/cbc-win64.zip) are broken and require a registration.

After a registration, the solver downloads can be accessed from [this page](https://ampl.com/products/solvers/open-source-solvers/) which links to the [download page](https://portal.ampl.com/user/ampl/download/list) which also contains the downloads for the highs solver. We might just add this, too.

p-snft:
Installation on more recent Linux systems is a bit cumbersome. There are packages for an old Ubuntu (which is still supported, though) and Debian oldstable. I tried compiling on my own but some of the dependencies are also not available in the package manager and have to be installed manually. I'll try the solver using an not up-to-date Linux system and would opt for unofficial if it works. (IMHO, we cannot document the installation process for this one.)

fwitte:
It does work now, the problem was, that the `oemof_installation_test` was broken (no components were added the the `EnergySystem`, which is no problem for `lp` `solver_io`. `scip` uses `nl`, which cannot handle an empty problem.

fwitte:
@ckaldemeyer: Sorry (not really) for just pushing into your branch, but I thought it would be nice to get it off the table :)

PR #931, opened 2023-03-27
Fix creation of time intervals from time index

p-snft:
The method implemented here seems to be more robust and is probably also more comprehensible.

Side note: I had a problem when creating the needed time intervals from the index of a DataFrame, where the original version did not work. The new implementation now works.


jokochems:
Looks good to me.

PR #935, opened 2023-05-04
Use list to describe periods

p-snft:
A list should easier to handle compared to a dictionary using integers counting from 0 as keys.


jokochems (approved):
Thank you very much, @p-snft. I must admit that I did not see that changing it to a list attribute would be "so easy" while at the same time further simplifying both code and usage. I'm happy to approve this. Imho, you can go ahead and merge it.

PR #936, opened 2023-05-15
Add storage_level_constraint

p-snft:
- [X] constraint that allows to (in) activate flows from/to a storage based on storage level
- [X] Example
- [X] Documentation (basic)
- [x] Tests

**From the example**

```Python
storage_level_constraint(
    model=model,
    name="multiplexer",
    storage_component=storage,
    multiplexer_bus=multiplexer,
    input_levels={in_1: 1 / 3},  # in_0 is always active
    output_levels={out_0: 1 / 6, out_1: 1 / 2},
)
```

There are two inflows (one always active, one active if storage level < 1/3) and two outflows (one active if storage level > 1/6 the second if storage level > 1/2). There is revenue for the outputs, the constraint inflow is free, the unconstraint one costs more than the revenue.

grafik

As the storage has losses, the outflows are used to empty it as fast as possible. It is held above the limit for the free inflow so that it can be used to recharge in the end. As the free inflow is constraint, the costly has to kick in in the end to fully recharge the storage.

### DIFF
--- a/src/oemof/solph/constraints/storage_level.py
+++ b/src/oemof/solph/constraints/storage_level.py
@@ -15,8 +15,8 @@ def storage_level_constraint(
     name,
     storage_component,
     multiplexer_bus,
-    input_levels=None,
-    output_levels=None,
+    input_levels,
+    output_levels,
 ):
     r"""
     Add constraints to implement storage content based access.
@@ -44,10 +44,6 @@ def storage_level_constraint(
 
     Verbose description can be found in https://arxiv.org/abs/2211.14080
     """
-    if input_levels is None:
-        input_levels = {}
-    if output_levels is None:
-        output_levels = {}
 
     def _outputs():
         OUTPUTS = po.Set(initialize=output_levels.keys())
--- a/src/oemof/solph/constraints/storage_level.py
+++ b/src/oemof/solph/constraints/storage_level.py
@@ -124,17 +124,17 @@ def storage_level_constraint(
                 \hat{y}_n \ge (E(t) - E_n) / E_{max}
             """
             for t in m.TIMESTEPS:
-                for o in input_levels:
+                for i in input_levels:
                     getattr(m, constraint_name).add(
-                        (o, t),
+                        (i, t),
                         (
                             m.GenericStorageBlock.storage_content[
                                 storage_component, t
                             ]
                             / storage_component.nominal_storage_capacity
-                            - input_levels[o]
+                            - input_levels[i]
                         )
-                        <= inactive_input[o, t],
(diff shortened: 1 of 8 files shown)


p-snft:
Coveralls lies: I added a file with 100 % coverage and it lead to a decreased coverage. Cannot be!

jokochems:
For the devs: I'll add my review.

src/oemof/solph/constraints/storage_level.py
```diff
@@ -0,0 +1,177 @@
...
+                <= active_output[o, t]
+            )
+
+        setattr(
+            model,
+            f"{name}_output_constraint",
+            po.Constraint(
+                OUTPUTS,
+                model.PERIODS,
+                model.TIMESTEPS,
```
jokochems, line 104: Couldn't you instead write model.TIMEINDEX here? This is what is used fo other flows.

src/oemof/solph/constraints/storage_level.py
```diff
@@ -0,0 +1,177 @@
...
+                        (
+                            m.GenericStorageBlock.storage_content[
+                                storage_component, t
+                            ]
+                            - input_levels[o]
+                        )
+                        / storage_component.nominal_storage_capacity
+                        <= inactive_input[o, t],
+                    )
+
```
jokochems: Formulation is different to the one in the docstring. This seems to need some fixing. Also, normalization looks somewhat odd.

src/oemof/solph/constraints/storage_level.py
```diff
@@ -0,0 +1,177 @@
...
+                INPUTS,
+                model.PERIODS,
+                model.TIMESTEPS,
+                rule=_constraint_input_rule,
+            ),
+        )
+
+    _inputs()
+
+    return
```
jokochems, line 177: Can't this be dropped? If I'm not mistaken this is the same as `return None` which has no effect, right?

jokochems (changes requested):
I do not understand the use case for this, but nevermind, I reviewed from a code and constraints point of view. There, one constraint formulation looks odd and is not what the docstrings state. Maybe you have a look and fix this.

src/oemof/solph/constraints/storage_level.py
p-snft: Actually, both formulations were wrong. I did an error copying the formula from the paper.

jokochems (approved):

p-snft:
The tests fail because of a different order of lines in the constraint test LP files which I did not touch. So, I allow myself to merge.

PR #942, opened 2023-05-26
Fix SinkDSM load shedding limit

jokochems:
This PR fixes one constraint formulation in the `SinkDSMDLRInvestmentBlock`. I was meaning to impose a periodical limit on shedding, but forgot to adjust it, resulting in an overall limit that did not make any sense. Now it is fixed.

Please review, @p-snft and if okay, feel free to merge back.

### DIFF
--- a/src/oemof/solph/components/experimental/_sink_dsm.py
+++ b/src/oemof/solph/components/experimental/_sink_dsm.py
@@ -4859,7 +4859,11 @@ class SinkDSMDLRInvestmentBlock(ScalarBlock):
                 for p in m.PERIODS:
                     if g.shed_eligibility:
                         # sum of all load reductions
-                        lhs = sum(self.dsm_do_shed[g, t] for t in m.TIMESTEPS)
+                        lhs = sum(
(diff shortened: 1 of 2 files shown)


jokochems:
@p-snft: I think, fixing the constraint tests in #943 should fix the failing tests in this PR as well. Maybe, the lp file needs to be re-created due to the changed order in the newer pyomo version.

p-snft:
Unfortunately, the order of the constraints is not reproducible in the latest version of Pyomo. They did this to increase performance of the write-out. What is reproducible, however, is meaningful rounding, e.g. 1/5 = 0.2 (in contrast to 0.1999999998 or something).

jokochems:
Hi again, @p-snft,

I re-arranged the lp file accordingly. Now it should be all working. Would you please review and merge in case it is okay?

Thank you and best regards

p-snft (approved):
I don't see any code issues. As the SinkDSM is experimental anyway, I will not step through the functionality (esp. if it really works as intended) with this review.

Issue #932, opened 2023-04-03
Bump pyomo version in requirements?


lensum:
**Describe the bug**
When installing oemof.solph via pip, the requirements for pyomo is given with: `"pyomo >= 6.0.0, < 7.0",`.
However, simple code such as the `basic_example.py` does not run with the `pyomo` versions between `6.0.0` and `6.4.2`(inclusive). The error returned points to `solph\buses\__init__.py`, line 471 with:

`TypeError: 'tuple' object does not support item assignment`

With `pyomo >= 6.4.3`, the error vanishes. 

I see two options to fix that:
1) Change the code in  `solph\buses\__init__.py`, or
2) Change the `pyomo`-dependency to `"pyomo >= 6.4.3, < 7.0"`

I assume 2) will be easier to implement.

**To Reproduce**
1.  Create a new venv
2.  Run `pip install oemof.solph`
3.  Check if `6.0.0 <= pyomo <= 6.4.2`, if not, run `pip install --upgrade --force-reinstall pyomo==6.0.1` (or any other version that fits the previous requirement)
4.  Run `basic_example.py` in the new venv
5.  See error message
6.  Run `pip install --upgrade --force-reinstall pyomo==6.4.3` (or higher)
7.  Marvel at the results of the script

**Expected behavior**
The examples should run with all versions specified in the install-requirements. (The expected behavior is ofc related to #922, but the reason for the error is not, I think).

**Desktop (please complete the following information):**
 -   OS: Windows10
 -   Python version: 3.11

**Additional context**
I should point out, that when **only** installing oemof.solph in a new environment, `pip` pulls the latest version of `pyomo` (6.5.0), so there won't be an error. But if other packages are installed beforehand, the pyomo version could be lower, so I think a fix would be great.

p-snft:
As far as I remember, the problem is caused by an incompatibility of old versions of pyomo with current numpy. Previously, before it was fixed by pyomo, we worked around this by setting a maximum version for numpy.

In fact, we could "depend on" `pyomo >= 6.4.3` just because the older versions are known to be broken.

p-snft:
Somehow, this outlived itself quite quickly. Pyomo 6.6 changed the structure of its lp files, so #932 suggest to require it to be able to test everything.

lensum:
Than I close this issue in favor of #943!

p-snft:
@Lensum: Would you mind leaving a review over there?

PR #943, opened 2023-05-28
Fix GitHub actions

p-snft:
There seems to be an issue with gh actions, as test succeed locally but fail at GitHub.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -9,7 +9,7 @@ envlist =
     py3-nocov,
     report
 
-[gh-actions]
+[gh]
 python =
     3.8: py38
     3.9: py39


p-snft:
Pyomo has a new LP writer (https://github.com/Pyomo/pyomo/pull/2823, https://github.com/Pyomo/pyomo/pull/2840). Hopefully, we could just move to (also new) `pyomo.repn.tests.lp_diff` which claims to offer “more robust comparisons of LP files”. (According to the issue, he new Pyomo LP writer also drops some lines that have no effect on the result.)

p-snft:
Reshuffling the lp is actually not needed because we do not test for the order of lines, but as rounding will lead to 0.19999999998 now being 0.2, the respective commit should not be reverted.

lensum:
Issue #932 will be closed with this PR. I can't quite grasp the changes associated with the switch to `lp_diff`, so I can't really review on that.

jokochems:
> Reshuffling the lp is actually not needed because we do not test for the order of lines, but as rounding will lead to 0.19999999998 now being 0.2, the respective commit should not be reverted.

Ah, thanks for that clarification. I was wondering why the order of lines within the lp files was changed which "blows up" the diffs for this PR.

PR #950, opened 2023-06-21
Set arguments defaults in link

### LINKED ISSUE #949: Set `None` as default values for explicit arguments in Link component
Due to the process of building `oemof.tabular.facades`, it's necessary that the Link component uses its default values during its instantiation, as no values are passed. These default values are updated afterward. 

Therefore, I will add `None` as default values for the explicit arguments and add a warning for this case.

nailend:
close #949

Due to the process of building oemof.tabular.facades, it's necessary that the Link component uses its default values during its instantiation, as no values are passed. These default values are updated afterward.

Therefore, I will add None as default values for the explicit arguments and add a warning for this case.

### DIFF
--- a/src/oemof/solph/components/_link.py
+++ b/src/oemof/solph/components/_link.py
@@ -80,10 +80,10 @@ class Link(on.Transformer):
 
     def __init__(
         self,
-        label,
-        inputs,
-        outputs,
-        conversion_factors,
+        label=None,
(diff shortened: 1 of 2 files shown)


p-snft (approved):
Allowing to instantiate without arguments is also consistent with the other components that we have. Sorry, that I did not see that issue before.

PR #954, opened 2023-07-06
Fix/example section in the readme is still refering to the oemof example repository

p-snft:
Example section in the README was still referring to the oemof-example repository. This PR fixes that. Secondly, the quality of the integration of the examples into the documentation is improved.


p-snft:
Looking at https://github.com/oemof/oemof-examples/issues/84, we should really have this fix done.

fwitte:
Maybe also discontinue the examples Repo? Or refer to the correct locations from there?

p-snft:
> Maybe also discontinue the examples Repo? Or refer to the correct locations from there?

I would also discontinue it. In the current form, it has no added value.

p-snft:
I (temporarily) activated this branch on RTD so that you can see how the [examples render in the solph documentation](https://oemof-solph.readthedocs.io/en/fix-example-section-in-the-readme-is-still-refering-to-the-oemof-example-repository/examples/storage_investment.html). I really like it and consider this PR done.

fwitte:
I can have a look on like Sunday if it is sufficient

p-snft:
> I can have a look on like Sunday if it is sufficient 

Sure. My plan is to create the release middle of next week.

p-snft:
I just realised that there are still links to the example repository, e.g. at https://oemof-solph.readthedocs.io/en/fix-example-section-in-the-readme-is-still-refering-to-the-oemof-example-repository/usage.html#using-the-excel-csv-reader.

examples/activity_costs/activity_costs.py
fwitte, line 1: For some reason it seems I cannot annotate single lines here, but why are the lists all indended? The very first list level should have no indent to my knowledge? I'd change them everywhere, but would like to make sure, that is desired before.

examples/flow_count_limit/flow_count_limit.py
```diff
@@ -6,10 +6,18 @@
 
 Something...
```
fwitte, line 7: This is missing a description.

(P.S.: Inline comments are working again somehow...)

p-snft, line 7:
```suggestion
Example that shows how to use "flow_count_limit".

This example shows a case where only one out of two Flows can be
active at a time. Another typical usage might be a connection to a
grid where energy can only flow into one direction or a storage that
cannot be charged and discharged at the same time.

Note that binary variables are computationally expensive. Thus, you
might want to avoid using this constraint if you do not really need it.
```

p-snft:
```suggestion
Example that shows how to use "flow_count_limit".

This example shows a case where only one out of two Flows can be
active at a time. Another typical usage might be a connection to a
grid where energy can only flow into one direction or a storage that
cannot be charged and discharged at the same time.

Note that binary variables are computationally expensive. Thus, you
might want to avoid using this constraint if you do not really need it.
```

fwitte (approved):

PR #957, opened 2023-07-19
Make varying period lenghts possible

### LINKED ISSUE #955: Determining commission period fails if multi-periods don't have the same length
**Describe the bug**

The constraint to determine the endogenous decommissioning fail in case the periods used in a multi-period approach have not the same length. This is rooted [**here**](https://github.com/oemof/oemof-solph/blob/2b995c9fdf397d5b5cae95bbd83f0e8ec443bc63/src/oemof/solph/components/_generic_storage.py#L1244) for the generic storage component, but probably applies to all other components as well. 

I have three investment periods in my example (2015, 2035, 2050) with the period lengths (0, 20, 15) and a defined component lifetime of 20 years and no initial installed capacity. Solving the example results in the following investment results (see screenshot). In 2015, there is an initial investment (**invest**) of 11.2 MW resulting in a **total** of 11.2 MW. In the next period (20 years later) an investment of 7 MW takes place but 11.2 MW of the existing endogenously installed capacity (**old_end**) is decommissioned, resulting in a **total** of 7 MW. Up to here, all is totally valid. In 2050, however, further decommissioning of 11.2 MW takes place (**old_end**), although there was only 7 MW capacity left  (**total** in 2035). To equal the `_total_storage_capacity_rule` an investment of 11.2 MW has to be done (**invest**).

image

This happens, as in period 2 (2050) the following constraint is generated:
`GenericInvestmentStorageBlock.old_end[el-storage,2]  ==  GenericInvestmentStorageBlock.invest[el-storage,0]`
The wrong commission period is selected. 
although in period 1 this constraint was already defined:
`GenericInvestmentStorageBlock.old_end[el-storage,1]  ==  GenericInvestmentStorageBlock.invest[el-storage,0]`

**To Reproduce**

I found this, implementing the multi-period approach to `oemof.tabular`. I recreated an example in `oemof.solph`:

1.  Checkout branch `reproduce_decomissioning_bug` 
5.  Open `examples/scripting/reproduce_decommissioning_bug.py`
6. Set a breakpoint at line 118, run the script and check the last 2 defined variables.

**Expected behavior**

The expected behavior would be a decommissioning of only 7 MW in 2050.

If you change the year of the last period from 2050 to 2055. Everything works as expected, as the period lengths are now the same.

 -   Ubuntu 20.04 focal
 -   Python 3.10

nailend:
### Changes

This feature fixes #955 and introduces the capability to utilize multi-periods with varying lengths. To achieve this, the `_old_capacity_rule_end` in the `investment_flow_block` and `_old_storage_capacity_rule_end` of `generic_storage` were reworked. It seems that the same approach needs to be applied to `SinkDSM` as well.

### Challenges

The methodology behind this implementation can be quite complex to grasp. I have attempted to provide as many comments as possible to aid understanding. The major challenges encountered were as follows:

1. Matching investment periods with decommissioning periods.
2. Incorporating periods without decommissioning.
3. Accounting for periods with decommissioning of multiple investment periods. (That's the reason for the complicated for loop)

### Description

The method is based on an upper-right matrix that describes the temporal distance between each of the periods. Let's consider an example with the following periods:
-  2000, 2020, 2035, 2045, 2050, 2060, 2075, 2095
- component lifetime of 20 years.

The period matrix appears as follows:

periods_matrix

In this matrix, the values represent the age of a component based on the period it was invested in (rows) and decommissioned (columns). The value at the top-right corner indicates that a component would be 95 years old if it was invested in period 0 and decommissioned in period 7.

To match investment periods with decommissioning periods, we need to select, for each row, the smallest value that is greater than or equal to 20 (the lifetime). We then obtain the respective indexes, where the row index corresponds to the investment period and the column index corresponds to the decommissioning period.

periods_matrix_2

Additionally, it is necessary to address periods without decommissioning (depicted in blue boxes) and periods where multiple investment periods are decommissioned (depicted in green boxes).

periods_matrix_3

### Remarks

It's probably possible to do this without [this complicated for-loop](https://github.com/oemof/oemof-solph/blob/d22e4722dfff72b72b52e88c9213a19f6ff01d77/src/oemof/solph/flows/_investment_flow_block.py#L525-L558) but I did't find a solution to group duplicated investment periods to chain and add them to the constraint.

### Help

I guess the LP-file test should be sufficient. Please double-check the logic for `c_e_GenericInvestmentStorageBlock_old_rule_end` and `c_e_InvestmentFlowBlock_old_rule_end`.

* Do you think there is anything else missing? 
* Is there anything to do in the documentation?

Thanks for your help

### TODO:

- [x] implement in `investment_flow_block`
- [x] implement in `generic_storage`
- [x] implement in `SinkDSM`
- [x] add `SinkDSM` to LP-file test
- [x] add test
- [x] remove same test
- [x] add LP-file test
- [x] reduce timesteps of LP-test


tests/test_scripts/test_solph/test_multi_period_model/test_multi_period_varying_period_length.py
```diff
@@ -0,0 +1,291 @@
+import pandas as pd
+from pandas.testing import assert_frame_equal
+
+from oemof import solph
+
+
+def test_multi_period_varying_period_length(lp=False):
```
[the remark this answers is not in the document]
nailend: Going to get rid of this test anyway

src/oemof/solph/_energy_system.py
```diff
@@ -192,6 +193,26 @@ def _extract_periods_years(self):
...
+        period_distance_matrix: np.array
+
+        """
+        periods_matrix = []
+        if self.periods is not None:
+            period_years = np.array(self.periods_years)
+            for k, v in enumerate(period_years):
+                row = period_years - v
+                row = np.where(row < 0, 0, row)
+                periods_matrix.append(row)
```
jokochems, line 213: Variable `k` is not used, but I see why you put it this way. It is elegant.

We might either change to `_` for assigning a value we don't need or use something like:
```python3
for i in range(period_years.shape[0]):
    row = period_years - period_years[i]
    row = np.where(row < 0, 0, row)
    periods_matrix.append(row)
```

src/oemof/solph/components/_generic_storage.py
```diff
@@ -1232,24 +1233,75 @@ def _old_storage_capacity_rule_end(block):
...
+                    expr = self.old_end[n, 0] == 0
+                    self.old_rule_end.add((n, 0), expr)
+
+                    # all periods not in decomm_periods have no decommissioning
                     for p in m.PERIODS:
-                        # No shutdown in first period
-                        if p == 0:
+                        if p not in decomm_periods and p != 0:
                             expr = self.old_end[n, p] == 0
                             self.old_rule_end.add((n, p), expr)
```
jokochems, line 1264: You could combine the 0th period into that statement, couldn't you?

I.e.

```python3
if p not in decomm_periods or p == 0:
    ...
```

src/oemof/solph/components/_generic_storage.py
```diff
@@ -1232,24 +1233,75 @@ def _old_storage_capacity_rule_end(block):
...
-                            expr = self.old_end[n, p] == 0
-                            self.old_rule_end.add((n, p), expr)
+                            expr = self.invest[n, invest_p]
+                            # overwrite decomm_p memory
+                            last_decomm_p = decomm_p
+
+                    if last_decomm_p != 0:
+                        # Add constraint of last iteration
+                        expr = self.old_end[n, last_decomm_p] == expr
+                        self.old_rule_end.add((n, last_decomm_p), expr)
```
jokochems, line 1304: Great solution, though complicated to follow along! :-)

jokochems (commented):
Thank you, @nailend! This looks sound to me. Congrats to coming up with a well-thought thorugh, though quite complicated fix for that. I guess the reason for having a rather complicated fix is that the issue itself brings that complication, right? ;-)

I suggested some minor code modifications, but would not alter the essence of it. Also, I think, we need to remove all that inline commenting and document it somewhere central. I guess, I'd rather prefer having a dedicated section on that in the sphinx documentation, compared to blowing up the docstrings for that constraint formulation, but we can argue about that.

jokochems:
> It seems that the same approach needs to be applied to `SinkDSM` as well.
You're correct about that. It is basically the same as for any investment flow. We might address these code "duplications" at some point in the future, but that's another issue.

Also you might look into the open black issues.

jokochems:
> Is there anything to do in the documentation?

Concerning that, my preference would be to have a dedicated subsection under the multi-period mode in the `usage.rst` file as well as a an extended docstring for that altered constraint rules. The current inline commenting should be reduced to a minimum.

src/oemof/solph/components/_generic_storage.py
nailend: No, this doesn't work as the `np.argmin(np.where..)` returns 0 if the condition is not matched. This would lead to having the 0th period constraint multiple times.

src/oemof/solph/_energy_system.py
nailend: You are right, that's probably leftovers from changed code.

nailend:
@jokochems Thanks a lot for your review. I took care of the modifications you suggested and a few more things. 

I tried to reduce the inline comments, but as you said yourself it's quite complicated, even though you are the person most knowledgeable about it. To make the code comprehensible and open for developer changes, I would prefer to keep them. I don't think that this level of detail is necessary for the user, so I added only a few words to the `usage.rst`.

Would you like to double-check the LP-file again, or do you think it's ready to merge? (To check the constraints ending with `_rule_end` is enough and luckily quite easy with the matrix from above)

Thanks again for your support.

docs/usage.rst
```diff
@@ -1030,7 +1030,10 @@ First, you start by defining your energy system as you might have done before, b
     my_energysystem = solph.EnergySystem(timeindex=my_index, periods=periods)
 
 If you want to use a multi-period model you have define periods of your energy system explicitly. This way,
-you are forced to critically think, e.g. about handling leap years, and take some design decisions.
+you are forced to critically think, e.g. about handling leap years, and take some design decisions. It is possible to
+define periods with different lengths, but remember that decommissioning of components is possible only at the
+beginning of each period. This means that if the life of a component is just a little longer, it will remain for the
+entire next period. This can have a particularly large impact the longer your periods are.
```
jokochems, line 1036: Nice clarification, thanks!

src/oemof/solph/components/_generic_storage.py
```diff
@@ -1229,27 +1241,76 @@ def _old_storage_capacity_rule_end(block):
...
-                            for k, v in enumerate(m.es.periods_years):
-                                if m.es.periods_years[p] - lifetime - v < 0:
-                                    # change of sign is detected
-                                    comm_p = k - 1
-                                    break
-                            expr = self.old_end[n, p] == self.invest[n, comm_p]
-                            self.old_rule_end.add((n, p), expr)
+
+                    # multiple invests can decommission in the same period
+                    # but only sequential ones, thus a memory is introduced and
```
jokochems, line 1277: Suggestion on wording: "bookkeeping" instead of "memory"?

jokochems (approved):
I did not work through each and every detail again. It's pretty much the same approach for each component that is eligible for investments, right?

I suggest to alter the wording to "bookkeeping" instead of "memory".

In terms of the inline comments, I'd be fine with it, but I'd like to hear @p-snft's take on that and would like to leave a final review and approval up to him.

src/oemof/solph/components/_generic_storage.py
```diff
@@ -1229,27 +1241,76 @@ def _old_storage_capacity_rule_end(block):
...
-                            comm_p = 0
-                            for k, v in enumerate(m.es.periods_years):
-                                if m.es.periods_years[p] - lifetime - v < 0:
-                                    # change of sign is detected
-                                    comm_p = k - 1
-                                    break
-                            expr = self.old_end[n, p] == self.invest[n, comm_p]
-                            self.old_rule_end.add((n, p), expr)
+
+                    # multiple invests can decommission in the same period
```
jokochems, line 1276: Another minor wording remark: I think, it should be "can be decommissioned" not "can decommision".

jokochems (commented):
I had another look into the decailts: To me it looks all good and I understand the logics as well as the analogies (you might say duplications) that are there.

The minor remarks on the wording hold for all code occurences, i.e. `_generic_storage.py`, `_sink_dsm.py` as well as `_investment_flow_block.py`.

One thing that was a bit counter-intuitive is that in the constraint test, the setup method gets somewhat "overwritten" as an energy system is set up twice (you'll notice when debugging and putting a break point into the new `_extract_periods_matrix` method). But I'd say, that's also okay, because this is an experts user area anyways and the test does what it is supposed to do and succeeds.

nailend:
Thanks a lot!
> One thing that was a bit counter-intuitive is that in the constraint test, the setup method gets somewhat "overwritten" as an energy system is set up twice.

Yes, this is done by purpose as the setup method only has three periods defined and there were more needed to check the various decommissioning cases.

p-snft (approved):
This is definitely an improvement. It might not be optimal, but as we still have the disclaimer for multi-period optimisation

> CAUTION! (…) Please be aware that the feature is experimental as of now.

I am very okay with that. For the future, we should talk about refactoring the tests, so that understanding the tests is easier. (I did not go through the lengthy LP files.) But that's a general issue with solph and not specific to the current PR.

PR #960, opened 2023-07-27
Implement periodic behaviour into  `plumbing.sequence` 

### LINKED ISSUE #959: Handling period-specific parameters in multi-period-investment
Dear all,

in our current project, we want to make use of period-specific parameters in a multi-period-investment. Currently, the only way to do this is passing an iterable to this parameter with values for every timestep. As we are using `omeof.tabular`, we could integrate a preprocessing to generate timeseries there, but we would find a `oemof.solph` solution more charming from user perspective. 

We would like to hear your opinion about this and collect possible challenges/restrictions. The idea is to adapt `_plumbing.sequence`, that if a tuple or maybe dict is passed, it returns period specific values. This would change the functionality of the function fundamentally, as the length of the list would need to be known. Further, the number of timesteps per periods need to be passed. This should be optional, so that the original way is preserved.

* Do you have any idea. how to handle this?
* What's the best way to pass periods information?
* Could there be any problems with detecting tuple or dict from csv? (This needs to be handled in `oemof.tabular` `read_from_datapackage`)

@jokochems, @p-snft, do you have any thoughts on this?

We are looking forward to your ideas and thank you for your help.

iuljan:
close #959 

A periodically changing `sequence` is handy for the multi-period investment, when the parameters should change per period.
For this I have adapted `plumbing.sequence` and `plumbing._sequence`. An additional distinction must be made for this behavior. 

Apparently, the approach with the distinction by variability of the passed iterables in `sequence` does not work, because pandas.series are interpreted as immutable sequence.

I therefore decided to specify periodic behavior only when a dictionary is passed. The explicit keys `len` and `values` also make it more clear

    >>> x = sequence({"len":9,"values":[1,2,3]})
    >>> print(x)
    [0, 0, 0, 1, 1, 1, 2, 2, 2]
    >>> x[1]
    0
    >>> x[4]
    1
    >>> x[7]
    2

I would appreciate your opinion on this @p-snft or @uvchik if this could break anything

### DIFF
--- a/src/oemof/solph/_plumbing.py
+++ b/src/oemof/solph/_plumbing.py
@@ -177,7 +177,10 @@ class _Sequence(UserList):
         return str([i for i in self])
 
     def __len__(self):
-        return max(len(self.data), self.highest_index + 1)
+        if self.periodic_values:
+            return self.highest_index
+        else:
+            return max(len(self.data), self.highest_index + 1)
 
     def __iter__(self):
         if self.periodic_values:


src/oemof/solph/_plumbing.py
```diff
@@ -39,22 +44,55 @@ def sequence(iterable_or_scalar):
...
+    the values passed in period_values and the lengths of the periods are equal
+    and sum up to the value of highest_index.
 
     Parameters
     ----------
     source:
-    default:
+    default: int or float or string
+        Default value for the sequence
+    period_values: iterable
```
FelixMau, line 92: I personally would prefer `periodical_values`

src/oemof/solph/_plumbing.py
```diff
@@ -73,21 +111,61 @@ class _Sequence(UserList):
...
+            return self.default
 
     def __init_list(self):
-        self.data = [self.default] * (self.highest_index + 1)
+        if self.period_values:
+            self.data = [
+                value
+                for value in self.period_values
+                for _ in range(self.highest_index)
+            ]
```
FelixMau, line 166: I am unsure if I understood your example from above correct. As far as I see here if 
`s = _Sequence(period_values=[0,1,2], highest_index=9)` is passed we would have a list that is 3*9 = 27 long?

nailend: Thanks for the hint, I missed correcting my first implementation. Now it's 9 time index in total, 3 per period.

And yes, you are right, the length of the years have to be the same. No mix of normal and leap year is possible.

src/oemof/solph/_plumbing.py
nailend: After some research, I will go for `periodic_values`

nailend:
We decided to solve this in `oemof.tabular` as this approach comes with more difficulties then expected...

PR #962, opened 2023-07-31
Add storage_costs to GenericStorage

p-snft:
Implement parameter `storage_costs` for `GenericStorage`: `storage_costs` might exist, e.g. if space can be rented depending on the stored energy. So, this might be seen as an alternative to investment optimisation. A second use case is to incentivise using the storage in the last time step if the storage is not periodic. This can be done by giving `storage_costs=n_time_steps * [0] + [-energy_price]`, so that stored energy in the last time step is valued with the energy price.

### DIFF
--- a/src/oemof/solph/components/_generic_storage.py
+++ b/src/oemof/solph/components/_generic_storage.py
@@ -108,7 +108,7 @@ class GenericStorage(network.Component):
         nominal_storage_capacity should not be set (or set to None) if an
         investment object is used.
     storage_costs : numeric (iterable or scalar), :math:`c_{storage}(t)`
-        Cost (per energy) for using the storage.
+        Cost (per energy) for having energy in the storage.
     lifetime_inflow : int, :math:`n_{in}`
         Determine the lifetime of an inflow; only applicable for multi-period
         models which can invest in storage capacity and have an
@@ -419,8 +419,8 @@ class GenericStorageBlock(ScalarBlock):
                                 :math:`\delta(t)` and
                                 timeincrement
                                 :math:`\tau(t)`
-    :math:`c_{storage}(t)`      costs for storaging     `storage_costs`
-                                energy
+    :math:`c_{storage}(t)`      costs of having         `storage_costs`
+                                energy stored
     =========================== ======================= =========
 
     **The following parts of the objective function are created:**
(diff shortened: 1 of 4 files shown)


src/oemof/solph/components/_generic_storage.py
```diff
@@ -596,7 +606,18 @@ def _objective_expression(self):
...
         self.fixed_costs = Expression(expr=fixed_costs)
-        return self.fixed_costs
+
+        storage_costs = 0
+
+        for n in self.STORAGES:
+            if n.storage_costs[0] is not None:
+                storage_costs += self.storage_content[n, 0] * n.storage_costs[0]
+                for t in m.TIMESTEPS:
+                    storage_costs += self.storage_content[n, t+1] * n.storage_costs[t+1]
```
nailend: The earlier described behavior probably results somehow from having the cumulated sum of costs over all timsteps. Maybe a note should be added that it might be complicated to directly steer the behavior?!

nailend:
Could you please briefly explain, what do you want to show in your example? Especially with your choice of setting the `storage_costs` to -1/4 of the marginal costs?

p-snft:
> Could you please briefly explain, what you want to show in your example? Especially with your choice of setting the `storage_costs` to -1/4 of the marginal costs? It's not quite clear to me.

I decided to go to a more practical example. The original one arose from the script I wrote to check the behaviour of the new functionality.

tests/lp_files/storage.lp
```diff
@@ -2,6 +2,10 @@
 
 min 
 objective:
++4000.0 ONE_VAR_CONSTANT
++0.1 GenericStorageBlock_storage_content(storage_no_invest_1)
++0.1 GenericStorageBlock_storage_content(storage_no_invest_2)
++0.1 GenericStorageBlock_storage_content(storage_no_invest_3)
```
nailend, line 8: Apparently there is no `period` index for the storage block. This is not a problem of this issue, but should be fixed for the sake of consistency, shouldn't it?

nailend (approved):
Thanks for the updated example, I think it adds to a better understanding.

tests/lp_files/storage.lp
p-snft: As the storage content is given in absolute values, only the limit (which might change from period to period) needs to have the index. I will leave it as it is.

PR #965, opened 2023-08-04
Define multi-period investment storage level for very first time step of optimization

jokochems:
Add a constraint to force initial storage level to 0 in case of multi-period investment modelling to prevent storage from being filled without any associated costs in case of investments in the first period.

### DIFF
--- a/src/oemof/solph/components/_generic_storage.py
+++ b/src/oemof/solph/components/_generic_storage.py
@@ -1032,21 +1032,20 @@ class GenericInvestmentStorageBlock(ScalarBlock):
         # ########################## CHECKS ###################################
         if m.es.periods is not None:
             for n in group:
-                error = (
+                error_fixed_absolute_losses = (
                     "For a multi-period investment model, fixed absolute"
                     " losses are not supported. Please remove parameter."
                 )
                 if n.fixed_losses_absolute.default != 0:
-                    raise ValueError(error)
-                warning = (
+                    raise ValueError(error_fixed_absolute_losses)
+                error_initial_storage_level = (
                     "For a multi-period model, initial_storage_level is"
-                    " not supported.\nIt is suggested to remove that"
-                    " parameter since it has no effect.\nstorage_content"
-                    " will be zero, until there is some usable storage "
-                    " capacity installed."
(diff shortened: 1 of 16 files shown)


src/oemof/solph/components/_generic_storage.py
```diff
@@ -1312,6 +1312,22 @@ def _old_storage_capacity_rule(block):
 
             self.old_rule_build = BuildAction(rule=_old_storage_capacity_rule)
 
+            def _initially_empty_rule(block):
+                """Ensure storage to be empty initially"""
+                for n in self.INVESTSTORAGES:
+                    for t in m.TIMESTEPS:
+                        if t == 0:
+                            expr = self.storage_content[n, 0] == 0
+                            self.initially_empty.add((n, t), expr)
```
nailend, line 1321: Is the timestep loop necessary? It should be enough to set this constraint just for `t=0`, since there will always be a time step 0, right?
```
                    expr = self.storage_content[n, 0] == 0
                    self.initially_empty.add((n, 0), expr)
```
Tried it and tests still work

nailend (approved):
The change in rule formulation that I am proposing is not mandatory.

I didn't check all the LP file changes in detail, as most of it are just changes in the line order besides the new rule. 
Thanks for fixing this right away!

nailend:
With this, the [SuspiciousUsageWarning](https://github.com/oemof/oemof-solph/blob/a213d2ea17c5dab6932df4dd5d1bf15f87ea67dc/src/oemof/solph/components/_generic_storage.py#L1049) can be changed to an error?

src/oemof/solph/components/_generic_storage.py
jokochems: Agree. That's better.

jokochems:
> With this, the [SuspiciousUsageWarning](https://github.com/oemof/oemof-solph/blob/a213d2ea17c5dab6932df4dd5d1bf15f87ea67dc/src/oemof/solph/components/_generic_storage.py#L1049) can be changed to an error?

Yes, that's a good suggestion.

p-snft:
Makes sense for me. I will merge this.

PR #967, opened 2023-08-09
Fix/add custom attributes link

MaGering:
With this PR the Link component gets the argument `custom_attributes`.

--------------------------------------------------------------------------------------
Because I can't create an issue - here's the description:
The link component is missing the argument `custom_attributes` in its [class](https://github.com/oemof/oemof-solph/blob/dev/src/oemof/solph/components/_link.py).

I've discovered this using the link component in [oemof-tabular](https://github.com/oemof/oemof-tabular). During the postprocessing I've been losing the `type = 'link'`. Instead it was set to None. This is due to [these lines](https://github.com/oemof/oemof-tabular/blob/8d26a4891461006901fa6178d8ecb44d73c6d8f8/src/oemof/tabular/_facade.py#L85-L91) where `kwargs_unexpected` are lost if the component does not have the argument "custom_attributes".

**Local system and env info**
 -   Linux Mint 20.2
 -   Python version 3.10

### DIFF
--- a/docs/whatsnew/v0-5-1.rst
+++ b/docs/whatsnew/v0-5-1.rst
@@ -43,6 +43,7 @@ Bug fixes
 
 * Fixed error when calling `oemof_installation_test` as console script.
 * Corrected several typos in the docs.
+* Add missing 'custom_attributes' for the link component
 
 Testing
 #######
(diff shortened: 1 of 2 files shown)


p-snft:
Thanks a lot for fixing this. You might want to add you name to AUTHORS.rst and CITATION.cff.

MaGering:
What do we do about `coverage/coveralls` failing @p-snft?

p-snft (approved):
This implements a pass-through for functionality that is proposed in oemof.network. As all other classes already implement it the same way, it's okay to have it here for consistency. (As the functionality is not implemented, I also do not see a way how to properly test this, so the decreasing coverage should be okay this time.)

PR #970, opened 2023-08-21
Replace setup.py invocation by build

fwitte:
Direct invocation of `setup.py` seems to be long deprecated (e.g. https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html), it can be replaced with the `build` package.

### DIFF
--- a/.github/workflows/packaging.yml
+++ b/.github/workflows/packaging.yml
@@ -26,9 +26,9 @@ jobs:
         python-version: ${{ matrix.python-version }}
     - name: Install dependencies
       run: |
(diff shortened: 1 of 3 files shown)


p-snft (approved):
Thanks for coming up with this. There should really be a deprecation warning. Maybe unrelated, there is (and was) a warning when building the documentation (`tox -e docs`):

> package config for docs, py310, py39 is editable, however the build backend setuptools.build_meta does not support PEP-660, falling back to editable-legacy - change your configuration to it

I hoped it would be solved by this, but this seems to be another issue.

PS: I am also updating https://github.com/oemof/oemof/wiki/How-to-release.

PR #975, opened 2023-09-11
Fix/examples in docs

### LINKED ISSUE #974: Error in first code example in user's guide
There seems to be a bug in the first code example in the [user's guide](https://oemof-solph.readthedocs.io/en/stable/usage.html#set-up-an-energy-system):

```
from oemof.solph import create_year_index
my_index = create_year_index(2011)
```

After 10 minutes of searching I found out that "create_year_index" seems to have been replaced by "create_time_index". This should be fixed as it can be enough to scare off potential new users.

p-snft:
There are small sources of confusion in the documentation that can be showstoppers for new users.

Closes #974


jokochems (approved):
Looks good, thanks @p-snft. I can approve.

PR #980, opened 2023-09-28
Integrate TSAM mode in oemof.solph

### LINKED ISSUE #973: Adding time series aggregation with the tsam repository
I am adding a pipeline to the tsam repository for time series aggregation. Tsam includes several aggregation algorithms. The quality of different algorithms are explained in the paper linked. 

Tsam will be integrated into a function into _helpers, called aggregate_time_series(). (**The rigth place to integrate tsam is a point to discuss**) 

The idea is:
1. Aggregate the input time series (aggregate_time_series())
2. Generating a energy system with the aggregated timeseries index 
3. Add the aggregated time series to the components. 

In the first draft all time steps of a series have to have the same resolution. This is based on the workflow of tsam. So it's not possible to have non euidistant time steps like in: [non_equidistant_time_step_example.py](https://github.com/oemof/oemof-solph/blob/dev/examples/time_index_example/non_equidistant_time_step_example.py)

Two main advantages of the aggregation:
1. Optimization with aggregated time series need less time and have low loss in accuracy
2. Optimization of storages, espically long-term/seasonal, with less timesteps because of aggregation are still accurate. For this the gaps in between two periods are setted and the losses are considered by the timeincrement.

tsam:
- https://github.com/FZJ-IEK3-VSA/tsam
- https://www.sciencedirect.com/science/article/abs/pii/S0960148117309783?via%3Dihub
- https://www.sciencedirect.com/science/article/abs/pii/S0960148117309783?via%3Dihub

henhuy:
Closes #973 

This PR wants to integrate timeseries aggregation into `oemof.solph`. Thereby, newly integrated multi-period mode for `oemof.solph` can be used as well. 
The idea is, to run TSAM with all needed timeseries before setting up the energysystem and use outcome of aggregation to set up the energysystem. The energysystem components are  build with regarding aggregated timeseries. Additionally, a `tsa_parameters` parameter is given to the energysystem, which is used to set up component equations and constraints to ensure TSAM-related component behaviour.
Following equations/constraints have to be adapted/are replaced in order to work correctly in TSAM mode:
* `Model` component needs six additional indexes in order to set up storage equations (see below)
* storage content equations for `GenericStorageBlock` and `GenericInvestmentStorageBlock` have to be adapted:
  * default storage content is replaced by two storage indexes, inter and intra storage index
  * minimum and maximum rule of default storage content are replaced by applying max and min rule to inter storage
  * default storage balance rule is replaced by two balance rules for intra and inter storage contents
  * default balanced storage rule is replaced by balanced storage rule for inter storage contents
* Variable cost have to be weighted related to occurrences of typical cluster periods
* Full time hours have to be weighted
* Others?

How does the API look?
In [storage investment example v5](https://github.com/oemof/oemof-solph/blob/feature/integrate_tsam/examples/storage_investment/v5_invest_optimize_all_technologies_using_mp_and_tsam.py#L118-L180) I set up an ES with aggregated timeseries and related TSA parameters for the energysystem.

The newly introduced parameter `tsa_parameters` should be set up as follows:
* a separate timeseries aggregation has to be performed for each period
* for each period (in a multi-period approach), a dictionary with outputs from TSAM aggregation has to be given to ES
* see example `tsa_parameters` and needed key-value pairs below:
```
tsa_parameters=[
    # First period
    {
        "timesteps_per_period": aggregation1.hoursPerPeriod,
        "order": aggregation1.clusterOrder,
        "occurrences": aggregation1.clusterPeriodNoOccur,
        "timeindex": aggregation1.timeIndex
    },
    # Second period
    {
        "timesteps_per_period": aggregation2.hoursPerPeriod,
        "order": aggregation2.clusterOrder,
        "occurrences": aggregation2.clusterPeriodNoOccur,
        "timeindex": aggregation2.timeIndex
    }
]
```

TODOS:
- [x] Implement storage indexes (inter and intra) and equations/constraints for `GenericStorageBlock`
- [x] Implement storage indexes (inter and intra) and equations/constraints for `GenericInvestmentStorageBlock`
- [ ] Implement initial SOC equation
- [x] Implement weighted cost
- [x] Implement weighted full time hours
- [ ] Add documentation for TSAM mode
- [ ] Add docstrings including Latex equations
- [x] Check results with examples and tests


henhuy:
You can already check storage equations if you want to help!
Check out my storage example (it's called storage investment - but in order to make testing simpler, I added a non-invest storage) - in order to check storage contents (intra&inter) it's useful to set a breakpoint at SOC calculation in postprocessing:
https://github.com/oemof/oemof-solph/blob/7684cde5b548a5cbefebe395b3d6e158c86bc17f/src/oemof/solph/processing.py#L487

Other things which are very welcome, are:
- docstrings
- storage equations written in LaTex
- README section regarding TSAM mode and idea behind it

Maxhi77:
I finished the implementation of weighted cost before I was on vacation. I will add that soon to the PR.

henhuy:
> I finished the implementation of weighted cost before I was on vacation. I will add that soon to the PR.
> 

Nice! Could you create an extra branch for this and do a PR to my branch or show me your branch so that I can review?
I think it's easier to look at one change at a time..

Maxhi77:
I would like to add an further todo. TSAM targets on the use of longterm storage. Right now, if tsam parameters are added to the energysystem each storage will have an inter and intra SOC. For short-term storages which are used on an daily base, f.e.  batteries, that is not necessary and it would slow down the optimization. Therfore I think a state_variable should be added to the generic_storage. On basis of this variable it should be decided to add an inter and intra SOC or not. 

Then it would be possible to optimize a long term storage and a short term storage with best possible performance.

Maxhi77:
Linked to issue: https://github.com/oemof/oemof-solph/issues/987 it would be handy to rename `objective_weighting` and add a tsam weighting. Right now, `objective_weighting` gets calculated during pre-processing and includes the period occurence of typical periods (https://github.com/oemof/oemof-solph/pull/1000/commits/f571759b7416ddc070e41f291f326e37cd7e59f7). 

The idea would be:

- `tsam_weighting` gets generated based on `tsa_paramets` in `energy_system.py`. It inlcudes the occurence of typical periods.
- `time_increment` gets generated based on the input time_series. It includes the length of a timestep, compared to others.
- `objective_weighting` can be setted by the user to consider individual problems. 

Variable cost would be:
`variable_costs += (
                            m.flow[i, o, p, t]
                            * m.tsam_weighting[t]
                            * m.time_increment[t]
                            * m.objective_weighting[t]
                            * m.flows[i, o].variable_costs[t]
                            * ((1 + m.discount_rate) ** -m.es.periods_years[p])
                        ) `

In total the idea is more intuitive for the user and easier to use than calculating the objective_weighting in a pre_processing, which includes time increment, period occurence and individual preferences. 

What are your opinions?

henhuy:
Things that might need changes/fixes:
- `GenericInvestStorage` always starts with initial SOC of zero
- `GenericStorage` initial SOC is free, but must equal last SOC
- TSAM mode is only available in multi-period approach (but it is allowed to only have one period)

tests/test_scripts/test_solph/test_tsam/test_storage_invest_tsam_integration.py
```diff
@@ -0,0 +1,216 @@
...
+logger.define_logging()
+logging.info("Initialize the energy system")
+
+tindex_original = pd.date_range("2022-01-01", periods=8, freq="H")
+tindex = pd.date_range("2022-01-01", periods=4, freq="H")
+
+energysystem = solph.EnergySystem(
+    timeindex=tindex,
+    timeincrement=[1] * len(tindex),
+    periods=[tindex],
```
nailend, line 58: this is not needed and should be None, as you dont use multi-period, do you?

tests/test_scripts/test_solph/test_tsam/test_storage_tsam_integration.py
```diff
@@ -0,0 +1,209 @@
...
+logger.define_logging()
+logging.info("Initialize the energy system")
+
+tindex_original = pd.date_range("2022-01-01", periods=8, freq="H")
+tindex = pd.date_range("2022-01-01", periods=4, freq="H")
+
+energysystem = solph.EnergySystem(
+    timeindex=tindex,
+    timeincrement=[1] * len(tindex),
+    periods=[tindex],
```
nailend, line 57: same here

tests/test_scripts/test_solph/test_tsam/test_storage_invest_tsam_integration.py
henhuy: No, it's the opposite indeed: Currently, TSAM-mode is only available in multi-period mode! But for those examples, I only use a single period to make things easier to test...

henhuy: I still wonder if multi-period should/could be activated ALWAYS, and an "old" default oemof.solph simulation simply uses only a single period... but this is another issue.

nailend: its basically translated into pyomo like this but a lot of extra constrains and variables are not necessary (e.g. for decommissioning etc)

henhuy:
@MaxHiDLR could you explain your latest commits?
As I understand it, `storage_level` constraint has not been set before?
Ans what are those multiplexer values in processing file? 
THX 
(As code in this PR has grown, I would suggest from now on we both start PRs on this branch whenever we add/fix something - this makes it much easier to review new PR and merge it into this branch, instead of merging directly and reviewing afterwards. I will start doing so as well!)

Maxhi77:
@henhuy 
The storage_level.py, which is used in example/storage_level_constraint, wasn't working with the additional time_grid of tsam. Therefore I added it. As well, I added to the processing function that the multiplexer added by storage_level.py can be processed.

We definitely can start to do PR to simplify the understanding of the changing history. 

In the close future I will change the naming of the tsam "soc" to "storage content", to be more consistent with oemof wording.

henhuy:
> The storage_level.py, which is used in example/storage_level_constraint, wasn't working with the additional time_grid of tsam. Therefore I added it. As well, I added to the processing function that the multiplexer added by storage_level.py can be processed.

Good that you found this - I did not see an error? Or was it silently not working?

Maxhi77:
Silently not working, since the storage_level doesn't get used in the usual operation of the storage. Therefore it might be good to add a test for this scenario as well.

p-snft:
I'd like to talk about time indexing in a more general way. Of course, we should harmonize multi-period optimization and time series aggregation, but there is more. For example, there are different markets working of different time grids. So, if you offer you capacity, you have to do it for a full time slot. A generic multi-level time grid would also simplify modeling these kinds of things.

p-snft:
Result of the last dev meeting (November 2024 in Oldenburg was): Instead of having multiple code paths ("regular mode" and "TSAM mode"), non-aggregated time could just be treated as one representative period containing every time step. This would also follow the spirit of https://github.com/oemof/oemof-solph/issues/1107, which aims to combine upfront invest and multi-period invest.

henhuy:
> Result of the last dev meeting (November 2024 in Oldenburg was): Instead of having multiple code paths ("regular mode" and "TSAM mode"), non-aggregated time could just be treated as one representative period containing every time step. This would also follow the spirit of #1107, which aims to combine upfront invest and multi-period invest.

Sounds good and reasonable! Currently, I'm juggling too much projects which is why I haven't the time to make the according changes - but there might be coming a project where we want to use TSAM-oemof. This might be a good opportunity to finish this PR.

p-snft:
Including changes in https://github.com/oemof/oemof-solph/pull/1158, this would get too big. In particular, we will have version conflicts with the new documentation, etc. Thus, I allow myself to merge this one and work on the finalisation in smaller, iterative steps.

p-snft:
Black complains about untouched stuff. Will leave it as it is.

PR #981, opened 2023-10-02
Fix outdated multi-period usage docs

jokochems:
* Correct the type of `periods` attribute of an energy system to list.
* Add hint to usage of `timeincrement` in combination with `timeindex`.
* Provide minor doc fixes alongside.

### DIFF
--- a/src/oemof/solph/_energy_system.py
+++ b/src/oemof/solph/_energy_system.py
@@ -178,11 +178,9 @@ class EnergySystem(es.EnergySystem):
     def _extract_periods_years(self):
         """Map years in optimization to respective period based on time indices
 
-        Returns
-        -------
-        periods_years: dict
-            year of the start of each period,
-            relative to the start of the optimization run and starting with 0
+        Attribute `periods_years` of type list is set. It contains
+        the year of the start of each period, relative to the
+        start of the optimization run and starting with 0.
         """
         periods_years = [0]
         if self.periods is not None:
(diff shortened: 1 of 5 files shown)


jokochems:
@henhuy This should address the outdated docs. Thanks for pointing this out. Would you please review?

I really thought, I changed that in the first place after @p-snft pointed me to the possibility for code simplification. Seems like I either forgot or falsely commited to a wrong local branch that was never pushed back. Nevermind, now it should be correct.

src/oemof/solph/_energy_system.py
```diff
@@ -176,12 +176,12 @@ def __init__(
...
     def _extract_periods_years(self):
-        """Map simulation years to the respective period based on time indices
+        """Map years in optimization to respective period based on time indices
 
         Returns
         -------
         periods_years: dict
-            the simulation year of the start of each a period,
+            year of the start of each period,
             relative to the start of the optimization run and starting with 0
```
henhuy, line 185: Two things:
- function does not return anything, but sets member variable instead - so `Returns` section is not needed, right?
- if you want to use `Returns` section anyway or want to mention `period_years`, `periods_years` should be of type `list`

henhuy (changes requested):
Minor issue found.
Thanks for adapting docs!

src/oemof/solph/_energy_system.py
jokochems: Thank you for pointing this out. I resolved it. Please have a look.

henhuy (approved):
Thanks for fixing the docs!

jokochems:
@p-snft since this only affects docs, all tests are succeeding and @henhuy has already reviewed, I'll merge this.

Issue #556, opened 2019-01-23
Fix Cost Implementation


oakca:
Usually energy-system-models have following costs for energy generators (for oemof case it is `transformer-objects`); investment costs, maintenance costs, and variable costs of the units.

With the `investment-mode` of oemof, investment costs are modeled. Variable costs can be defined also inside `transformer-object`. However; since for the maintenance costs there is none attribute in `transformer-objects`, you have to find an other way to implement them. One approach is shown in following code-piece (adding the maintenance costs to `ep_costs`):

```
solph.Transformer(label='coal',
    inputs={bus['coal']: solph.Flow(
        investment=solph.Investment(
            ep_costs=invest + maintenance,
            maximum=100000,
            existing=0),
        variable_costs=0.1)},
    outputs={bus['Elec']: solph.Flow()},
    conversion_factors={bus['Elec']: 0.75})
```

However in some cases this trick won't work. Assume you got a transformer, which has a maximum investment capacity of 100000 and existing capacity of 50000. If the demand can be satisfied by 50k power, then the model will not invest more of the transformer. So the `ep_costs` even after adding the maintenance costs will not matter, and you will not be able to calculate the maintenance costs of the power plant. 

This is also important for the optimization. (e.g.: If the maintenance cost of a power plant is too expensive, maybe model will not invest in that power plant)

For the sake of the mentioned reasons, it would be best if the transformer objects have an attribute called `fixed_costs`, which eventually calculates amount of maintenance cost with multiplying the 'existing_capacity' and the `maintance_cost' of the power plant.

uvchik:
I cannot see how adding `fixed_cost` will help to solve your problem, because `existing_capacity * fixed_costs` will always be the same and therefore won't affect the solution.

What you need is a `shutdown` attribute with cost you will save if the power plant is totally shut down (never used).

`shutdown_total = shutdown_costs - maintenance`

if `shutdown_total`is negative you could have the case that an existing power plant is not used and a new one will be build instead due to a difference in maintenance costs.

At the moment we do not have something like this in oemof but you could implement that similar to the Investment object as an negative investment.

jnnr:
In earlier versions of oemof, there has been a fixed_costs attribute. It was discussed and decided to be removed here [PR396](https://github.com/oemof/oemof/pull/396) and here [PR407](https://github.com/oemof/oemof/pull/407). 

[...]

PR #982, opened 2023-10-02
Account for remaining values in multi-period models

jokochems:
* Subtract remaining values for investments expressing their value at the end of the optimization horizon.
* Limit fixed costs to actual optimization horizon.


jokochems:
I'll update the tests shortly. Feel free to already have a glimspe into the implementation. In case I haven't been mistaken, it hasn't been too hard to implement.

src/oemof/solph/flows/_investment_flow_block.py
```diff
@@ -887,6 +887,24 @@ def _objective_expression(self):
...
+                            capex=m.flows[i, o].investment.ep_costs[-1],
+                            n=lifetime,
+                            wacc=interest,
+                        )
+                        remaining_value = (
+                            self.invest[i, o, p]
+                            * remaining_annuity
+                            * remaining_lifetime
+                        ) * (
+                            (1 + m.discount_rate) ** (-m.es.periods_years[-1])
```
nailend, line 905: Why don't you use this formula, as you have just calculated the annuity? image [formula table](https://mathresearch.utsa.edu/wiki/index.php?title=Annuities#Formula_table)

src/oemof/solph/flows/_investment_flow_block.py
```diff
@@ -887,6 +887,24 @@ def _objective_expression(self):
                         * lifetime
                         * ((1 + m.discount_rate) ** (-m.es.periods_years[p]))
                     )
+                    remaining_value = 0
+                    if lifetime > m.es.periods_matrix[p, -1]:
+                        remaining_lifetime = (
+                            lifetime - m.es.periods_matrix[p, -1]
+                        )
+                        remaining_annuity = economics.annuity(
+                            capex=m.flows[i, o].investment.ep_costs[-1],
```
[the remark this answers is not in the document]
jokochems: Yeah, as if you were to reinvest (what Patrick suggested).

src/oemof/solph/flows/_investment_flow_block.py
jokochems: > Why don't you use this formula, as you have just calculated the annuity? image [source](https://mathresearch.utsa.edu/wiki/index.php?title=Annuities#Formula_table)

You are totally right about that and I'm a tiny bit ashamed, I didn't implement it like this in the first place. Actually, this is what you need to transfer annuities, i.e. identical periodical payments, to NPVs. I'll also integrate this.

jokochems:
After checking back with @nailend, I'll revise the implementation and guess that it gets leaner through the changes we discussed bilaterally.

src/oemof/solph/flows/_investment_flow_block.py
jokochems: The term 
$$\frac{(1 + i)^n -1}{i \cdot(1+i)^n}$$
is the reciprocal of the annuity factor. In German, it is called "Rentenbarwertfaktor". I think, there is no proper English equivalent for the term. I'm deriving it using `oemof.tools.economics.annuity()`

jokochems:
@nailend Please have a look when time allows:
* As we discussed, I limited the (per year) annuities and (per year) fixed costs to the optimization horizon.
* This way, there is no (urgent) need to account for remaining values.

@p-snft 
I see two ways to account for remaining values:
  1. Assume that investment expenses are "locked in".
  2. Account for changes in investment expenses over time - what you suggested.

In terms of implementation, I also see two ways:
- A. Limit to payments that happen within the optimization horizon.
- B. Consider all payments over the lifetime and subtract the outside of optimization horizon payments again.

Here is how I proceeded:
* Starting off, I wanted to go for approach 2.B. and tried to implement that.
* In the course of the discussion with @nailend, we found that approach A. would lead to cleaner code and processing (e.g. not having to account for outside horizon fixed costs as it was before). 
* Thus, I went for approach 1.A. in the first place. The main reason is that it was easier to implement, compared to 2.A. Alongside, I also fixed some bugs of the previous implementation.

In principle, we could also go for approach 1.A., but there are some design questions:
* We would need to compare at the end of the optimization horizon and add/subtract the difference in value. I'd have a rough idea how to do that.
* Additionally, we would require the information of the investment expenses at the end of the last period, i.e. the beginning of the next upcoming one that we do not simulate. While in principle, this information could be passed, it might also be somewhat counter-intuitive and potentially error-prone that you have to provide p + 1 instead of p periods.

What do you think, @p-snft?

jokochems:
@p-snft Adding to what I stated above, I gave it some thought: 
* What I have included in this pull request and what was the main motivation is actually rather a bug fix to prevent the bias towards earlier investments in the optimization horizon.
* Considering the cost change over time when evaluating an actual remaining value, to me is rather a feature. Thus, I suggest to leave it out here and open a dedicated pull request for that. Does this make sense to you?

src/oemof/solph/components/_generic_storage.py
```diff
@@ -1729,12 +1755,15 @@ def _objective_expression(self):
...
-                    investment_costs_increment = (
-                        self.invest[n, p]
-                        * annuity
-                        * lifetime
-                        * ((1 + m.discount_rate) ** (-m.es.periods_years[p]))
+                    duration = min(
+                        end_of_optimization - m.es.periods_years[p], lifetime
+                    )
+                    present_value_factor = 1 / economics.annuity(
+                        capex=1, n=duration, wacc=m.discount_rate
```
nailend, line 1762: Shouldn't this `wacc` also be `interest` like in L1756

docs/whatsnew/v0-5-2.rst
```diff
@@ -21,3 +26,8 @@ Other changes
 
 Contributors
 ############
+
+* Patrik Schönfeldt
+* Johannes Kochems
+* Julian Endres
+* Hendrik Huysgens
```
nailend, line 33: its spelled: Huyskens

nailend (commented):
Thanks for fixing this so fast and extending the docstrings. This looks good, I am just not sure about the different `wacc` assignments. Please check. 

I also took a look at the changed factors for the objective in the LP-files. The changes are quite strong. Have you tried a model run yet with the new code? This should definitely have some strong influences in the investments. Would be interested in your feedback on this.

nailend:
@jokochems does the range for the fixed costs of the `_simple_flow_block` also need to be adjusted [here](https://github.com/oemof/oemof-solph/blob/6fb392bc15581011c8b247017f27125b1ef2c892/src/oemof/solph/flows/_simple_flow_block.py#L497)?

docs/whatsnew/v0-5-2.rst
jokochems: Ty, and sorry for misspelling @henhuy

src/oemof/solph/components/_generic_storage.py
jokochems: Yeah, thank you. I might have overthought this:
* The annuity is calculated from an investors' point of view. You may assume the investor to be a "social planner" (whatever that shall be).
* Indeed, to be consistent, for converting it into a net present value, you want to stick to the investors' perspective (which I didn't). Mathematically, you could apply any interest rate here. I thought about only accounting for the overall model discount rate (I only use an assumed inflation rate here), but then I mix up the perspectives as you pointed out correctly.
* The (overall) discounting to t = 0 is already taken care about by the discount factor multiplier.

I'll fix this.

jokochems:
> @jokochems does the range for the fixed costs of the `_simple_flow_block` also need to be adjusted [here](https://github.com/oemof/oemof-solph/blob/6fb392bc15581011c8b247017f27125b1ef2c892/src/oemof/solph/flows/_simple_flow_block.py#L497)? And shouldn't the fixed costs for `units with limited lifetime` and `...age` also be accounted per period?

Agree and agree. Thank you for having a critical eye on this PR.

jokochems:
> > @p-snft Adding to what I stated above, I gave it some thought:
> > What I have included in this pull request and what was the main motivation is actually rather a bug fix to prevent the bias towards earlier investments in the optimization horizon.
> > Considering the cost change over time when evaluating an actual remaining value, to me is rather a feature. Thus, I suggest to leave it out here and open a dedicated pull request for that. Does this make sense to you?
> 
> I agree. Actually, my suggestion was rather meant as a quick fix: Instead of calculating the remaining value at the end of the time horizon, it would just have been left to the user to give it. Maybe, it can be considered the sloppy version of 2A.

Also thanks for commenting, @p-snft. I already have an implementation idea for 2A. It shouldn't be that difficult, but first, I'll adress the remainder.

jokochems:
> > @jokochems does the range for the fixed costs of the `_simple_flow_block` also need to be adjusted [here](https://github.com/oemof/oemof-solph/blob/6fb392bc15581011c8b247017f27125b1ef2c892/src/oemof/solph/flows/_simple_flow_block.py#L497)? And shouldn't the fixed costs for `units with limited lifetime` and `...age` also be accounted per period?
> 
> Agree and agree. Thank you for having a critical eye on this PR.

Well for units without a lifetime limit, it has been implemented per period instead of a per year basis. I changed this to a per year basis and limited it to the optimization horizon to be consistent. So it is actually the same as for the `_investment_flow_block`, but without the offset for the investment occuring in period p, i.e. at the start of period p. Again, thanks for pointing this out.

jokochems:
@nailend could you please re-review? @p-snft Could you add your review as well or accept in case @nailend finds nothing more that needs to be changed?

FYI: The failing docs test seems to be due to a failing docs test for a zenodo link. I assume it to be a temporary server problem for zenodo.

Thank you!

jokochems:
@nailend and @p-snft: I added additional functionality, namely allowing to re-evaluate assets at the end of the optimization horizon as an (optional) feature, as @p-snft suggested in the issue. Please have a look at this as well. Thank you!

p-snft:
I am just wondering if it is possible to move the economic calculations to https://github.com/oemof/oemof-tools/blob/dev/src/oemof/tools/economics.py. (That might be another PR, though.)

jokochems:
> I am just wondering if it is possible to move the economic calculations to https://github.com/oemof/oemof-tools/blob/dev/src/oemof/tools/economics.py. (That might be another PR, though.)

Sounds like a good idea, indeed for a future PR.

docs/usage.rst
```diff
@@ -1270,6 +1270,11 @@ Besides the `invest` variable, new variables are introduced as well. These are:
       has not yet been tested.
     * For now, both, the `timeindex` as well as the `timeincrement` of an energy system have to be defined since they
       have to be of the same length for a multi-period model.
+    * You can choose whether to re-evaluate assets at the end of the optimization horizon. If you set attribute
```
nailend, line 1273: That's a great feature!

docs/usage.rst
```diff
@@ -1270,6 +1270,11 @@ Besides the `invest` variable, new variables are introduced as well. These are:
...
+      `use_remaining_value` of the energy system to True (defaults to False), this leads to the model evaluating the
+      difference in the asset value at the end of the optimization horizon vs. at the time the investment was made.
+      The difference in value is added to or subtracted from the respective investment costs increment,
+      assuming assets are to be liquidated / re-evaluated at the end of the optimization horizon.
```
nailend, line 1277: That's a great feature!

nailend (approved):
I think it's great! Thanks a lot for all the changes and improvements!

src/oemof/solph/flows/_investment_flow_block.py
```diff
@@ -945,6 +1048,89 @@ def _objective_expression(self):
...
+                            m.flows[i, o].investment.offset[-1]
+                            - m.flows[i, o].investment.offset[p]
+                        )
+                    ) * (1 + m.discount_rate) ** (-end_year_of_optimization)
+                else:
+                    return (
+                        self.invest[i, o, p]
+                        * (remaining_annuity - original_annuity)
+                        * present_value_factor_remaining
+                    ) * (1 + m.discount_rate) ** (-end_year_of_optimization)
```
p-snft, line 1128: I think, code duplication can be avoided here.:
```suggestion
                convex_investment_costs = (
                    self.invest[i, o, p]
                    * (remaining_annuity - original_annuity)
                    * present_value_factor_remaining
                ) * (1 + m.discount_rate) ** (-end_year_of_optimization)
                if nonconvex:
                    return (
                        convex_investment_cost
                        + self.invest_status[i, o, p]
                        * (
                            m.flows[i, o].investment.offset[-1]
                            - m.flows[i, o].investment.offset[p]
                        )
                    )
                else:
                    return convex_investment_costs
```

p-snft (approved):
I checked for code style and if it breaks "stable" functionality. I see no real issues with that regard. There is no reason I see against the merge, I only have small remarks.

src/oemof/solph/flows/_investment_flow_block.py
```diff
@@ -809,28 +811,45 @@ def _objective_expression(self):
...
-                    \frac {(1+i)^l \cdot i} {(1+i)^l - 1} \cdot
+                    \frac {(1+i)^l \cdot i} {(1+i)^l - 1}\\
+                &\\
+                &
+                ANF(d, ir)=\frac {(1+ir)^d \cdot ir} {(1+ir)^d - 1}
 
-        It is retrieved, using oemof.tools.economics annuity function. The
-        interest rate is defined as a weighted average costs of capital (wacc)
-        and assumed constant over time.
+        They are retrieved, using oemof.tools.economics annuity function.
```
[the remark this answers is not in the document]
jokochems: Good hint. I will phrase it that way.

src/oemof/solph/flows/_flow.py
```diff
@@ -86,7 +86,8 @@ class Flow(on.Edge):
         :class:`~oemof.solph.flows._simple_flow_block.SimpleFlowBlock`.
     fixed_costs : numeric (iterable or scalar), :math:`c_{fixed}`
         The fixed costs associated with a flow.
-        Note: These are only applicable for a multi-period model.
+        Note: These are only applicable for a multi-period model
+        and given on a yearly basis.
```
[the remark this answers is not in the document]
jokochems: That has been a discussion in a very old issue #556. I generally like the idea as well, but did not do so, because my intention was to add functionality, not to take major design choices for the existing framework.

src/oemof/solph/flows/_investment_flow_block.py
```diff
@@ -809,28 +844,48 @@ def _objective_expression(self):
...
-            whereby:
-
-            * :math:`A(c_{invest,var}(p), l, ir)` A is the annuity for
-              investment expenses :math:`c_{invest,var}(p)` lifetime :math:`l`
-              and interest rate :math:`ir`
-            * :math:`DF=(1+dr)` is the discount factor with discount rate
-              :math:`dr`
-
-        The annuity hereby is:
+        whereby:
```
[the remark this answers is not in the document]
jokochems: I found "whereby" as a translation for the German "wobei", but I'm really indifferent to what is of the two is used.

src/oemof/solph/flows/_investment_flow_block.py
jokochems: Yes, your suggestion lacks the discounting for the offset, though (for the case you want to provide costs in nominal terms). But thanks a lot, I'll integrate it.

jokochems:
Thanks a lot for your reviews, suggestions and approvals, @nailend and @p-snft. I will merge this back.

PR #985, opened 2023-10-06
Feature/add thermal building model

Maxhi77:
This pull request aims to add a 5RC thermal building model to oemof. The building model is based on capacitors and resistances and is explained in detail in ISO 13790:2008. Building configurations for the 5RC model can be calculated based on the tabula database.

The idea is to extract all information from a tabula database CSV file and calculate a 5RC building configuration based on this information. For this purpose, functions to calculate solar gains through windows were added as well. Solar gains through windows are connected to window-specific values, which are extracted from the CSV file. For the calculation of this solar gains, a weather file was added. With the building configuration, a building model can be generated. The user has to set temperatures for heating and cooling in the building model.

grafik

Right now, the heat_flows goes into the model as input and cooling_flows as output. All in all the idea is to optimize the internal temperature of the building, called t_air, to use the building as thermal storage and save energy and cost. 

In this bullet points I want to summarize open questions / todos
- Should the calculation of solar gains be part of the model? (If yes)
  -  Data structure of the weather file
- building model heat flows as input and cooling flows as output, or both as input

ISO 13790:2008 “Energy performance of buildings –Calculation of energy use for space heating and cooling“. 
Tabula - https://webtool.building-typology.eu/#bm ,  https://episcope.eu/building-typology/tabula-webtool/


Maxhi77:
Closed and continued in https://github.com/[oemof/thermal-building-model](https://github.com/oemof/thermal-building-model)

PR #989, opened 2023-10-11
BUGFIX: GenericStorage investment initialization

Raul-Ciria-Aylagas:
The __init__ method was malfunctioning with the preferred way to declare investments in the attr nominal_storage_capacity. Probably due to code legacy from older versions.

The investment and nominal_storage_capacity attributes were mixed together and assigned in different parts of the initialization.


p-snft (approved):
You're right, this bug is due to an incomplete change to the new way. Thanks for fixing it!

Issue #979, opened 2023-09-20
Processing results fails in multi-period setup if timeincrement is not set


henhuy:
I tried to adapt storage investment example v1 into a multi-period system (see [gist](https://gist.github.com/henhuy/24fe9cd10880aa5af5a75adf9f4f8ab9)) - @oemof.solph==dev
At first I set up the energysystem like this:
```
t1 = pd.date_range("2022-01-01", periods=8760, freq="H")
t2 = pd.date_range("2023-01-01", periods=8760, freq="H")
tindex = t1.append(t2)

energysystem = solph.EnergySystem(
    timeindex=tindex,
    periods=[t1, t2],
    infer_last_interval=False,
)
```
which resulted in a length mismatch error in pandas when processing results (`results = solph.processing.results(om)`):
```
ValueError: Length mismatch: Expected axis has 17519 elements, new values have 17520 elements
```
pointing to line https://github.com/oemof/oemof-solph/blob/f02297ea7db0134bfff12c3ff6c7c909ca4405ed/src/oemof/solph/processing.py#L178

Only after setting `timeincrement` parameter like this:
```
t1 = pd.date_range("2022-01-01", periods=8760, freq="H")
t2 = pd.date_range("2023-01-01", periods=8760, freq="H")
tindex = t1.append(t2)

energysystem = solph.EnergySystem(
    timeindex=tindex,
    timeincrement=[1] * len(tindex),
    periods=[t1, t2],
    infer_last_interval=False,
)
```
postprecessing executed successfully.

Now I wonder, why this error occurs? 
Why oemof.solph cannot build `timeincrement` correctly internally?

jokochems:
Thanks, @henhuy, I'll have a look, but currently, I have quite a lot on my plate.

jokochems:
I started digging into this issue:
* Unfortunately, it is not so trivial. Thus, for now, I'd recommend to explicitly define the `timeincrement` attribute of an `EnergySystem` when working with a multi-period model.
* The reason for the discrepancy is the decision to split `TIMESTEPS` (the former index for the flowvariables) and `TIMEPOINTS` (the index for storage states) and alter the default behaviour of the `timeincrement` in this course. I think it has been with #811.
* I did not fully align the multi-period model with this new paradigm. Apparently, I also have been using a different definition for the `timeincrement`.
* If you do not specify it explicitly, the `timeincrement` has one entry less than your `TIMESTEPS` set (17519 instead of 17520). This leads to the code failing.
[...]

Issue #977, opened 2023-09-17
Multi-period investments are biased since the remaining value is not accounted for


jokochems:
## Background
In `v0.5.1`, multi-period investments have been introduced. For those not familiar with the concept: The idea is to allow multiple time steps where investments are allowed, namely at the beginning of every period, e.g. year, and to introduce discounting for costs that occur at different periods resp. time steps as well as perform life time tracking. So it is especially useful for long-term system planning.

## Problem description
Investments are accounted for by their annuities (annual equivalent of capital expenditures) over the full lifetime of a plant. So far all good since this is what you want to have. However, there is an implicit bias towards earlier investments introduced by this. It is not as attractive to invest in later periods as the plant might not run for its full lifetime, but still carries the total capital expenditures.

## Suggestion
Thus, it would be sensible to account for remaining values at the end of the optimization horizon. What I have in mind is to simply evaluate for each `InvestmentFlowBlock` - and the other blocks allowing for investments (storages, demand repose) - what the remaining lifetime at the end of the simulation is and to decrement the costs by the respective annuities. E.g., if you have a plant with 20 years lifetime that has been operating for 8 years at the end of the optimization horizon, subtract 12 times the annuity of that plant to account for its remaining value.

With the newly introduced `periods_matrix` of the energy system, see #957, it should be easier to introduce this approach.

@nailend and @p-snft
* What is your take on the general idea?
* I'd self-assign and provide a fix somewhat soon. I'd be happy for at least one of you to review.

nailend:
I'm happy to help with that and give a review. It's also on our to-do list, just hasn't moved up far enough yet.

p-snft:
Uh! How could we let this slip through? ;P

This should definitely be fixed. Alternatively to the proposed or as an additional aspect of the implementation, the investment object could take a value for the remaining value at the end of the optimization period, analogue to the value for the investment. The (possibly default) value would be the summed annuity of the remaining lifetime, but it should be possible to choose differently. For example, one might decide that the remaining value considers a learning curve, modelling the fact that later installations can be cheaper.

jokochems:
> Uh! How could we let this slip through? ;P
> 
[...]

PR #990, opened 2023-10-11
Fix/eliminate warnings

p-snft:
There are a lot of FutureWarnings partly from imported packages but also due to our own prepared API changes. This PR is meant to adapt to these (planned) changes.


docs/usage.rst
```diff
@@ -643,8 +643,8 @@ The following example pictures a Pumped Hydroelectric Energy Storage (PHES). Bot
 
     solph.components.GenericStorage(
         label='PHES',
-        inputs={b_el: solph.flows.Flow(investment= solph.Investment(ep_costs=500))},
-        outputs={b_el: solph.flows.Flow(investment= solph.Investment(ep_costs=500)},
+        inputs={b_el: solph.flows.Flow(nominal_value= solph.Investment(ep_costs=500))},
+        outputs={b_el: solph.flows.Flow(nominal_value= solph.Investment(ep_costs=500)},
```
fwitte, line 647: Is there a specific reason for the spaces after the first `=` signs?

src/oemof/solph/_energy_system.py
```diff
@@ -130,13 +130,6 @@ def __init__(
                     "conflicting to each other."
                 )
                 raise AttributeError(msg)
-            else:
-                msg = (
-                    "Ensure that your timeindex and timeincrement are "
-                    "consistent.\nIf you are not considering non-equidistant "
-                    "timeindices, consider only specifying a timeindex."
-                )
-                warnings.warn(msg, debugging.SuspiciousUsageWarning)
```
fwitte, line 139: I cannot say things about this, because I am not really knowledgable where this comes (or came) into place.

fwitte:
I'd say everything looks fine here, thank you!

One note on using `pytest.approx`: The checks for the integers seem fine, what about the others? Some compare to `0.1` deviation, bevore it was rounded to 1 decimal, which would be `0.05` right? That is really the only thing that could be changed if necessary.

fwitte (approved):

src/oemof/solph/_energy_system.py
p-snft: I deleted the message as the warning only showed up when using experimental features anyway. I think, @jokochems should know best if it should show up. (If so, it should be an `ExperimentalFeatureWarning`.)

docs/usage.rst
p-snft: Just typo, thanks.

```suggestion
        inputs={b_el: solph.flows.Flow(nominal_value=solph.Investment(ep_costs=500))},
        outputs={b_el: solph.flows.Flow(nominal_value=solph.Investment(ep_costs=500)},
```

p-snft:
> One note on using `pytest.approx`: The checks for the integers seem fine, what about the others? Some compare to `0.1` deviation, bevore it was rounded to 1 decimal, which would be `0.05` right? That is really the only thing that could be changed if necessary.

By default, `pytest.approx` takes a relative accuracy of 1e-6 and an absolute accuracy of 1e-12. If the test offers that accuracy, I did not weaken the test on purpose.

p-snft:
Decreased coverage is decreased because I deleted lines. I hope this is in order.

jokochems:
Thank you, @p-snft! I must admit that I overlooked this PR as I was quite busy with other developments. I'll add my review shortly.

src/oemof/solph/_energy_system.py
jokochems: Yes, I think it is important to be aware of this when setting up a multi-period model. Also, I think the second sentence does not hold as @henhuy found that both needs to be explicit currently (see #979). I'm fine with transferring it to an `ExperimentalFeatureWarning`, but would vote against a removal.

jokochems: > Yes, I think it is important to be aware of this when setting up a multi-period model. Also, I think the second sentence does not hold as @henhuy found that both needs to be explicit currently (see #979). I'm fine with transferring it to an `ExperimentalFeatureWarning`, but would vote against a removal.

Of course, the desired longer-term fix would be to properly address and resolve #979.

p-snft: I re-introduced the warning, now without the second sentence and as a `ExperimentalFeatureWarning`.

tests/test_models.py
```diff
@@ -73,7 +73,10 @@ def test_multi_period_default_discount_rate():
     warnings.filterwarnings("ignore", category=FutureWarning)
     timeindex = pd.date_range(start="2017-01-01", periods=100, freq="D")
     es = solph.EnergySystem(
-        timeindex=timeindex, periods={0: timeindex}, infer_last_interval=True
+        timeindex=timeindex,
+        timeincrement=[1] * len(timeindex),
+        periods=[timeindex],
+        infer_last_interval=False,
```
jokochems, line 79: I see that you where looking to explicitly set `infer_last_interval`. 

I noticed that this test was somewhat outdated since I used the old `periods` definition (type dict) here which I fixed in #977 which I just merged. 

Another thing is that though it was syntactically correct with the list attribute as in the current dev branch, it would not have worked to set up a model this way. Since the test is only to demonstrate a warning that occurs before the model setup, I'm sorry that have not detected this. Anyways, as it is now, also the model build would work. 

Concerning the (not so nice) inconsistency to be resolved some day, see #979.

jokochems (approved):
Thank you very much, @p-snft and sorry for me taking quite some time for the review. I'm happy to approve. Also, I resolved some conflicts that were introduced by my last feature and detected and resolved one bug in the course of doing so that was neither detected by the test nor by me (please see my comment on this one).

PR #993, opened 2023-10-12
Adapt cost accounting for multi-year-periods

### LINKED ISSUE #992: Use representative year for multi-year-periods in multi-period-investment
The yet implemented multi-period-investment was designed by the assumption to provide time series for every year (equivalent period) within the optimization horizon. Due to a long optimization horizon, we need to use representative years/time series for the years within a multi-year-period. 

Fortunately, @jokochems well-designed approach already facilitates most of this. Currently, only the variable costs for the flows are not handled accordingly. This could be done by adjusting values for the `objective weighting` but might collide with its use in the TSAM integration (as discussed in #987) which we also want to use. 

Therefore, I propose the  **boolean-flag** `use_representative_year_in_multi_year_periods` for `solph.EnergySystem` . 
This would lead to a further condition at the calculation for the `variable_costs`, [here](https://github.com/oemof/oemof-solph/blob/6fb392bc15581011c8b247017f27125b1ef2c892/src/oemof/solph/flows/_simple_flow_block.py#L466).

In case it's True, variable costs would be calculated like this:

```python
            for i, o in m.FLOWS:
                if m.flows[i, o].variable_costs[0] is not None:
                    if self.use_representative_year_in_multi_year_periods:
                        for p, timesteps in m.TIMESTEPS_IN_PERIOD.items():
                            variable_costs_increment = 0
                            for t in timesteps:
                                variable_costs_increment += (
                                    m.flow[i, o, p, t]
                                    * m.objective_weighting[t]
                                    * m.flows[i, o].variable_costs[t]
                                    * ((1 + m.discount_rate) ** - m.es.periods_years[p])
                                )

                            for pp in range(0, get_period_duration(p)):
                                variable_costs += (
                                    variable_costs_increment
                                    * ((1 + m.discount_rate) ** (-pp))
                                )
                    else:
                        ...
```

further `get_period_duration()` from #982 would need some adaption for multi-year-periods as well:

```python
def get_period_duration(self, period):
   
   if period == len(self.es.periods):
     duration = (
        m.es.periods[period+1].min().year
         - m.es.periods[period].min().year
         )
   else:
     duration =  (
           self.periods[period].max().year
            - self.periods[period].min().year
            + 1
           )

    return duration
```

nailend:
fix #992 

as desribed in urbs
https://urbs.readthedocs.io/en/latest/theorydoc/intertemp_prob.html?highlight=annuity#costs

- [x] variable costs
- [x] fixed costs
- [x] adjust `get_period_duration()`
- [x] flag `use_representative_year`
- [ ] <s>add base-year offset in case base_year not first representative year(e.g. 2021=first representativ year but base_year =2020) </s>
- [ ] Add test


p-snft:
At the November 2024 developer meeting, we decided to replace the current multi-period model by something more analogue to flows (annuities as cost for having the capacity) with "ramping constraints".

PR #1003, opened 2023-10-20
make BasicModel consider all constraint groups

### LINKED ISSUE #1002: BasicModel can ignore some constraint blocks in cellular models
When using the cellular modelling approach, the basic model considers only the constraint groups that are present in the first energy cell. This isn't a problem if all cells contain the same component types (that's why I didn't notice it earlier). But if, for example, only one cell contains a storage, the constraints of the storage are neglected.

lensum:
Closes #1002 by implementing a loop over all energy systems passed to the model class.

### DIFF
--- a/src/oemof/solph/_models.py
+++ b/src/oemof/solph/_models.py
@@ -132,12 +132,21 @@ class BaseModel(po.ConcreteModel):
             "constraint_groups", []
         )
 
-        self._constraint_groups += [
-            i
-            for i in self.es.groups
-            if hasattr(i, "CONSTRAINT_GROUP")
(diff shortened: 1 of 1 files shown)


jokochems:
Looks plausible and also like an important fix to introduce. Is it already finalized, @lensum?

lensum:
Yes, it's ready to be merged (from my pov).

jokochems (approved):
Thanks for the fix, @lensum. Does make sense to me.

lensum:
Thanks for the approval, @jokochems! Should I merge it or is it more "up to code" if someone else performs the merge?

PR #1006, opened 2023-10-31
Feature/integrate tsam seg objwei reviewed

henhuy:
Changes I made:
- I think it is not necessary to pass the whole TSAM aggregation result to oemof, that is why I only added `tsa_parameters` which holds `timesteps_per_period`, `order`, `occurrences` and `timeindex` per period. This is sufficient for building up all indexes the weighting and recreation of original timeindex in processing. This is why I removed your TSAM methods out of energysystem creation.
- TSAM weighting is now created in `Model` from `order` and `occurrences` from tsa_parameters
- I removed setting inter storage to zero in each period - but we can discuss that!

TODOS we should address in this PR:
- [x] apply tsam_weighting to full_load_hours
- [x] add test for invest/variable cost checks (this could be added to my new tsam tests in https://github.com/oemof/oemof-solph/blob/d13b2310cd5af84615e4d5fe7416b46fd9343272/tests/test_scripts/test_solph/test_tsam)

Afterwards we can merge this int my branch, fix other failing tests (there seem to be some bugs due to new feature), make linter happy and so on...


henhuy:
Removing TSAM aggregation in `helpers` module also aims to not add `tsam` package to dependencies

PR #1007, opened 2023-10-31
Allow for fixing investments

### LINKED ISSUE #1005: Fix investment results in repeated solving
## Problem statement
I want to fix investment results and re-optimize a model in order to be able to interpret dual values (as prices which I can't as long as investments are endogenously optimized). This could be achieved by fixing investment variables using the `.fix()` method from pyomo on them on a solved model instance (i.e. effectively transforming them into parameters) and then re-optimizing the model.

## Idea for implementation
* My idea would be to introduce a new boolean attribute `fix_investments` into the `Model` class.
* If this was set to `True` (`False` should be the default), a method to fix investment-related variables should be called by the dedicated child blocks of the model, i.e. `InvestmentFlowBlock`, `GenericInvestmentStorageBlock` resp. the SinkDSM investment blocks which iterates over the indices of the variables and fixes each value.
* In case no model solution is there yet, an Error should be raised.

In my head, this doesn't seem to be to hard to add to the current status quo.  I'll self-assign and try to provide the feature. 

What do you think, @p-snft, @simnh, @nailend and @uvchik? Am I overlooking something or do you think it is reasonable like I described?

## Outlook
In principle, I think a more general version of this feature could not be limited to the investment variables, but unfortunately, I do not have a use case nor capacities to tackle such a broad approach.

jokochems:
Closes #1005 

* Introduces an attribute `fix_investments` to class `Model`
* Iterates over investment flows resp. variables and fixes them.
* Raises an error in case model has not yet been solved.


jokochems:
This is a draft with yet untested functionality, just to get a first impression. I need to add tests as well as to touch the `GenericInvestmentStorageBlock` as well as the SinkDSM investment blocks.

jokochems:
@nailend and @p-snft: In my view, it is complete and functional. Would you please be so kind to review? Thank you! To check what it does, you may compare lp-files before and after shifting side by side and you'll notice how the investments are fixed, thus the respective variables are turned into parameters.

nailend (commented):
I haven't gone into full detail, but the code changes seem pretty simple and the LP file comprehensive (the fixed variable/parameter values are not explicitly in the LP file, are they?) Seems like an interesting feature, but I don't fully understand the benefit of analysing the duals yet. I will have to do some research on that. 
From my understanding, though, a test comparing the actual dispatch results (flows) before and after the fix would make sense.

p-snft:
Honestly, to me this feature suggests that it was possible to change other parts of the energy system as well. So, you optimise the investment assuming an energy demand, then change the energy demand in the second and re-run the optimisation.

jokochems:
> From my understanding, though, a test comparing the actual dispatch results (flows) before and after the fix would make sense.

I agree. That would make sense to highlight the differences.

jokochems:
> Honestly, to me this feature suggests that it was possible to change other parts of the energy system as well. So, you optimise the investment assuming an energy demand, then change the energy demand in the second and re-run the optimisation.

Well, yes. In principle, you would be free to alter other things before another solve of the model, I guess. This would be up to the user. My use case was just for interpreting the duals. (I'll comment on that in the issue as there was another question.)

Do you have any objection against the feature in it's current form or was this just concerning the naming or what you could do with it in princple?
I guess, it will most likely be something for expert's users only, anyways ...

p-snft:
> > Honestly, to me this feature suggests that it was possible to change other parts of the energy system as well. So, you optimise the investment assuming an energy demand, then change the energy demand in the second and re-run the optimisation.
> 
> Well, yes. In principle, you would be free to alter other things before another solve of the model, I guess. This would be up to the user. My use case was just for interpreting the duals. (I'll comment on that in the issue as there was another question.)
> 
> Do you have any objection against the feature in it's current form or was this just concerning the naming or what you could do with it in princple? I guess, it will most likely be something for expert's users only, anyways ...

My issue is that it adds complexity, requirements for maintenance and possibly technical dept for a rather limited use case. In particular, I dislike the fact that it's a global setting. Imagine you have a problem where you have a genuine investment decision (adding PV panels) but also a demand rate for electricity that incentives peak shaving. If I want to re-run the model with a fixed investment but different demand time series, it does not make sense to have the demand rate (i.e. maximum peak load from the grid) fixed.

So, I'd rather have something like
```python
for entity in energy_system.entities:
    entity.fix_investments()
```
but I will need to think about how this might be implemented. Maybe at the dev meeting. (Also, if it seems infeasible to have the more flexible solution, it would be okay to merge this one.)

jokochems:
> > > Honestly, to me this feature suggests that it was possible to change other parts of the energy system as well. So, you optimise the investment assuming an energy demand, then change the energy demand in the second and re-run the optimisation.
> > 
> > 
> > Well, yes. In principle, you would be free to alter other things before another solve of the model, I guess. This would be up to the user. My use case was just for interpreting the duals. (I'll comment on that in the issue as there was another question.)
> > Do you have any objection against the feature in it's current form or was this just concerning the naming or what you could do with it in princple? I guess, it will most likely be something for expert's users only, anyways ...
> 
> My issue is that it adds complexity, requirements for maintenance and possibly technical dept for a rather limited use case. In particular, I dislike the fact that it's a global setting. Imagine you have a problem where you have a genuine investment decision (adding PV panels) but also a demand rate for electricity that incentives peak shaving. If I want to re-run the model with a fixed investment but different demand time series, it does not make sense to have the demand rate (i.e. maximum peak load from the grid) fixed.
> 
> So, I'd rather have something like
> 
> ```python
> for entity in energy_system.entities:
>     entity.fix_investments()
> ```
> 
> but I will need to think about how this might be implemented. Maybe at the dev meeting. (Also, if it seems infeasible to have the more flexible solution, it would be okay to merge this one.)

Well, I don't entirely get what you are aiming at, but I'm open to a more generic / less globals approach. Let's discuss at the dev meeting. I can still install it from this feature branch in the meantime, as I need the feature short-notice for my modelling activities. Unfortunately, I'll only be able to join the dev meeting remotely, but I think, we can manage to have a call on this and I'm looking forward to that anyways.

jokochems:
Though the future of this feature is subject to debate, I came across an issue which might be of interest:
* I'm solving a really large model instance and have disactivated the crossover algorithm of the solver due to complexity and runtime reasons and use barrier algorithm for solving (which provides a super-close to optimum solution).
* Thus, one solution for my particular problem would be to round investment variables before a repeated solve. I think, I'll add that as an optional feature add-on (since I kind of require it super short notice).
* We can still discuss on how to proceed, @p-snft. I see your points concerning the maintability and structure issues (also reaching beyond this PR) - I just don't have the capacities at the moment to provide a more general approach.

p-snft:
I do see your personal need for this adjustment. However, I fear that this is introducing code dept for a very limited use case. There are several use cases that require a very similar functionality but would not be possible with this one:

* Optimise investment as well as operation including demand rate for electricity. Then fixing the investments but leaving the peak demand as a variable. This would not work as optimising the peak load is not distinguished from investments.
* Optimise day ahead energy trading based on a forecast. Then switch to slightly different energy demands and optimise intra-day trading.
* Run a multi-period investment model, then fix the sizes just of the technologies that have a long planning phase (but not e.g. small scale PV installations) to emulate multiple medium-term scenarios that an still work out once an initial decision is taken.

p-snft (changes requested):
I do see your personal need for this adjustment. However, I fear that this is introducing code dept for a very limited use case. There are several use cases that require a very similar functionality but would not be possible with this one:

* Optimise investment as well as operation including demand rate for electricity. Then fixing the investments but leaving the peak demand as a variable. This would not work as optimising the peak load is not distinguished from investments.
* Optimise day ahead energy trading based on a forecast. Then switch to slightly different energy demands and optimise intra-day trading.
* Run a multi-period investment model, then fix the sizes just of the technologies that have a long planning phase (but not e.g. small scale PV installations) to emulate multiple medium-term scenarios that an still work out once an initial decision is taken.

Unfortunately, I am also lacking the time to implement this in a more generic way. Thus, I propose to leave this open as a reference or until someone find the time to do it.

jokochems:
> I do see your personal need for this adjustment. However, I fear that this is introducing code dept for a very limited use case. There are several use cases that require a very similar functionality but would not be possible with this one:
> 
> * Optimise investment as well as operation including demand rate for electricity. Then fixing the investments but leaving the peak demand as a variable. This would not work as optimising the peak load is not distinguished from investments.
> * Optimise day ahead energy trading based on a forecast. Then switch to slightly different energy demands and optimise intra-day trading.
> * Run a multi-period investment model, then fix the sizes just of the technologies that have a long planning phase (but not e.g. small scale PV installations) to emulate multiple medium-term scenarios that an still work out once an initial decision is taken.
> 
> Unfortunately, I am also lacking the time to implement this in a more generic way. Thus, I propose to leave this open as a reference or until someone find the time to do it.

Hi, thanks for the response. I totally understand it and find it a good suggestion to simply leave it open. (Btw: I'm still facing some numerical issues which lead the model to be solved again to be infeasible. But maybe I just have to choose an even higher rounding precision for my particular use case.)

You did not request any particular code changes, did you? This was more on the general need for a revision and it to be more generic.

What would be a cool thing would be to provide a list of variables you want to fix for another run / units you want to exclude from fixing. That might also be dispatch-related ones and allow you to exclude dedicated investment variables as for the example you provided above.

p-snft:
Yes, I did not come up with specific requests.

> What would be a cool thing would be to provide a list of variables you want to fix for another run / units you want to exclude from fixing. That might also be dispatch-related ones and allow you to exclude dedicated investment variables as for the example you provided above.

If this was implemented, the approach should be general enough to be merged, I guess.

jokochems:
I currently do not have the capacities to tackle this Pull Request. Also, I faced numerical issues and infeasibilities when I wanted to test it in a large scale model. Thus, I'll close this pull request and the dedicated issue.

@oemof-developer Feel free to reopen in case someone has a use case or capacities to retackle it.

PR #1009, opened 2023-11-09
Assure Model.nodeis a list

p-snft:
It was assumed that `oemof.network.energy_system.nodes` is a list, but this is not granted. (There are plans to work with a dictionary in network.) Also, handling of cellular systems is slightly improved.


p-snft:
Coverage decreased because I deleted a tested line. This should be in order.

lensum:
Looks good to me! Thanks for the improvement.

lensum (approved):
Looks good to me! Thanks for the improvement.

PR #1011, opened 2023-11-14
Add multi-input-multi-output converter

henhuy:
This PR adds a new converter which can group inputs and outputs into multiple groups.
Inputs/outputs within a group can be exchangeable used to provide group flow.

Compare conversion_plus technology from calliope: https://calliope.readthedocs.io/en/v0.6.8/user/advanced_constraints.html#the-conversion-plus-tech

For each input/output group a new pyomo Variable/Index has to be set up.

### DIFF
--- a/tests/test_scripts/test_solph/test_mimo_converter/test_mimo_converter.py
+++ b/tests/test_scripts/test_solph/test_mimo_converter/test_mimo_converter.py
@@ -22,8 +22,8 @@ def test_invalid_flow_shares():
         ValueError, match="Invalid flow share types found: {'maxx'}"
     ):
         b_gas = Bus(label="gas")
-        b_hydro = Bus(label="gas")
-        b_electricity = Bus(label="gas")
+        b_hydro = Bus(label="hydro")
+        b_electricity = Bus(label="electricity")
         MultiInputMultiOutputConverter(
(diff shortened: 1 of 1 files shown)


p-snft:
I don't get why this is needed. Everything sketched at https://calliope.readthedocs.io/en/v0.6.8/user/advanced_constraints.html#the-conversion-plus-tech could be modelled using solph `Converter`s (sometimes additional `Bus`es might be needed). I would rather suggest to add a workflow to have sub-graphs grouping different nodes (and allow hiding the respective internal Flows).

henhuy:
Hey @p-snft - I see what you are suggesting and like the idea of a more general approach. Started an issue for that: #1015 
Maybe you could have a look, if I understood you correctly and my implementation idea is a good way to go.

p-snft:
I think this is equivalent to multiplying the conversion_factors of the converters that are put in series and is actually not needed. Maybe, a tutorial, a facade or a helper function would be an option?

Bachibouzouk:
During [oemof-dev meeting](https://github.com/oemof/oemof/wiki/Meeting-2025.02) it was suggested we could do this with a Facade, as we will discuss about possibility to bring an abstract Facade class into oemof-solph

PR #1012, opened 2023-11-15
Fix y intercept of OffsetConverter

### LINKED ISSUE #1010: Misleading/missing documentation of OffsetConverter
**Describe the bug**
In the [usage documentation for the OffsetTransformer](https://oemof-solph.readthedocs.io/en/stable/usage.html#offsetconverter-component) an example is given to calculate the coefficients `c0` and `c1`. If calculated like this, the results are illogical. I narrowed it down to the calculation of `c0`, which should actually  look like this: 
`c0 = (P_out_max-c1*P_in_max)/P_out_max`
Which means, that `c0` isn't really the y-intercept anymore (which would be `c0 = P_out_max-c1*P_in_max`).

**Possible solutions**
I see two possible solutions:
1) Fix the documentation to include the example calculations of the parameters described above, or
2) Change the calculation in the `OffsetConverterBlock` to not multiply with the `status_nominal` (which is the product of `P_max(t) * Y(t)`) but with the `status` variable. This way, `c0` would remain to be the y-intercept, which, from a mathematical, logical and usage standpoint seems much nicer to me.

Am I overlooking something here or doing something wrong? I'm actually a bit surprised that this has not been noticed before.

p-snft:
Fixes the `OffsetConverter` to work as documented.

Closes #1010

### DIFF
--- a/src/oemof/solph/components/_offset_converter.py
+++ b/src/oemof/solph/components/_offset_converter.py
@@ -245,7 +245,7 @@ class OffsetConverterBlock(ScalarBlock):
                             # `NonConvexFlow`.
                             try:
                                 expr += (
-                                    m.InvestNonConvexFlowBlock.status_nominal[
+                                    m.InvestNonConvexFlowBlock.status[
                                         n, o, t
                                     ]
                                     * n.coefficients[0][t]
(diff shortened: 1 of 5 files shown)


lensum (approved):
Thank you for the quick fix! I haven't looked into the .lp files in detail, but the changes you've made make sense to me.

PR #1021, opened 2023-11-22
Fix minimum up-/ downtime

### LINKED ISSUE #1017: Minimum uptime/ downtime
For minimum_uptime and minimum_downtime time steps are counted. However, time steps might have different lengths. This can lead to inconsistencies.

p-snft:
For minimum_uptime and minimum_downtime time steps were counted. However, time steps might have different lengths. This can lead to inconsistencies.

Fixes #1017

### DIFF
--- a/src/oemof/solph/_options.py
+++ b/src/oemof/solph/_options.py
@@ -247,6 +247,8 @@ class NonConvex:
         for attribute, value in custom_attributes.items():
             setattr(self, attribute, value)
 
+        self.first_flexible_timestep = max(minimum_uptime, minimum_downtime)
+
     @property
     def max_up_down(self):
         """Return maximum of minimum_uptime and minimum_downtime.
--- a/src/oemof/solph/_options.py
+++ b/src/oemof/solph/_options.py
@@ -13,6 +13,7 @@ SPDX-FileCopyrightText: Johannes Kochems
 SPDX-License-Identifier: MIT
 
 """
+from warnings import warn
 
 from oemof.solph._plumbing import sequence
 
@@ -259,4 +260,9 @@ class NonConvex:
         The maximum of both is used to set the initial status for this
         number of time steps within the edge regions.
         """
+        warn(
+            "Use of max_up_down is deprecated. "
+            "The attribute will be deleted in a future version.",
+            FutureWarning,
+        )
         return max(self.minimum_uptime, self.minimum_downtime)
--- a/src/oemof/solph/_options.py
+++ b/src/oemof/solph/_options.py
@@ -233,8 +233,8 @@ class NonConvex:
             custom_attributes = {}
 
         self.initial_status = initial_status
-        self.minimum_uptime = minimum_uptime
-        self.minimum_downtime = minimum_downtime
+        self.minimum_uptime = sequence(minimum_uptime)
+        self.minimum_downtime = sequence(minimum_downtime)
         self.maximum_startups = maximum_startups
         self.maximum_shutdowns = maximum_shutdowns
 
@@ -249,20 +249,6 @@ class NonConvex:
             setattr(self, attribute, value)
 
         if initial_status == 0:
-            self.first_flexible_timestep = minimum_downtime
+            self.first_flexible_timestep = self.minimum_downtime[0]
         else:
-            self.first_flexible_timestep = minimum_uptime
-
-    @property
-    def max_up_down(self):
-        """Return maximum of minimum_uptime and minimum_downtime.
-
-        The maximum of both is used to set the initial status for this
-        number of time steps within the edge regions.
-        """
-        warn(
-            "Use of max_up_down is deprecated. "
-            "The attribute will be deleted in a future version.",
-            FutureWarning,
-        )
-        return max(self.minimum_uptime, self.minimum_downtime)
+            self.first_flexible_timestep = self.minimum_uptime[0]
--- a/src/oemof/solph/_options.py
+++ b/src/oemof/solph/_options.py
@@ -247,7 +247,10 @@ class NonConvex:
         for attribute, value in custom_attributes.items():
             setattr(self, attribute, value)
 
-        self.first_flexible_timestep = max(minimum_uptime, minimum_downtime)
+        if initial_status == 0:
+            self.first_flexible_timestep = minimum_downtime
+        else:
+            self.first_flexible_timestep = minimum_uptime
 
     @property
     def max_up_down(self):
--- a/src/oemof/solph/_options.py
+++ b/src/oemof/solph/_options.py
@@ -199,10 +199,10 @@ class NonConvex:
         Integer value indicating the status of the flow in the first time step
         (0 = off, 1 = on). For minimum up and downtimes, the initial status
         is set for the respective values in the beginning e.g. if a
-        minimum uptime of four timesteps is defined and the initial status is 
-        set to one, the initial status is fixed for the four first timesteps 
-        of the optimization period. Otherwise if the initial status is set to 
-        zero and the first timesteps are fixed for the number of minimum 
+        minimum uptime of four timesteps is defined and the initial status is
+        set to one, the initial status is fixed for the four first timesteps
+        of the optimization period. Otherwise if the initial status is set to
+        zero and the first timesteps are fixed for the number of minimum
         downtime steps.
     negative_gradient_limit : numeric (iterable, scalar or None)
         the normed *upper bound* on the positive difference
--- a/src/oemof/solph/_options.py
+++ b/src/oemof/solph/_options.py
@@ -183,11 +183,11 @@ class NonConvex:
         from the actual output.
     inactivity_costs : numeric (iterable or scalar)
         Costs associated with not operating the flow.
-    minimum_uptime : numeric (1 or positive integer)
+    minimum_uptime : numeric or list of numeric (1 or positive integer)
         Minimum number of time steps that a flow must be greater then its
         minimum flow after startup. Be aware that minimum up and downtimes
         can contradict each other and may lead to infeasible problems.
-    minimum_downtime : numeric (1 or positive integer)
+    minimum_downtime : numeric or list of numeric (1 or positive integer)
         Minimum number of time steps a flow is forced to zero after
         shutting down. Be aware that minimum up and downtimes can
         contradict each other and may to infeasible problems.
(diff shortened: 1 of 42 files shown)


p-snft:
@AntonellaConsolinno, @gnn, @lensum : In a first step, I tried to simplify the constraint. I still need to adjust the summation index so that the old behaviour is reproduced.

src/oemof/solph/flows/_non_convex_flow_block.py
```diff
@@ -469,24 +475,22 @@ def min_downtime_rule(_, i, o, t):
...
-                expr += (
-                    self.status[i, o, t - 1] - self.status[i, o, t]
-                ) * m.flows[i, o].nonconvex.minimum_downtime
-                expr += -m.flows[i, o].nonconvex.minimum_downtime
-                expr += sum(
-                    self.status[i, o, t + d]
-                    for d in range(0, m.flows[i, o].nonconvex.minimum_downtime)
+            expr = 0
+            expr += (
+                self.status[i, o, t] - self.status[i, o, t + 1]
```
jokochems, line 480: Hi @p-snft. I like your solution draft. I was wondering whether this line here does run into an indexing error if you don't limit it to t < m.TIMESTEPS.at(-1) and formulate the constraint over the TIMESTEPS set of the model?!

AntonellaConsolinno: Hey, I was thinking about the same thing. Maybe you can also reduce the number TIMESTEPS for the Constraint buliding

p-snft: I see that t+1 should not exist for `t == m.TIMESTEPS.at(-1)`, however for some reason it created the constraint. For backward compatibility, I'd rather test `if t > 0` so that we can continue working with
`self.status[i, o, t - 1] - self.status[i, o, t]`.

src/oemof/solph/flows/_non_convex_flow_block.py
```diff
@@ -469,24 +471,25 @@ def min_downtime_rule(_, i, o, t):
             """
             Rule definition for min-downtime constraints of non-convex flows.
             """
-            if self._time_step_allows_flexibility(
-                t, m.flows[i, o].nonconvex.max_up_down, m.TIMESTEPS.at(-1)
-            ):
```
p-snft, line 474: Removing this calls for a different handling of the first time steps. Maybe, we continue to not create the constraint for the first N time steps so that the first time of switching down can continue to be our marker.

AntonellaConsolinno: or the first timesteps have to be fixed and afterwards implement only the constraints of the timesteps > uptime/downtime

p-snft: Just to have this more graphic. The old implementation interpreted the minimum uptime or downtime to force the flow up or down in the beginning and in the end.

Old implementation:
image

Current implementation:
image

Thinking about this a bit longer, there is a bug.

```python
pp2 = solph.components.Source(
    label="plant_min_up_constraints",
    outputs={
        bel: solph.Flow(
            nominal_value=10,
            min=0.5,
            max=1.0,
            variable_costs=10,
            nonconvex=solph.NonConvex(minimum_uptime=2, minimum_downtime=4, initial_status=1),
        )
    },
)
```

Results in the flow being forced active for four time steps. I think backwards compatibility wit this can be dropped, as its an obvious mistake.

image

src/oemof/solph/_options.py
```diff
@@ -247,6 +247,8 @@ def __init__(
         for attribute, value in custom_attributes.items():
             setattr(self, attribute, value)
 
+        self.first_flexible_timestep = max(minimum_uptime, minimum_downtime)
```
AntonellaConsolinno, line 250: Understanding you correct: The first timestep is forced to be the maximum. should it not be deepding on the initial_status?

src/oemof/solph/flows/_non_convex_flow_block.py
AntonellaConsolinno: The goal is only to force the up or down (depending on initial_status) only in the beginning. Or is there a reason, why it has to be forced to be up / down in the end?

In  my opinion the plot of "current implementation" is showing the correct behavor.

p-snft: The "current implementation" in this branch does only force the status up/down in the beginning. However, it does so for longer than needed (third plot and https://github.com/oemof/oemof-solph/pull/1021#discussion_r1407297634).

I think, it would make sense to allow forcing the status up/down in the end of the optimisation horizon as an equivalent to the "balanced" storage. If the storage has minimum downsteps of 4, it can have 2 in the end plus 2 in the beginning. But this should just be an option. The old implementation (in v0.5.1 and before) really made no sense.

src/oemof/solph/_options.py
p-snft: No. This is a direct translation from the old `NonConvex.max_up_down`: The first time step that allows flexibility is after the maximum of minimum_uptime and minimum_downtime. This does not make sense, though, as minimum_uptime should be relevant if the initial status is on, and minimum_downtime if the initial status is off.

src/oemof/solph/flows/_non_convex_flow_block.py
```diff
@@ -469,24 +474,29 @@ def min_downtime_rule(_, i, o, t):
...
                 return expr <= 0
             else:
-                expr = 0
-                expr += self.status[i, o, t]
-                expr += -m.flows[i, o].nonconvex.initial_status
-                return expr == 0
+                return Constraint.Skip
 
         return Constraint(
             self.MINDOWNTIMEFLOWS, m.TIMESTEPS, rule=min_downtime_rule
```
p-snft, line 502: Wouldn't something like
```suggestion
            self.MINDOWNTIMEFLOWS, m.TIMESTEPS[first_flexible_timestep:-1], rule=min_downtime_rule
```
work?

p-snft (commented):
I think it makes sense to first refactor the old code before extending it. So, here we go.

src/oemof/solph/flows/_non_convex_flow_block.py
AntonellaConsolinno: m.TIMESTEPS[first_flexible_timestep:-1] does not work, because flexible_timestep  can variate for each element of the set MINDOWNTIMEFLOWS. Further m.TIMESTEPS is not a list (anymore), so the access has to done be in different way, e.g. [t for t in m.TIMESTEPS if t>y]

p-snft: You are right. We have a 2D array of constraints, so the current way might be the best one.

p-snft:
The classical mode with constant number of steps seems to work in the refactored version. However, giving a series does not do what it should. I wrote a small script to investigate this:

```python
from oemof import solph

def test_min_uptime():
    # create an energy system
    idx = solph.create_time_index(2024, number=24)
    es = solph.EnergySystem(timeindex=idx, infer_last_interval=False)

    slack1 = solph.Bus(label="source", balanced=False)
    slack2 = solph.Bus(
        label="sink",
        inputs={
            slack1: solph.Flow(
                nominal_value=10,
                min=0.1,
                max=[
                    1.01, 1.02, 1.03, 1.04, 1.05, 1.06,
                    1.07, 1.08, 1.09, 1.10, 1.11, 1.12,
                    1.13, 1.14, 1.15, 1.16, 1.17, 1.18,
                    1.19, 1.20, 1.21, 1.22, 1.23, 1.24,
                ],
                variable_costs=[
                    0.01, 0.02, 0.03, -9, 0.05, 0.06,
                    0.07, 0.08, 0.09, 0.10, 0.11, -9,
                    0.13, 0.14, 0.15, -9, 0.17, 0.18,
                    0.19, 0.20, 0.21, -9, 0.23, 0.24,
                ],
                nonconvex=solph.NonConvex(
                    minimum_uptime=[
                        4, 3, 2, 1, 0, 2,
                        1, 0, 3, 2, 1, 0,
                        3, 2, 1, 5, 2, 1,
                        3, 2, 1, 3, 2, 1,
                    ],
                )
            ),
        },
        balanced=False,
    )

    es.add(slack1, slack2)
    om = solph.Model(es)
    om.solve(solver="cbc", solve_kwargs={"tee": False})
    results = solph.processing.results(om)

    data = solph.views.node(results, slack1)
    print(data)


if __name__ == "__main__":
    test_min_uptime()
```

src/oemof/solph/flows/_non_convex_flow_block.py
```diff
@@ -514,23 +527,30 @@ def _min_uptime_rule(_, i, o, t):
...
+                ) * m.flows[i, o].nonconvex.minimum_uptime[t]
                 expr += -sum(
-                    self.status[i, o, t + u]
-                    for u in range(0, m.flows[i, o].nonconvex.minimum_uptime)
+                    self.status[i, o, u]
+                    for u in range(
+                        t,
+                        min(
+                            t + m.flows[i, o].nonconvex.minimum_uptime[t],
+                            len(m.TIMESTEPS),
```
AntonellaConsolinno, line 547: I think the iteration has to go one timestep further.

src/oemof/solph/flows/_non_convex_flow_block.py
```diff
@@ -469,24 +474,32 @@ def min_downtime_rule(_, i, o, t):
...
+                expr += -m.flows[i, o].nonconvex.minimum_downtime[t]
                 expr += sum(
-                    self.status[i, o, t + d]
-                    for d in range(0, m.flows[i, o].nonconvex.minimum_downtime)
+                    self.status[i, o, d]
+                    for d in range(
+                        t,
+                        min(
+                            t + m.flows[i, o].nonconvex.minimum_downtime[t],
+                            len(m.TIMESTEPS),
```
AntonellaConsolinno, line 496: I think the iteration has to go one timestep further.

p-snft:
I keep forgetting that `range(items[-1])` will count to `items[-1]-1`. Thanks.

p-snft:
I think we should merge this already and prepare the calculation of the required steps for a given length in hours to a second PR.

AntonellaConsolinno (approved):
I proved only the code without investment. Maybe someone else can check this. The constraints itself I approve

jokochems (approved):
Seems plausible to me, thank you @p-snft and @AntonellaConsolinno. Admittedly, I only quickly scrolled through the code, but I like the additions also in terms of tests.

PR #1028, opened 2023-11-29
Remove duplicated discounting

### LINKED ISSUE #1019: Discounting duplication for fixed costs
There seem to be a duplication of discounting for fixed_costs [here](https://github.com/oemof/oemof-solph/blob/f542a7caf8e73efd03b9297ffda9e9bd326fa04b/src/oemof/solph/flows/_investment_flow_block.py#L1029
)
I will remove the second discounting as its already covered by the list comprehension. Do you agree?@jokochems ?

nailend:
close #1019

Remove duplicated discounting for fixed costs.

The discounting is already taken care of as the discount starts with the temporal distance of optimization start and investment period.

### DIFF
--- a/docs/whatsnew/v0-5-2.rst
+++ b/docs/whatsnew/v0-5-2.rst
@@ -23,6 +23,7 @@ Bug fixes
   bias towards investments happening earlier in the optimization horizon.
 * Fix bugs in multi-period documentation.
 * Fix y intersect of OffsetConverter
+* Fix duplicated discounting of fixed costs for multi-period investment
 
 Testing
 #######
--- a/src/oemof/solph/components/_generic_storage.py
+++ b/src/oemof/solph/components/_generic_storage.py
@@ -1869,7 +1869,7 @@ class GenericInvestmentStorageBlock(ScalarBlock):
                                 m.es.periods_years[p],
                                 range_limit,
                             )
-                        ) * (1 + m.discount_rate) ** (-m.es.periods_years[p])
+                        )
 
             for n in self.EXISTING_INVESTSTORAGES:
                 if n.investment.fixed_costs[0] is not None:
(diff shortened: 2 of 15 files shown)


nailend:
No changes in any tests needed....

jokochems:
> No changes in any tests needed....
> 
> Shouldn't this have an effect on the objective and therefore change the LP file for any Investment test with fixed costs?

Yes, it should. I was a bit shocked that there is no effect at first, but a brief glimpse into the code shed some light on it:
* This is not the only code occurence. It needs to be fixed for storages and SinkDSM as well.
* For the multi-period constraint tests, there are simply no tests, yet, that test fixed costs integration for investment flows.
* They are used in the "full-featured toy model" `test_multi_period_investment_model.py`, but the test does only check for the optimization outcome. Supposingly, the bug was "not that bad" that it led to a deviating optimization outcome in terms of investments taken.

Thus:
* I corrected it for the storage and SinkDSM as well. - Please check: It was a real quick'n'dirty search for code occurences.
* Would you mind adjusting the tests accordingly?
* I'm happy to add my review then, but as I mentioned in another issue, unfortunately, currently my time ressources are quite limited ... sorry about that and thanks a ton for detecting and fixing that issue!

nailend:
I updated the LP files and checked them roughly. I also added fixed costs to the multi-period converter tests!

nailend:
pytest is now failing due to changes in pyomo which sometimes prints zero integers instead of floats in the LP-files. It's only one LP-file, but not sure if this should be done separately as it's unrelated changes...  
I fixed it now in this PR as it was only one LP file @p-snft, what do you think?

p-snft:
I fixed the same thing in another branch. XD

jokochems (approved):
Thanks @nailend. I can approve it. Once the fix for the test has been included from that other feature branch (@p-snft Could you please add that or post a link to the PR you are referring to? I vaguely remember your efforts to align with the new pyomo lp writer ...), in my view it should be merged.

PR #1032, opened 2023-12-15
Revision/explicit network kwargs

p-snft:
Adapt to explicitly named arguments in oemof.network. Will need to be merged before the next release of network due to https://github.com/oemof/oemof-network/pull/40.

### DIFF
--- a/src/oemof/solph/components/_converter.py
+++ b/src/oemof/solph/components/_converter.py
@@ -117,7 +117,7 @@ class Converter(Node):
             label=label,
             inputs=inputs,
             outputs=outputs,
-            **custom_attributes,
+            custom_properties=custom_attributes,
         )
 
         if conversion_factors is None:
--- a/src/oemof/solph/components/_converter.py
+++ b/src/oemof/solph/components/_converter.py
@@ -101,13 +101,9 @@ class Converter(Node):
         conversion_factors=None,
         custom_attributes=None,
     ):
-        self.label = label
-
         if inputs is None:
-            warn_if_missing_attribute(self, "inputs")
             inputs = {}
         if outputs is None:
-            warn_if_missing_attribute(self, "outputs")
             outputs = {}
 
         if custom_attributes is None:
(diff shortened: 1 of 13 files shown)


p-snft:
When using the branch aimed to be merged in https://github.com/oemof/oemof-network/pull/40, I locally observe an issues I do not understand, yet:

```
tests/test_scripts/test_solph/test_lopf/test_lopf.py:112: in test_lopf
    om = Model(es)
src/oemof/solph/_models.py:386: in __init__
    super().__init__(energysystem, **kwargs)
src/oemof/solph/_models.py:151: in __init__
    self.flows = self.es.flows()
../network/src/oemof/network/energy_system.py:205: in flows
    return {
../network/src/oemof/network/energy_system.py:208: in <dictcomp>
    for target in source.outputs
E   AttributeError: 'ElectricalLine' object has no attribute 'outputs'
```

I do not see why this should be an error.  `ElectricalLine` is a special `Flow`, and as such should not have `outputs`.

p-snft:
> I do not see why this should be an error. `ElectricalLine` is a special `Flow`, and as such should not have `outputs`.

This is because the `ElectricalLine` is added to the `EnergySystem`. I changed this but also deactivated the test as the constraints are not built if it's not added as a Node. It's marked experimental, anyway.

p-snft:
Coverage decreases because experimental `ElectricalLine` is no longer tested.

src/oemof/solph/components/_converter.py
```diff
@@ -117,8 +113,12 @@ def __init__(
             label=label,
             inputs=inputs,
             outputs=outputs,
-            **custom_attributes,
+            custom_properties=custom_attributes,
```
nailend, line 116: @p-snft why aren't these parameter-names aligned? I guess it started [here](https://github.com/oemof/oemof-network/pull/40) Wouldn't it be best to change in `oemof.network` to `custom_attributes` as well, especially as `custom-attributes` are frequently used in the `oemof-tabular` API?!

p-snft: In Python, an attribute is a fixed term referring to a variable defined by the class in the scope of an object. The implementation that some though we had (but never had) in network added attributes

```python
my_node = Node("name", some_feature= 5)  # API was never implemented
my_node.some_feature # "some_feature" is an attribute, API was never implemented
# 5
```

However, the API now is
```python
my_node = Node("name", custom_properties={"some_feature": 5})
my_node.custom_properties["some_feature"]  # "some feature" is not an attribute
# 5
```

nailend: well `custom_properties` is an attribute though :laughing:

PR #1049, opened 2024-02-08
Replace the specification of frequency for hours in pandas data_range

### LINKED ISSUE #1046: Pandas gives a future warning for usage of "H" in a date_range
https://github.com/oemof/oemof-solph/blob/544ffc8ee21f257b475030bc4a1a41e27279d3cd/src/oemof/solph/_energy_system.py#L305

The capital `H` should be replaced by smaller case `h` here, I checked with the current minimum pandas version dependency (1.5.3), it does not seem to interfere with it. I will open an PR.

fwitte:
Closes #1046

### DIFF
--- a/.github/workflows/tox_pytests.yml
+++ b/.github/workflows/tox_pytests.yml
@@ -16,7 +16,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.9, "3.10", "3.11", "3.12"]
+        python-version: [3.9, "3.10", "3.11"]
 
     steps:
     - uses: actions/checkout@v1
--- a/.github/workflows/tox_pytests.yml
+++ b/.github/workflows/tox_pytests.yml
@@ -16,7 +16,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.9, "3.10", "3.11"]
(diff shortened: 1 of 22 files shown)


.github/workflows/tox_pytests.yml
```diff
@@ -16,7 +16,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.8, 3.9, "3.10"]
+        python-version: [3.9, "3.10", "3.11"]
```
p-snft, line 19:
```suggestion
        python-version: [3.9, "3.10", "3.11", "3.12"]
```
We should definitely support the current stable Python version, which is 3.12. I don't know, if we should even continue supporting Python 3.9: Debian 12 (current stable) has Python 3.11 already, Ubuntu 22.04 has Python 3.10 as default.

p-snft (approved):
This is a required hotfix. Maybe, supporting different versions of Python is something for another PR.

fwitte:
What is the current state here? I am getting tired of the pandas warnings... :D

p-snft:
Feel free to merge. (Didn't manage to get GitHub to test using Python 3.12 but that is a separate issue.)

fwitte:
Thank you, merged it!

PR #1051, opened 2024-02-09
Require pandas 2

p-snft:
I think it will increase maintainability if we do not have to care for compatibility with Pandas 1.x. (I used new features for a solph tutorial and people had unexpected problems because of unsupported features in their Pandas installation.)


p-snft:
From #1073 (by @fwitte):

- python 3.8 is supported until pandas 2.0.3 (we are at 2.2.2 currently)
- with python 3.8 and pandas 2.0.3 the function call of `pd.date_range()
  - **fails**: `pd.date_range("1/1/2024", periods=10, freq=f"0.5h")`
  - **succeeds**: `pd.date_range("1/1/2024", periods=10, freq=f"1.0h")`
  - **succeeds**: `pd.date_range("1/1/2024", periods=10, freq=f"0.5H")`
- with python 3.9/3.10/3.11 and pandas 2.0.3 the function call of `pd.date_range()` also has the same behavior as the python 3.8 version
- only, with pandas>=2.2 the `0.5h` will start working (which is a pandas version not supported by python 3.8)

> So what I actually did, when changing `"H"` to  `"h"` was under the hood require pandas >= 2.2, I think. How should we resolve this then? Requiring pandas>=2 I see no issue with in that context, we could even bring back python 3.8, but we would have to handle, how non-integer frequencies are dealt with by the `create_date_index` function.

henhuy:
Why don't we go forward and set python version to 3.9?
Soon there will be a version 3.13 - so why this extreme legacy support?

p-snft:
> Why don't we go forward and set python version to 3.9? Soon there will be a version 3.13 - so why this extreme legacy support?

We plan to not officially support Python 3.8 (no tests) but do not forbid using it.

p-snft:
Coverage decreases because of a pandas < 2.2 compatibility branch (needed esp. for Python 3.8). This, however would add a lot of overhead to test.

fwitte (approved):
Looks reasonable and the tests work, thank you!

PR #1054, opened 2024-02-16
Multiple Outputs for OffsetConverter

lensum:
This PR enables the OffsetConverter to output two (or more) flows with part-load depending efficiencies.
1. `outputs` are allowed to contain multiple busses.
2. `coefficients` expects a dictionary, where the keys are the output busses and the values are the tuples containing the necessary parameters. I took the liberty to implement a deprecation warning for the old behaviour (passing a tuple without specifying the output bus it corresponds to). This way, the OffsetConverter is more in line with the regular Converter and there is less room for confusion. But this is totally up for debate!
3. The issues discussed in https://github.com/oemof/oemof-solph/issues/1010 are fixed (misleading docu, usage guide and erroneous example).

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -748,6 +748,7 @@ which results in a nonlinear relation:
    :align: center
 
 The parameters :math:`C_{0}` and :math:`C_{1}` can be given by scalars or by series in order to define a different efficiency equation for every timestep.
+It is also possible to define multiple outputs.
 
 .. note:: See the :py:class:`~oemof.solph.components._offset_converter.OffsetConverter` class for all parameters and the mathematical background.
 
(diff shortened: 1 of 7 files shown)


lensum:
Can't figure out what problem `tox` is having, would be glad for a hint in case someone knows!

p-snft:
I'd suggest to close this and have an example that models the same thing using multiple components instead. @fwitte asked about an electrolyser and waste heat, but the example could also be a CHP with increasing electrical and reducing thermal efficiency towards full load.

fwitte:
I also quickly looked into the `OffsetConverter`, and I would now actually suggest implementing this feature (I already did locally before I found this PR)... It is quite straight forward, and it could be done using only a single binary variable.

p-snft:
I see. Then, the complementary `MultipleInputsOffsetConverter` should also be provided, I guess.

fwitte:
I will make another PR, I would like to change the naming conventions and styles of the equations (to have the exact same structure as in the `Converter`, i.e. using `conversion_factors`, which should be supplied by a dictionary using the keywords `offset` and `slope` or something like that). We can discuss there then :).

PR #1058, opened 2024-02-28
Feature/integrate tsam no multiple timegrid storage

Maxhi77:
Added possibility to have diverse storage with and without inter_storage_content in tsam mode.

For this purpose, multiple_time_grid has been added to generic_storage. The naming can be improved in the future.

### DIFF
--- a/src/oemof/solph/components/_generic_storage.py
+++ b/src/oemof/solph/components/_generic_storage.py
@@ -186,6 +186,7 @@ class GenericStorage(Node):
         lifetime_inflow=None,
         lifetime_outflow=None,
         custom_attributes=None,
+        multiple_tsam_timegrid=True,
     ):
         if inputs is None:
             inputs = {}
@@ -243,7 +244,7 @@ class GenericStorage(Node):
         self.invest_relation_output_capacity = invest_relation_output_capacity
         self.lifetime_inflow = lifetime_inflow
         self.lifetime_outflow = lifetime_outflow
-
+        self.multiple_tsam_timegrid = multiple_tsam_timegrid
         # Check number of flows.
         self._check_number_of_flows()
         # Check for infeasible parameter combinations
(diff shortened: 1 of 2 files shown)


henhuy:
Thanks @Maxhi77 - will have a look.

Maxhi77:
I just saw I forgot to add the possibility for investment, which I am going to do now.

Maxhi77:
I see a logical fault in my approach, I do not know to solve in this moment.

For a storage with a multiple_tsam_timegrid:
```
for n in group:
   if n.multiple_tsam_timegrid:
      self.storage_content_inter = Var(
          self.STORAGES, m.CLUSTERS_OFFSET, within=NonNegativeReals
      )
      self.storage_content_intra = Var(
          self.STORAGES, m.TIMEINDEX_TYPICAL_CLUSTER_OFFSET
      )
   else:
      self.storage_content_intra = Var(
          self.STORAGES, m.TIMEINDEX_TYPICAL_CLUSTER_OFFSET,
          bounds=_storage_content_bound_intra_rule
      )
```  

 When I understand oemof correctly, the GenericStorageBlock is for all storages the same. Right now the last storage of the group would define in which if statment is true, which affects all storages. In my case I want to define variables differently, if multiple_tsam_timegrid is True. All solutions I am thinking about are linked to a lot of changes.
 
 Do you have an idea?

henhuy:
You are right, it's not that simple.
I only see two options:
1. split storages internally like this is done for balanced storages in https://github.com/oemof/oemof-solph/blob/b00fd5f975dc72646e431d65bc07c650ace2bfd3/src/oemof/solph/components/_generic_storage.py#L474
In this case all variables would have to be changed whether they depend on your new generated Set or the original set.
2. Create a new block class which inherits from `GenericStorageBlock` and overwrite some of the initialization functions to your needs. Then in `GeneriStorage` at `https://github.com/oemof/oemof-solph/blob/b00fd5f975dc72646e431d65bc07c650ace2bfd3/src/oemof/solph/components/_generic_storage.py#L328` you have to define which block shall be used, depending on your boolean variable `multiple_tsam_timegrid`. This seems a bit easier IMO

Maxhi77:
Thanks for the answer. I had the first possibility in mind, but thought it is quite complex. 

The second one sounds easier, but might be not clean and fancied by reviewers. Therefore I am going to link them (@gnn @p-snft )  for their opinion.

jokochems:
Not sure if it affects this PR, but concerning the time indexing discussion, this was agreed upon as suggested by @p-snft in this issue: https://github.com/oemof/oemof-solph/issues/1081.

Maxhi77:
Closed because outdated, but Issue is published: [Issue 1214](https://github.com/oemof/oemof-solph/issues/1214)

PR #1060, opened 2024-03-01
Explicitely pass `custom_properties` to `Node` class

### LINKED ISSUE #1059: Using `custom_attributes` of `Source` and `Sink` classes fails
Currently, using `custom_attributes` of `Source` and `Sink` classes will lead to an exception. A similar issue was addressed in #1032, but the fix did not cover these classes. #929 Seems also to be related to this.

If there isn't more to it, I'll create a PR for a fix.

### Reproduction
Setup fresh venv, run only `pip install oemof-solph` to get the following dependencies:

```
blinker==1.7.0
dill==0.3.8
networkx==3.2.1
numpy==1.26.4
oemof.network==0.5.0
oemof.solph==0.5.2
oemof.tools==0.4.3
pandas==2.2.1
ply==3.11
Pyomo==6.7.1
python-dateutil==2.9.0
pytz==2024.1
six==1.16.0
tzdata==2024.1
```

Run the following script:

```python
from oemof.solph import components

components.Source(custom_attributes={"foo": "bar"})
components.Sink(custom_attributes={"foo": "bar"})
```

Get the exception:

```
  File "/Users/richartkeil/.../bug.py", line 3, in <module>
    components.Sink(custom_attributes={"foo": "bar"})
  File "/Users/richartkeil/.../.venv/lib/python3.12/site-packages/oemof/solph/components/_sink.py", line 66, in __init__
    super().__init__(label=label, inputs=inputs, **custom_attributes)
TypeError: Node.__init__() got an unexpected keyword argument 'foo'
```

(same for `_source.py`)

richartkeil:
This fixes #1059 by explicitly passing `custom_attributes` to the base class instead of unpacking them.

### DIFF
--- a/docs/whatsnew/v0-5-3.rst
+++ b/docs/whatsnew/v0-5-3.rst
@@ -13,6 +13,9 @@ Documentation
 Bug fixes
 #########
 
+* Fix custom attributes of Sink, Source, GenericCHP and Converter components
+  not being properly passed to base Node class
+
 Other changes
 #############
 
@@ -21,3 +24,5 @@ Known issues
 
 Contributors
 ############
+
+* Richard Keil
--- a/src/oemof/solph/components/_generic_chp.py
+++ b/src/oemof/solph/components/_generic_chp.py
@@ -131,7 +131,7 @@ class GenericCHP(Node):
     ):
         if custom_attributes is None:
             custom_attributes = {}
-        super().__init__(label, **custom_attributes)
+        super().__init__(label, custom_properties=custom_attributes)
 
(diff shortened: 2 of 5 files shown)


p-snft (approved):
Thanks for fixing this. You might want to add this fix and your contribution to [whatsnew/v0-5-3.rst](https://github.com/richartkeil/oemof-solph/blob/fix/custom-attributes/docs/whatsnew/v0-5-3.rst) before I merge.

richartkeil:
Great, should be ready to merge now 👍

PR #1041, opened 2024-01-24
Add new tutorial example (using NamedTuple as label)

p-snft:
This aims to be a tutorial for beginners that already introduces cocepts that until now are mostly used by experienced users.

PR #1061, opened 2024-03-06
Improved variable naming and formatting in 'basic_example.py' for enh…

MedinavonLevetzow:
The primary objective of this pull request is to offer new users a clearer understanding of the example and improve readability for all.

=> **Changed File**: 'basic_example.py'

- Refactored code to streamline by extracting unnecessary steps into dedicated functions, ensuring clarity and focus on the framework's core example. 
- Enhanced variable naming for readability, particularly important for newcomers.

*   Other comments and questions

- Could you please confirm that the enhanced variable names adequately describe their purpose?
- - Like the BATTERY_STORAGE which was introduced as a constant value, there were several places that just use the String "storage".
- Any suggestions for changes are welcome

### DIFF
--- a/examples/basic_example/basic_example.py
+++ b/examples/basic_example/basic_example.py
@@ -78,7 +78,7 @@ from oemof.solph import helpers
 from oemof.solph import processing
 from oemof.solph import views
 
-BATTERY_STORAGE = "battery_storage"
+STORAGE_LABEL = "battery_storage"
 
 
 def get_data_from_file_path(file_path: str) -> pd.DataFrame:
--- a/examples/basic_example/basic_example.py
+++ b/examples/basic_example/basic_example.py
@@ -71,7 +71,7 @@ from oemof.tools import logger
 from oemof.solph import EnergySystem
 from oemof.solph import Model
 from oemof.solph import buses
-from oemof.solph import components as solver_components
+from oemof.solph import components
 from oemof.solph import create_time_index
 from oemof.solph import flows
 from oemof.solph import helpers
--- a/examples/basic_example/basic_example.py
+++ b/examples/basic_example/basic_example.py
@@ -107,7 +107,7 @@ def plot_figures_for(element: dict) -> None:
 
 def main():
     # *************************************************************************
-    # ********** PART 1 - Define and optimize the energy system ***************
+    # ********** PART 1 - Define and optimise the energy system ***************
     # *************************************************************************
 
     # Read data file
@@ -245,7 +245,7 @@ def main():
     # Optimise the energy system and plot the results
     ##########################################################################
 
-    logging.info("Optimize the energy system")
+    logging.info("Optimise the energy system")
 
     # initialise the operational model
     energysystem_model = Model(energysystem)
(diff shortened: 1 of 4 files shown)


examples/basic_example/basic_example.py
```diff
@@ -71,28 +71,49 @@
 from oemof.solph import EnergySystem
 from oemof.solph import Model
 from oemof.solph import buses
-from oemof.solph import components as cmp
+from oemof.solph import components as solver_components
```
p-snft, line 74: It's actually not the "solver" components. I'd either not give an alias name
```suggestion
from oemof.solph import components
```
or just `import oemof.solph` and explicitly name the submodes in the code.

examples/basic_example/basic_example.py
```diff
@@ -71,28 +71,49 @@
...
+        ncol=2,
+    )
+    figure.subplots_adjust(top=0.8)
+    plt.show()
+
 
 def main():
     # *************************************************************************
-    # ********** PART 1 - Define and optimise the energy system ***************
+    # ********** PART 1 - Define and optimize the energy system ***************
```
p-snft, line 110: We agreed to use British English in our project.

examples/basic_example/basic_example.py
```diff
@@ -71,28 +71,49 @@
...
 from oemof.solph import buses
-from oemof.solph import components as cmp
+from oemof.solph import components as solver_components
 from oemof.solph import create_time_index
 from oemof.solph import flows
 from oemof.solph import helpers
 from oemof.solph import processing
 from oemof.solph import views
 
+BATTERY_STORAGE = "battery_storage"
```
p-snft, line 81: This is actually the label of the storage. It doesn't really matter if it's  battery or not, but I like defining the constant instead of having the string in the code several times.
```suggestion
STORAGE_LABEL = "battery_storage"
```

examples/basic_example/basic_example.py
```diff
@@ -71,28 +71,49 @@
...
+
+
+def get_data_from_file_path(file_path: str) -> pd.DataFrame:
```
p-snft, line 84: We do not use type hints in other locations, but I like them and we have to start somewhere.

p-snft (changes requested):
Hi @MedinavonLevetzow,

thanks for your valuable proposals. We actually know that the examples are a bit of a weak spot, so I started working on a new tutorial-like example a while ago (https://github.com/oemof/oemof-solph/pull/1041). Still, as this one exists, it can really profit from clarifications.

I changed the PR to merge into the existing revision, so that we can ask for a common review of another dev once I approve your changes. (See inline comments.)

Cheers,
Patrik

examples/basic_example/basic_example.py
MedinavonLevetzow: Hi Patrik, thanks for the headsup - I will introduce your suggestion on this

examples/basic_example/basic_example.py
MedinavonLevetzow: That wasn't quite clear to me, thank you for the clarification.

MedinavonLevetzow: I will rollback the changes to the British English

p-snft (approved):
Looks good now. Thanks.

PR #1063, opened 2024-03-25
Make invest relations iterable

nailend:
Currently the invest relations can not change in between the periods of a multi-period investment. I dont see a reason why this shouldnt be possible, so I just change the few necessary parts. 

@jokochems was there a specific reason why you did not adapt this as well or did you just not need it?


SabineHaas:
Only black and coveralls fails, as in all the PRs. 

@jokochems could you review the adaptions in this PR please?

p-snft (approved):
To me, this looks reasonable. I approve, as it shouldn't break single-period optimization. I'd also be glad to hear @jokochems if it makes sense to have this for multi-period models.

PR #1066, opened 2024-05-16
Delete unreachable code

p-snft:
Code checked if investment and nominal_value are both given. As investment is now handed to nominal value, this can no longer happen.


p-snft:
Black complains about code I did not touch. Will merge.

PR #1073, opened 2024-05-30
Merge configurations into `pyproject.toml`

### LINKED ISSUE #712: Define version at a single location
Currently, the version is defined in the following locations:
* VERSION
* docs/conf.py
* setup.py
* `src/oemof/solph/__init__.py`

It would simplify the release process, if all but one would import from the other one.

### LINKED ISSUE #1072: Aggregate all sorts of tools configurations in `pyproject.toml`
Having many configuration files (many of the files starting with `.` in the root) as well as the `setup.cfg`, `setup.py` and `pyproject.toml` at the same time really increases maintenance effort, since you always have to search multiple files to find what you are trying to change. Therefore, I suggest

- we move everything possible into `pyproject.toml`,
- delete `setup.py` and `setup.cfg` and
- put all configurations not available for `pyproject.toml` which were formerly in `setup.cfg` into their respective configuration files e.g. `.flake8` as it seems

On top of that, I would like to critically review, what of all those things is actually required and take out the trash. E.g. the `pep257` thing in `setup.cfg` seems ancient.

fwitte:
Resolve #1072 
Resolve #712


fwitte:
@oemof-developer: I think this should be ready for review. There are two remaining questions:

- Is anybody knowledgeable if `pydocstyle` actually does anything?
- The black error is in some source files I never touched, should that be fixed or not?

Quick explanation/overview of the changes:

- `setup.cfg` (deleted)
  - flake8 related stuff moved to new file: `.flake8`
  - `pep257` deleted, seems to be the same as `pydocstyle`
  - everything else moved to `pyproject.toml`
- `setup.py`, `MANIFEST.in` (all deleted)
  - completely migrated to `pyproject.toml`
- `.prospector` (deleted)
  - same as `pydocstyle` and `pep257`, does not seem to affect any test

fwitte:
One more comment to the build system: I changed from `setuptools` to `flit`, because `setuptools` based on `pyproject.toml` did not support `pip install -e` last time I checked, and also because I am using it in other tools, where I migrated to `pyproject.toml`, e.g. `tespy`.

pyproject.toml
```diff
@@ -1,3 +1,94 @@
...
+    "Operating System :: Unix",
+    "Operating System :: POSIX",
+    "Operating System :: Microsoft :: Windows",
+    "Operating System :: MacOS",
+    "Operating System :: OS Independent",
+    "Programming Language :: Python",
+    "Programming Language :: Python :: 3",
+    "Programming Language :: Python :: 3.8",
+    "Programming Language :: Python :: 3.9",
+    "Programming Language :: Python :: 3.10",
```
p-snft, line 52: Tox tests Python 3.9, 3.10, and 3.11.

p-snft (approved):
First of all, you questions:

> Is anybody knowledgeable if pydocstyle actually does anything?

I don't but I think this is nothing to be fixed here. If it does not do anything, it should be removed independend from the switch to pyproject.toml.

> The black error is in some source files I never touched, should that be fixed or not?

This is an unrelated problem, as it should have been wrong before you changed things. So, it should not be included in this merge request.

The same is true for the last comment I gave. Making support for Python 3.11 (which is tested) official might be something for a separate merge.

pyproject.toml
fwitte: The minimal supported Python version according to the specifications of the dependencies is 3.8 I just noticed. With the fix in the pandas H to h for the frequency (do not know how to search for merges on GH mobile app... It was the most recent PR merged to dev), that compatibility is actually broken via pandas. I am not sure, how the procedure is to change the minimum supported version of python.

fwitte:
One more question, what is  
the VERSION file for?

pyproject.toml
p-snft: There is no real procedure. Also, technically we do not break compatibility by removing support for outdated versions of Python, we just don't provide new updates. (If Pandas 2.x requires Python 3.9+, we might include this change in #1051.)

fwitte: I did a bit of research:

- python 3.8 is supported until pandas 2.0.3 (we are at 2.2.2 currently)
- with python 3.8 and pandas 2.0.3 the function call of `pd.date_range()
  - **fails**: `pd.date_range("1/1/2024", periods=10, freq=f"0.5h")`
  - **succeeds**: `pd.date_range("1/1/2024", periods=10, freq=f"0.5h")`
- with python 3.9/3.10/3.11 and pandas 2.0.3 the function call of `pd.date_range()` also has the same behavior as the python 3.8 version
- only, with pandas>=2.2 the `0.5h` will start working (which is a pandas version not supported by python 3.8)

So what I actually did, when changing `"H"` to  `"h"` was under the hood require pandas >= 2.2, I think. How should we resolve this then? Requiring pandas>=2 I see no issue with in that context, we could even bring back python 3.8, but we would have to handle, how non-integer frequencies are dealt with by the `create_date_index` function.

fwitte: After a coffee, I suggest:

- we can leave python 3.8 support, but already prompt people that it is untested and will be removed with the next major release
- we come up with a fix for the `create_date_index` to work with pandas >= 2 but pandas < 2.2
- we require pandas >= 2

The pandas required stuff can be done in #1051. I will add the prompt for the depreciation of python3.8 in this pr.

pyproject.toml
```diff
@@ -1,3 +1,94 @@
...
+    ".flake8",
+    ".pep8speaks.yml",
+    ".readthedocs.yaml",
+    ".scrutinizer.yml",
+]
+exclude = ["docs/_build"]
+
+[project]
+name = "oemof.solph"
+dynamic = ["version"]
```
fwitte, line 32: This reads the version number from the `__version__` variable of the project, one could also make the `description` dynamic, that would be the module docstring in the projects top level `__init__.py`.

PR #1075, opened 2024-06-04
Fix non-existing variable names

### LINKED ISSUE #1070: The basic_example.py is broken
It seems https://github.com/oemof/oemof-solph/blob/dev/examples/basic_example/basic_example.py is broken on dev since about two months. It seems the variable naming went wrong there in https://github.com/oemof/oemof-solph/pull/1061. I kind of think, that ChatGPT was confused by `solph.components` and made up `solver_components` ;).

p-snft:
There were some misnamed variables, probably introduced by "AI programming". I fixed them.

Closes #1070

### DIFF
--- a/examples/basic_example/basic_example.py
+++ b/examples/basic_example/basic_example.py
@@ -160,7 +160,7 @@ def main(dump_and_restore=False):
 
     # create excess component for the electricity bus to allow overproduction
     energysystem.add(
-        solver_components.Sink(
+        components.Sink(
             label="excess_bus_electricity",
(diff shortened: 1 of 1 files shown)


p-snft:
The Black issues are in code I did not touch.

fwitte:
Thank you for fixing it. The results plots look strange:

Figure_1

Figure_2

p-snft:
> Thank you for fixing it. The results plots look strange:

The problem is that default data (for one time step) was created when the data file is not found. The script now fails in that case.

PR #1077, opened 2024-06-05
Refine time index of Flows

p-snft:
As different types of periods will be used for multi-period optimisation (long-term investment planing) and working with aggregated time-series (e.g. using tsam), I suggest to change the indexing. This PR implements Flows that use timesteps [int] instead of timeindex currently ([int, int]).


p-snft:
Maybe, we should split this into several PRs and merge this as a first step.

src/oemof/solph/components/_offset_converter.py
```diff
@@ -365,16 +366,14 @@ def __init__(
         inputs,
         outputs,
         label=None,
-        conversion_factors=None,
-        normed_offsets=None,
+        coefficients=None,
```
fwitte, line 369: Is the deletion and addition here intended? Or some kind of relict from a merge?

src/oemof/solph/components/_offset_converter.py
```diff
@@ -365,16 +366,14 @@ def __init__(
...
         custom_attributes=None,
     ):
         super().__init__(
             label=label,
             inputs=inputs,
             outputs=outputs,
-            conversion_factors=conversion_factors,
-            normed_offsets=normed_offsets,
+            coefficients=coefficients,
```
fwitte, line 376: Deletion and insertion intended here?

src/oemof/solph/flows/_non_convex_flow_block.py
```diff
@@ -743,17 +743,11 @@ def _positive_gradient_flow_constraint(_):
                             m.flow[
                                 i,
                                 o,
-                                m.TIMEINDEX[index][0],
-                                m.TIMEINDEX[index][1],
+                                m.TIMESTEPS[index],
```
fwitte, line 746: Kind of not directly related, but why does line `740` iterate over a numeric `index` instead of `t` as is done everywhere else? I guess there is a good reason for it...

src/oemof/solph/flows/_non_convex_flow_block.py
```diff
@@ -790,33 +782,28 @@ def _positive_gradient_flow_constraint(_):
         def _negative_gradient_flow_constraint(_):
             r"""Rule definition for negative gradient constraint."""
             for i, o in self.NEGATIVE_GRADIENT_FLOWS:
-                for index in range(1, len(m.TIMEINDEX) + 1):
-                    if m.TIMEINDEX[index][1] > 0:
+                for index in range(1, len(m.TIMESTEPS) + 1):
+                    if m.TIMESTEPS[index] > 0:
```
fwitte, line 786: Same here, why is there no iteration on the timesteps but over the index?

fwitte:
Thank you for the work @p-snft, since the tests are working things look fine for me. However, I am not very knowledgeable yet on how multi-period would work and cannot really check the changes for content. Also, I did not check the lp-files changes... That is simply too much, so #1079 is really welcome! :)

src/oemof/solph/components/_offset_converter.py
p-snft: Eventually, I made a mistake when merging. That would mean that i also made another one when reviewing #1068. (The `OffsetTransformer` as a transitional wrapper should offer the old API, shouldn't it?)

fwitte: It does: https://github.com/oemof/oemof-solph/blob/80298509a7e8e485cd2b2c35cf2fcfa0fcfbf399/src/oemof/solph/components/_offset_converter.py#L117-L119, there also is a test: https://github.com/oemof/oemof-solph/blob/80298509a7e8e485cd2b2c35cf2fcfa0fcfbf399/tests/test_components/test_offset_converter.py#L420

fwitte: And, in this branch it is also available: https://github.com/oemof/oemof-solph/blob/revision/refine_timeindex/tests/test_components/test_offset_converter.py

Therefore, I guess that those keywords do not have any effect (!?)...

src/oemof/solph/flows/_non_convex_flow_block.py
p-snft: To my understanding, this is because of the `positive_gradient_constr` needing the `TIMEINDEX` (including periods) to be allowed to consider investment (periods) in the ramping. (@jokochems might now better.) There is probably a more elegant way to do this, but I did not want to touch this now.

src/oemof/solph/flows/_non_convex_flow_block.py
p-snft: Here, I do not see the point I made before. Maybe, the current implementation is faulty?

src/oemof/solph/components/_offset_converter.py
p-snft: As the keywords `conversion_factors` and `normed_offsets` are not part of the old API, I thought they were in the merge by mistake (or because I changed something). In my opinion, it does not make sense to have an `OffsetTransformer` with these keywords.

fwitte: Ah that is on the transformer... I overlooked that, then all is fine! :)

src/oemof/solph/flows/_non_convex_flow_block.py
fwitte: I mean it works, it was just a side comment. I would say: merge the feature and maybe also make a release to ship the latest changes to the users.

src/oemof/solph/flows/_non_convex_flow_block.py
fwitte: Maybe also something to be discussed in a separate issue to not delay the progress too much

fwitte (approved):
All remaining issues are side issues, which can be handled in a different PR

jokochems (approved):
Thank you, @p-snft!

It has always bugged me that the multi-period feature brought along this inconsistency between indexing flows and other variables. 

I thought about indexing every not investment-related variable with the `TIMEINDEX` set, but getting rid of the `TIMEINDEX` and indexing flows in `TIMESTEPS` instead, seems more user-friendly and intuitive to me. I just did not come up with that because I seem to have been too routine-blinded.

Preserving periods where needed can be achieved by using the `TIMEINDEX` set for iterations, as you did. Well actually you just did not touch upon that and kept it as is.

Long story short: I am happy to approve this as well. I only had one minor suggestion for renaming.

P.S.: It has been a while since I looked into the code base, but I still have the feeling of understanding it, which is a very good thing. ;-)

src/oemof/solph/processing.py
```diff
@@ -277,8 +277,8 @@ def results(model, remove_last_time_point=False):
         )
         for bus, timeindex in grouped:
             duals = [
-                model.dual[model.BusBlock.balance[bus, p, t]]
-                for _, p, t in timeindex
+                model.dual[model.BusBlock.balance[bus, t]]
+                for _, t in timeindex
```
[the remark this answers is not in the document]
p-snft: It should be a timeindex, as it contains something (we don't care about) an the time step `_, t`. Or am I wrong?

src/oemof/solph/components/_generic_storage.py
```diff
@@ -1258,7 +1258,7 @@ def _storage_investvar_bound_rule(block, n, p):
         i = {n: [i for i in n.inputs][0] for n in group}
         o = {n: [o for o in n.outputs][0] for n in group}
 
-        reduced_periods_timesteps = [(p, t) for (p, t) in m.TIMEINDEX if t > 0]
+        reduced_periods_timeindex = [(p, t) for (p, t) in m.TIMEINDEX if t > 0]
```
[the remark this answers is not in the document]
p-snft: To my understanding, this removes all indexes with t <= 0. If we can assume the index to be sorted ans starting with 0, this can be more efficient, sure.

src/oemof/solph/flows/_non_convex_flow_block.py
p-snft: If it is really the periods, I did something wrong, as `m.TIMESTEPS[p]` does not make sense.

src/oemof/solph/components/_generic_storage.py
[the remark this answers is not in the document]
jokochems: Okay, I think, this can be resolved as it was rather a remark, not a demand for a change.

jokochems:
@p-snft I'm fine with you addressing the two comments according to your preferences (as I already approved it) and then merging it rightaways. Thank you very much!

src/oemof/solph/processing.py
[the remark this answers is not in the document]
p-snft: With some days of distance: You are right.

p-snft:
As the gradient constraint has problems anyway, I will not touch it in this PR.

PR #1082, opened 2024-06-12
Remove cellular modelling stub

p-snft:
The approach to model cellular systems was experimental in incomplete, also did not allow to have more then two levels. The same models can be created without code in solph. So, it might be reasonable to delete it.


p-snft:
Decreasing coverage due to removing code should be accepted. I think this is done already.

lensum (approved):
Looks good to me! Thanks for cleaning this up!

PR #1086, opened 2024-06-25
Release v0.5.3

p-snft:
See docs/whatsnew/v0-5-3.rst for list of changes.


p-snft:
This version includes pinning of numpy. Without it, solph appears to be broken due to incompatibility of Pyomo and numpy >= 2.0.0. Thus, I will release without additional reviews of the whole package on top of the feature reviews we already have.

PR #1088, opened 2024-07-02
Allow sinks/sources to have multiple inputs/outputs

p-snft:
There was a warning and a comment that having multiple inputs/outputs for one Sink/Source is unintended but should be possible. I think we can remove that "strong recommendation" as it just works.

To not let beginners fall into the trap of having a dangling Sink or Source, like the old warning also might have prevented, inputs and outputs are now explicitly named arguments without a default.

### DIFF
--- a/docs/whatsnew/v0-5-4.rst
+++ b/docs/whatsnew/v0-5-4.rst
@@ -4,6 +4,8 @@ v0.5.4
 API changes
 ###########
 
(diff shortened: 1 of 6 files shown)


p-snft:
@uvchik gave positive feedback on another channel and suggested to have tests to explicitly make this part of the API.

PR #1089, opened 2024-07-03
Pandas warnings fixes

uvchik:
A lot of inputs in solph can be single values or sequences.

Recently a `FutureWarning` from pandas warns that Series will not work in the future. A small change in our plumbing module will keep the old behaviour and avoid the `FutureWarning`


uvchik:
My IDE complains if the docstring is not correct. I know, this should be a different PR. Please forgive my laziness 😄

src/oemof/solph/_plumbing.py
```diff
@@ -43,7 +54,7 @@ def sequence(iterable_or_scalar):
     if isinstance(iterable_or_scalar, abc.Iterable) and not isinstance(
         iterable_or_scalar, str
     ):
-        return iterable_or_scalar
+        return np.array(iterable_or_scalar)
```
p-snft, line 57: There is a problem with this, as `sequence` is used to create sequences of practically infinite length from scalar values.  If you return a numpy.array instead, the following will happen:

```python
# current implementation
s = sequence(sequence([5]))
s[0] = 5
s[5] = 5

# your suggestion
s = sequence(sequence([5]))
s[0] = 5
s[5] -> out of range
```

p-snft (changes requested):
I see that this feels like a quick fix, but unfortunately it is not. I see two options:

1. Cast to numpy.array when it is really needed.
2. Forbid sequences as arguments to sequences. (There might be further implications I do not see at the moment.)

src/oemof/solph/components/_offset_converter.py
```diff
@@ -282,12 +282,12 @@ def normed_offset_and_conversion_factors_from_coefficients(
...
-                slope += [c0]
-                offset += [c1]
+                slope.append(c0)
+                offset.append(c1)
 
             if max_len == 1:
-                slope = sequence(slope[0])
-                offset = sequence(offset[0])
+                slope = slope[0]
+                offset = offset[0]
```
fwitte, line 291: Thank you for jumping in here @p-snft. I was actually wrapping my head around this sequence stuff the first time I saw it when implementing the changes for the `OffsetConverter` and it took me quite a while to figure out what was going on. I see the practical point of the sequences, but I was wondering, if we could not just use scalars or `numpy` arrays instead in general. Because if you (by accident) provide a half length sequence (with respect to the timeindex) you would end up with some kind of nonsense anyway, right? So why not let people either pass scalars or correct-length lists or arrays?

p-snft: The problem is that we do not know the correct length before the Entity is added to the EnergySystem. An option would be to just let scalar stay scalars until we need to have arrays.

PR #1090, opened 2024-07-09
Introduce variable for storage losses

p-snft:
I think it makes a lot of sense, e.g. for adding custom constraints, to have an explicit variable for the losses of the storage. Note that the formula for the relative lasses is more complicated
```
losses(t) = energy(t) * (1-(1-b)**tau))
energy(t+1)= energy(t)-losses(t)
``` 
vs
```
energy(t+1) = energy(t) * (1-b)**tau
```
but I think the additional variable and constraint is useful and the 1-(1-b)**tau instead of (1-b)**tau changes a value in the lp file, so it will not significantly increase computational time.

PS: I existentially pushed this change to dev, so this PR reverts the revert.

### DIFF
--- a/examples/storage_balanced_unbalanced/storage.py
+++ b/examples/storage_balanced_unbalanced/storage.py
@@ -67,7 +67,7 @@ def storage_example():
 
     # create an energy system
     idx = pd.date_range("1/1/2017", periods=len(timeseries), freq="h")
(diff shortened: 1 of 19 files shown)


Maxhi77:
The possibility to add custom constraint for the losses looks advantageous to me

PR #1092, opened 2024-07-22
Display inline math correctly and hint towards the usage of the inbuilt slope and offset calculation methods

fwitte:
Basically what the title says.

There was an issue, where the inline math inside the NOTE block did not render correctly. On top of that, the API documentation of the `OffsetConverter` now also hints towards the usage of the inbuilt methods to calculate slope and offset.


p-snft (approved):
This is a good improvement.

PR #1094, opened 2024-07-31
Fix/storage costs for initial level

p-snft:
* For optimisation using an assumed cyclic time, it does not make sense to consider both, the 0th and N+1st time point. This can be seen in particular as the storage is typically balanced, so it is really one value that gains twice the weight.
* If time is assumed linear (no balanced storage), there are N+1 points in time for N time steps. Thus, the implementation makes sense to me. However, typically the initial content is predefined, so it just gives an offset to the objective value.

Thus, I suggest to drop the costs for the 0th time step.

Targeted for the next "major" release, because it can change the behaviour in some cases.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -9,6 +9,7 @@ These are new features and improvements of note in each release
     :backlinks: top
 
 
+.. include::  whatsnew/v0-6-0.rst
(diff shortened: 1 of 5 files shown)


jokochems (approved):
Thanks. Looks sound to me @p-snft.

PR #1096, opened 2024-08-02
Formulate full_load_time docstring specific for energy

p-snft:
This way, it should be much easier to understand.


jokochems (approved):
Yes, that is more precise. Imho can be merged if nobody contradicts. Also clearly visible in the constraint test you wrote on that.

PR #1097, opened 2024-08-02
Allow lower limit for generic_integral_limit

p-snft:
If it's possible to give an upper limit, a lower limit should also be posible.


Maxhi77 (changes requested):
>>> model = solph.constraints.generic_integral_limit(
    ...     model, "my_factor", flow_with_keyword, **limit**=777)
    
    Has to be changed to upper and lower limit. The rest looks good in my opinion.

Maxhi77 (approved):

PR #1101, opened 2024-08-12
Replace _Sequence by _FakeSequence

### LINKED ISSUE #1100: Chained sequence yields unexpected result
Casting to sequences can give unexpected results, as it sometimes returns a (fixed length) array, sometimes a (flexible) _Sequence:

```python
# previous implementation
s = sequence(sequence([5]))
s[0] = 5
s[5] = 5

# current implementation
s = sequence(sequence([5]))
s[0] = 5
s[5] -> out of range
```

The suggested solution is to not track `_Sequence.highest_index` in `_Sequence.__get_item__`. This way, the `_Sequence` does not have a defined length if it not explicitly given.

PS: We learned this in #1089.

p-snft:
The auto-stretch feature of the _Sequence contained hidden magic, as it tracked the last "entry" that has been accessed and pretended to have a length based on this. So, outcomes depend on something that is assumed to be a read access. Also, self.data was erived from UserList but never really supported all of the expected functionality in a consistent way.

The new _FakeSequence is a lot simpler. In particular it has an explicit (fake) length that needs to be set before casting to (stretched) lists. The default length is not set, because member functions using the same names as in `numpy.array` can help to work without an iterator.

Fixes #1100.

### DIFF
--- a/docs/whatsnew/v0-5-4.rst
+++ b/docs/whatsnew/v0-5-4.rst
@@ -21,6 +21,8 @@ Bug fixes
 Other changes
 #############
 
+* Refined internal sequence generation. (For e.g. constraint formulations,
+  Scalars are internally mapped to fake sequences without a defined length.)
 
 Known issues
 ############
--- a/src/oemof/solph/_plumbing.py
+++ b/src/oemof/solph/_plumbing.py
@@ -128,3 +128,7 @@ class _FakeSequence:
             return np.full(length, self._value)
         else:
             return np.full(len(self), self._value)
+
+    @property
+    def value(self):
+        return self._value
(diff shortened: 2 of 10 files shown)


uvchik:
> TODO: solph.processing still expects the auto-stretching to have happened.

Where is auto-stretching needed and why are all tests fine? I would expect a failing test that will be fine if we have fixed that issue.

p-snft:
> > TODO: solph.processing still expects the auto-stretching to have happened.
> 
> Where is auto-stretching needed and why are all tests fine? I would expect a failing test that will be fine if we have fixed that issue.

I fixed that already (in [3e94ccc](https://github.com/oemof/oemof-solph/pull/1101/commits/3e94cccf4cfb38233feaac6055306f8ca1432162)).

uvchik (approved):
Looks fine to me and in my opinion things are clearer now.

PR #1111, opened 2024-08-21
Remove internal discounting functionality

p-snft:
Discounting functionality is not actually needed, as we allow time series anyway. So, we can remove code without reducing functionality.


p-snft:
This removal was agreed on. As stated above, this just externalizes complexity and does not remove functionality.

PR #1113, opened 2024-08-22
Fix/numpy2 compatibility

p-snft:
It's almost impossible to have doctests that work with both, numpy 1.x and numpy 2.x: The string representations changed to include the data type. Example:

```
> data = numpy.array([1,2,3])
> print(data[1])  # no longer prints "2"
np.int64(2)
```

Introducing a lower bound to Pandas as well allows to remove Pandas <2.2 compatibility code.


p-snft:
Tests fail just because of broken link check (GitHub seems to be locked out.) Will merge.

PR #1114, opened 2024-08-22
Revision/rename nominal value

p-snft:
I think, the capacity of a `Flow` (e.g. a power line) is actually easier to interpret as the "nominal value". Actually, I have been asked that quite some times. If we have a `nominal_capacity` of a `Flow`, it would be arguable why  it's `nominal_storage_capacity` for the `GenericStorage`. So we need to say either `nominal_flow_capacity` which is actually redundant. In my opinion, it makes sense to call both `nominal_capacity`.

In fact, they are quite similar. In particular, they both accept a number (fixed capacity) or an Investment object.


examples/tuple_as_labels/tuple_as_label.py
```diff
@@ -225,18 +227,24 @@ def main():
...
             conversion_factors={bel: 0.58},
         )
     )
 
     # create storage object representing a battery
     nominal_storage_capacity = 5000
     storage = comp.GenericStorage(
-        nominal_storage_capacity=nominal_storage_capacity,
+        nominal_capacity=nominal_storage_capacity,
         label=Label("storage", "electricity", "battery"),
```
jokochems, line 241:
```suggestion
        nominal_capacity=nominal_capacity,
        label=Label("storage", "electricity", "battery"),
```

examples/tuple_as_labels/tuple_as_label.py
```diff
@@ -225,18 +227,24 @@ def main():
...
-            outputs={bel: flows.Flow(nominal_value=10000, variable_costs=50)},
+            outputs={
+                bel: flows.Flow(nominal_capacity=10000, variable_costs=50)
+            },
             conversion_factors={bel: 0.58},
         )
     )
 
     # create storage object representing a battery
     nominal_storage_capacity = 5000
```
jokochems, line 238:
```suggestion
    nominal_capacity = 5000
```

examples/tuple_as_labels/tuple_as_label.py
```diff
@@ -225,18 +227,24 @@ def main():
...
+        nominal_capacity=nominal_storage_capacity,
         label=Label("storage", "electricity", "battery"),
-        inputs={bel: flows.Flow(nominal_value=nominal_storage_capacity / 6)},
-        outputs={bel: flows.Flow(nominal_value=nominal_storage_capacity / 6)},
+        inputs={
+            bel: flows.Flow(nominal_capacity=nominal_storage_capacity / 6)
+        },
+        outputs={
+            bel: flows.Flow(nominal_capacity=nominal_storage_capacity / 6)
+        },
```
jokochems, line 247:
```suggestion
        inputs={
            bel: flows.Flow(nominal_capacity=nominal_capacity / 6)
        },
        outputs={
            bel: flows.Flow(nominal_capacity=nominal_capacity / 6)
        },
```

src/oemof/solph/components/_generic_storage.py
```diff
@@ -207,20 +208,34 @@ def __init__(
...
                 warn(msg, FutureWarning)
             nominal_storage_capacity = investment
         # --- END ---
+        # --- BEGIN: The following code can be removed for versions >= v0.6 ---
+        if nominal_storage_capacity is not None:
+            msg = (
+                "For backward compatibility,"
+                " the option nominal_storage_capacity overwrites the option"
+                + " nominal_capacity."
+                + " Both options cannot be set at the same time."
```
jokochems, line 223: Minor remark: The usage of string concatenation is inconsistent here. I'd suggest to remove the pluses as they are not necessarily needed.

src/oemof/solph/flows/_flow.py
```diff
@@ -176,25 +179,40 @@ def __init__(
             full_load_time_min = summed_min
         # --- END ---
 
+        # --- BEGIN: The following code can be removed for versions >= v0.7 ---
+        if nominal_value is not None:
+            msg = (
+                "For backward compatibility,"
+                " the option nominal_value overwrites the option"
+                " nominal_capacity."
+                + " Both options cannot be set at the same time."
```
jokochems, line 188: Same as above mentioned with the plus here.

tests/test_components/test_storage.py
```diff
@@ -340,3 +344,28 @@ def test_invest_content_maximum():
...
+    with pytest.raises(AttributeError, match="nominal_storage_capacity"):
+        bus = solph.Bus()
+        _ = solph.components.GenericStorage(
+            nominal_storage_capacity=5,
+            nominal_capacity=5,
+            inputs={bus: solph.Flow()},
+            outputs={bus: solph.Flow()},
+        )
+
+
```
jokochems, line 370: I think the blank lines might be removed.

tests/test_flows/test_simple_flow.py
```diff
@@ -48,10 +48,24 @@ def test_full_load_time_min():
...
         full_load_time_min=4.5,
     )
     flow_result = list(_run_flow_model(flow)["flow"][:-1])
 
     assert flow_result == pytest.approx(4 * [2] + [1] + 5 * [0])
+
+
+# --- BEGIN: The following code can be removed for versions >= v0.7 ---
+def test_nominal_capacity_warning():
+    with pytest.warns(FutureWarning, match="nominal_capacity"):
```
jokochems, line 62: I guess it would be more concise to match for `nominal_value` as for the storage above, where the test checked for the previous attribute `nominal_storage_capacity`.

tests/test_scripts/test_solph/test_storage_investment/test_storage_investment.py
```diff
@@ -120,7 +125,7 @@ def test_optimise_storage_size(
         label="pp_gas",
         inputs={bgas: solph.flows.Flow()},
         outputs={
-            bel: solph.flows.Flow(nominal_value=10e10, variable_costs=50)
+            bel: solph.flows.Flow(nominal_capacity=10e10, variable_costs=50)
```
jokochems, line 128:
```suggestion
            bel: solph.flows.Flow(nominal_capacity=1e11, variable_costs=50)
```

tests/test_scripts/test_solph/test_variable_chp/test_variable_chp.py
```diff
@@ -126,7 +130,7 @@ def test_variable_chp(filename="variable_chp.csv", solver="cbc"):
     energysystem.add(
         solph.components.Converter(
             label=("fixed_chp", "gas"),
-            inputs={bgas: solph.flows.Flow(nominal_value=10e10)},
+            inputs={bgas: solph.flows.Flow(nominal_capacity=10e10)},
```
jokochems, line 133:
```suggestion
            inputs={bgas: solph.flows.Flow(nominal_capacity=1e11)},
```

tests/test_scripts/test_solph/test_variable_chp/test_variable_chp.py
```diff
@@ -136,7 +140,7 @@ def test_variable_chp(filename="variable_chp.csv", solver="cbc"):
     energysystem.add(
         solph.components.ExtractionTurbineCHP(
             label=("variable_chp", "gas"),
-            inputs={bgas: solph.flows.Flow(nominal_value=10e10)},
+            inputs={bgas: solph.flows.Flow(nominal_capacity=10e10)},
```
jokochems, line 143:
```suggestion
            inputs={bgas: solph.flows.Flow(nominal_capacity=1e11)},
```

jokochems (commented):
@p-snft I had a look and introduced some fixes and suggestions.

I have some doubts concerning the renaming for the storage which is a consequence of the fact that the English term "capacity" similarly to the German term "Kapazität" can have different meanings.

For storages, we have a value in MWh, while for flows, we have a value in MW or whatever energy resp. power unit. Both are named "capacity", though, which might be slightly misleading, esp. for unexperienced users.

Maybe, one idea would be to rename to something like `nominal_energy` for storages to make the physical dimensions clearer.

But anyways, feel free to decide on how to proceed with the renaming.

src/oemof/solph/components/_generic_storage.py
```diff
@@ -207,20 +208,34 @@ def __init__(
...
                 raise AttributeError(msg)
             else:
                 warn(msg, FutureWarning)
             nominal_storage_capacity = investment
         # --- END ---
+        # --- BEGIN: The following code can be removed for versions >= v0.6 ---
+        if nominal_storage_capacity is not None:
+            msg = (
+                "For backward compatibility,"
+                " the option nominal_storage_capacity overwrites the option"
```
p-snft, line 221:
```suggestion
                + " the option nominal_storage_capacity overwrites the option"
```

src/oemof/solph/components/_generic_storage.py
p-snft: The plus helps to see that the line continues the last one. Thus, I added the missing one.

src/oemof/solph/flows/_flow.py
```diff
@@ -176,25 +179,40 @@ def __init__(
             full_load_time_min = summed_min
         # --- END ---
 
+        # --- BEGIN: The following code can be removed for versions >= v0.7 ---
+        if nominal_value is not None:
+            msg = (
+                "For backward compatibility,"
+                " the option nominal_value overwrites the option"
+                " nominal_capacity."
```
p-snft, line 187:
```suggestion
                "For backward compatibility,"
                + " the option nominal_value overwrites the option"
                + " nominal_capacity."
```

tests/test_components/test_storage.py
p-snft:
```suggestion

```

p-snft:
> @p-snft I had a look and introduced some fixes and suggestions.
> 
> I have some doubts concerning the renaming for the storage which is a consequence of the fact that the English term "capacity" similarly to the German term "Kapazität" can have different meanings.
> 
> For storages, we have a value in MWh, while for flows, we have a value in MW or whatever energy resp. power unit. Both are named "capacity", though, which might be slightly misleading, esp. for unexperienced users.
> 
> Maybe, one idea would be to rename to something like `nominal_energy` for storages to make the physical dimensions clearer.
> 
> But anyways, feel free to decide on how to proceed with the renaming.

Storages only have energy, Flows only have power. Thus, I think harmonisation is more beneficial. Also, "energy" should not be a keyword. In fact, we do not enforce units. For example, you could have a storage with a certain mass of hydrogen or wood pallets. It still represents energy, but the unit does not match (we are off by a factor c² XP).

tests/test_components/test_storage.py
p-snft: Black really wants those.

p-snft:
The issue has been open for a year now, and nobody objected or even came up with a better idea. Thus, I will merge this now.

PR #1119, opened 2024-08-29
Add v0.5.5 fixes to dev

p-snft:
This will typically allow to iterate over the `_FakeSequence` for built models.


p-snft:
As we do not want the changes from dev in v0.5, we need a new branch, first.

Issue #1056, opened 2024-02-20
Results processing should check if optimisation succeeded


p-snft:
It would be easier to pin down issues like #1045 if we checked the status of the optimisation before collecting the results.

p-snft:
Might be a topic for https://github.com/oemof/oemof/issues/124

PR #1139, opened 2024-11-22
Add kwarg fail_on_infeasable in model.solve

esske:
Added a new keyword to `model.solve()`: `allow_nonoptimal`, which is `False` on default.

If  `allow_nonoptimal` is False an error message is raised if the model does return a nonoptimal soltion, such as `infeasable` or `unbounded`.

Implements #1056.


3LL3RM4NN:
I like the idea of having this new keyword in the `solve()` function. 
However, how do you define `nonoptimal`?
From [this pyomo blogpost](http://www.pyomo.org/blog/2015/1/8/accessing-solver), I would say there are nonoptimal termination conditions and failed termination conditions like infeasible. 
In my opinion there should be a differentiation.

Additionally, I dont get why solver status and termination condition are accessed like this:
`status = solver_results["Solver"][0]["Status"]` and `termination_condition = solver_results["Solver"][0]["Termination condition"]`
and not this:
`status = solver_results.Solver.Status` and `termination_condition = solver_results.Solver.Termination_condition`

p-snft:
> I like the idea of having this new keyword in the `solve()` function. However, how do you define `nonoptimal`? From [this pyomo blogpost](http://www.pyomo.org/blog/2015/1/8/accessing-solver), I would say there are nonoptimal termination conditions and failed termination conditions like infeasible. In my opinion there should be a differentiation.

Thanks for your feedback. The feature is targeted at novice users. We expect that they assume a result to be optimal by default. Thus, `model.solve()` would fail on everything else - which is unexpected behaviour in these cases. If you are fine with a different solution you get e.g. after a time limit is reached, you should also be able to handle the differentiation yourself.

> Additionally, I dont get why solver status and termination condition are accessed like this: `status = solver_results["Solver"][0]["Status"]` and `termination_condition = solver_results["Solver"][0]["Termination condition"]` and not this: `status = solver_results.Solver.Status` and `termination_condition = solver_results.Solver.Termination_condition`

I like the API that works without the `[...]` operators. It has been like that since 2017 (see https://github.com/oemof/oemof-solph/commit/4fefc4bb71f36ac6b33d0193180a5fd546bea283). The message, however, does not explain why it is done this way and not differently. (Interestingly, the commit claims to fix  #292, which it did not really. I guess, we would not have #1056 if it really introduced the promised error raising.) Long story short: It seems to make sense to change this.

esske:
I changed the access of solver status and termination condition as suggested and wrote test to check if infeasable and unbounded problems are caught.

I noticed that if the the 'tee=True' argument of pyomo is set. There is an additional warning from Pyomo in the form of 

```
WARNING: Loading a SolverResults object with a warning status into
model.name="Model";
    - termination condition: infeasible
    - message from solver: Model was proven to be infeasible.
 ```
 
Seems a bit redundant, but I am not sure if we should supress this warning?

**Comment by p-snft** [comment]
> I noticed that if the the 'tee=True' argument of pyomo is set. There is an additional warning from Pyomo (...)
> Seems a bit redundant, but I am not sure if we should supress this warning?

The redundant warning was there before. Now we fail after the Pyomo warning, if not set otherwise. I won't worry too much, as this is a clear improvement. (We could decide not to warn if both, tee and returning without an optimal solution, are set, but that's a separate topic from this PR.)

File: tests/test_models.py [hunk 1]
```diff
@@ -17,7 +17,7 @@
 from oemof import solph
 
 
-def test_infeasible_model():
+def test_infeasible_model_warning():
```
p-snft, line 20:
```suggestion
def test_infeasible_model():
```

tests/test_models.py
```diff
@@ -34,9 +34,53 @@ def test_infeasible_model():
...
     )
     m = solph.Model(es)
-    with warnings.catch_warnings(record=True) as w:
-        m.solve(solver="cbc")
-        assert "Optimization ended with status" in str(w[0].message)
+    with warnings.catch_warnings(record=False) as w:
+        m.solve(solver="cbc", allow_nonoptimal=True)
+        assert "The solver did not return an optimal solution" in str(
+            w[0].message
+        )
```
p-snft, line 41: IMHO, it makes sense to also switch to modern pytest usage.

```suggestion
        with pytest.warns(UserWarning, match="The solver did not return an optimal solution"):
            m.solve(solver="cbc", allow_nonoptimal=True)
```

tests/test_models.py
```diff
@@ -34,9 +34,53 @@ def test_infeasible_model():
...
+        solph.components.Source(
+            outputs={bel: solph.flows.Flow(variable_costs=-5)}
+        )
+    )
+    m = solph.Model(es)
+
+    try:
+        m.solve(solver="cbc", allow_nonoptimal=False)
+    except Exception as e:
+        assert "The solver did not return an optimal solution" in str(e)
```
p-snft, line 83:
```suggestion
    with pytest.raises(RuntimeError, match="The solver did not return an optimal solution"):
        m.solve(solver="cbc", allow_nonoptimal=False)
```

tests/test_models.py
```diff
@@ -34,9 +34,53 @@ def test_infeasible_model():
...
+    es.add(
+        solph.components.Source(
+            outputs={bel: solph.flows.Flow(nominal_value=4, variable_costs=5)}
+        )
+    )
+    m = solph.Model(es)
+    try:
+        m.solve(solver="cbc", allow_nonoptimal=False)
+    except Exception as e:
+        assert "The solver did not return an optimal solution" in str(e)
```
p-snft, line 62: The test case is almost the same as before, so this should do:

```suggestion
    with pytest.raises(RuntimeError, match="The solver did not return an optimal solution"):
        m.solve(solver="cbc", allow_nonoptimal=False)
```

p-snft (approved):
I think it makes sense to update the style as we touch the tests anyway.

PR #1143, opened 2024-11-25
Draft: Calculate Mipgap in the meta results

leroum:
in the processing.py the mipgap is calculated using the Upper and Lower bound returned by pyomo and written in the meta results dictionary.
A test  tests this functionality.


p-snft:
Replaced by #1271.

PR #1145, opened 2024-11-29
Rename custom_attributes to custom_properties for Nodes

p-snft:
Setting the `oemof.network.Node.custom_properties` using an argument called `custom_attributes` was rather confusing. Additionally, the class Bus already called the argument `custom_properties`.


p-snft:
I asked for feedback two weeks ago. As nobody objected, I'll merge.

PR #1155, opened 2025-02-09
Enabling Examples To Pass on Installation

RobSisson:
- Edited GenericStore instances to use nominal_storage_capacity
- Edited Flow instances to use nominal_value
- Removed duplicated "investment_nonconvex": False, in test_nodes_with_none_exclusion_old_name and test_nodes_with_none_exclusion in test_processing.py
- Added openpyxl to optional dependencies (required for dispatch.py)
- Updated the readme to include more comprehensive instructions

Suggestions and contributions were also made by https://github.com/RainerGaier


p-snft:
Thanks again for your effort. I feel very sorry that we did not provide enough guidance and probably did duplicate work. Rainer already worked through the examples and I think we got them all fixed. If you still found issues, it might be because of problems in the presentation.

In fact, the naming `nominal_capacity` is for solph v0.6+, `nominal_value` for older versions. So, here you reverted a lot of changes that we already did for the upcoming release. For examples that target the current (at the moment v0.5) state of solph, you can look at https://oemof-solph.readthedocs.io/en/stable/examples/index.html or https://github.com/oemof/oemof-solph/tree/master (instead of https://github.com/oemof/oemof-solph/tree/dev).

RobSisson:
Ah I see! So my issue was that I was using examples which were for a newer version of OEMOF compared to the version of OEMOF I had installed?

p-snft:
Yes. The actual issue is that the example still tells (in the docstring) that it is for solph 0.5.x but it is for 0.6.x.

Now, I rely on your feedback: Is it really caused by the wrong docstring or didn't you read it anyway? If so, should we put something like `assure_solph_version("0.6")` into the executable code that will fail with a proper error message if run?


Just that I don't forget. A simple implementation would work like this:
```python
def assure_solph_version(version_string: str) -> None:
"""
Tests given version string against the version of the presently installed version of solph.
"""

    solph_version = solph.__version__[:3]
    if solph_version != version_string:
        raise (RuntimeError(f"Unsupported version of solph: {solph_version}"))

assure_solph_version("0.6")
```

p-snft:
It's cleaner to continue at a separate issue. I will close this.

PR #1158, opened 2025-02-24
Revision/simplify indexes

p-snft:
Simplify indexes in the TSAM compatible version (esp. removing the current multi-period investment).


p-snft:
Diverged too much to be easily merged. Will cherry-pick changes.

PR #1168, opened 2025-02-28
Docs/grid icons

leroum:
Grid icons are now matching to the articles. Alls light mode icons are there now. But the ones for the ldark mode which need oemof.visio are still not there (waiting for some ajustments in visio)


docs/_files/code/Visio_beispiel.py
```diff
@@ -0,0 +1,274 @@
...
+
+    logging.info("Optimise the energy system")
+
+    # initialise the operational model
+    energysystem_model = Model(energysystem)
+
+    esgr = ESGraphRenderer(
+        energysystem_model,
+        legend=False,
+        filepath="D:/ES.svg",
```
Bachibouzouk, line 257:
```suggestion
        filepath="ES.svg",
```

The example with `D:/` will only work for windows users and for windows users having a drive under this address.

docs/_files/code/plots_grids.py
```diff
@@ -0,0 +1,285 @@
+import matplotlib.colors as mcolors
```
Bachibouzouk, line 1: A docstring to this module would be helpful. As far as I understood it is generating figures of the doc, right?

Bachibouzouk (approved):
The comments in the two python files are in german, it is intentional? 

Apart from this I would say it is good to merge

PR #1170, opened 2025-02-28
Produce images of examples energysystems and check that they run

Bachibouzouk:
This is useful to check that the examples are all working, there are currently 2 which fails locally for me

```
dispatch.py failed
simple_dispatch.py failed
```
When the script is run with `test_optimize=False`, the optimization will not be run, the system's network will instead be saved to the documentation as an image.

### DIFF
--- a/examples/activity_costs/activity_costs.py
+++ b/examples/activity_costs/activity_costs.py
@@ -47,7 +47,7 @@ import pandas as pd
 from oemof import solph
 
 
-def main():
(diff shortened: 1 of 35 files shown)


p-snft:
I just had a deeper look and some examples are not so easy to split because `Node`s need to be defined for result processing. Thus, I agree that having this now is more valuable than having better examples. (They need some rework anyway.)

Issue #1171, opened 2025-03-07
Multi-Objective optimization


MaxHiDLR:
I believe that oemof lacks a constrained multi-objective optimization approach for operation and investment. While there are workarounds, they become limited, especially when applying multi-period optimization. I see several possible approaches:

1. **Separate optimization for each objective** – Operation and design are optimized individually for each objective. This can be easily implemented by performing two separate optimizations.

2. **Optimization based on one objective with constraints on others** – Operation and design are optimized for a primary objective while being constrained by additional objectives. These constraints can vary over multiple periods and can be set individually for each period.

Approach 2 better reflects real-world scenarios. Typically, investments and operations are monetary cost-driven but subject to limitations such as CO₂ emissions, land use, or resource availability. These constraints can even become stricter over time by applying a multi-period investment strategy—similar to the concept of a "transformation path" of countries.

My proposed approach:

- Add a `restricted_variable_costs` attribute to flows, defined as a list of numeric values (either iterable or scalar).
- Modify the `investment class` to consider additional `restricted_investment_costs`.
- Adapt and extend `constraints.additional_investment_flow_limit` to generate constraints for each additional objective.

At this stage, this would primarily enable multi-objective-constrained optimization, laying the foundation for multi-objective optimization. The actual multi-objective procedure can then be handled by the user and explained in an example by the following steps:

1. Run the optimization without constraints and extract all objective values.
2. Set constraints for the additional objectives.
3. Gradually tighten these constraints.

What do you think about the idea? Am I missing something, and is it already possible? Do you use something similiar and already have an workaround?

If anyone is interersted in collaboration, pls contact me.

p-snft:
Technically, both methods are just different ways to sample the same Pareto front. This is commented on, e.g. starting from page 15 in [doi:10.1016/j.adapen.2024.100190](https://www.sciencedirect.com/science/article/pii/S2666792424000283). Also, all options are possible already, with different levels of manual work needed.

* Weighted sum method: Straightforward to implement. Just need to implement a weighted cost function to calculate costs (variable_costs and investment).
* 𝜀-constraint method: Straightforward for Flows (operation), possibly also doable with `shared_limit` (much to do by hand) or other constraint.
* MGA: Has been done, e.g. here: https://github.com/fgres/oemof-mga


I think, there are two things to do:
1. Have a proper example to make that clear how weighted sum works.
2. Improve handling of 𝜀-constraint method. (I do not think, we need new keywords for that.)

[...]

PR #1174, opened 2025-03-12
Feature/custom attributes for investments

MaxHiDLR:
Constraints Investments for costume attributes missed the possibility of constraining not only the investment flow, but the sum of investment and operation. Furthermore no OffSet for costume Attributes in investments were considers.

This PR adds both and explains it in an example. 

It is handy to have these to run a restricted optimization, as explained in [Issue 1171](https://github.com/oemof/oemof-solph/issues/1171)

### DIFF
--- a/examples/generic_invest_limit/example_generic_invest.py
+++ b/examples/generic_invest_limit/example_generic_invest.py
@@ -74,7 +74,7 @@ from oemof import solph
 
 def main(optimize=True):
 
-    data = [10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10]
+    data = [2, 2, 12, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10]
     # create an energy system
     idx = solph.create_time_index(2020, number=len(data))
     es = solph.EnergySystem(timeindex=idx, infer_last_interval=False)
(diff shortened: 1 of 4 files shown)


p-snft (commented):
At the moment, I would see this as a [Showcase](https://oemof.org/category/showcase/) entry. Sure, the constraint is written in a generic way, but an example works without tests and discussions about the API.

(@gnn and I were discussing how to harmonise different types of variables, so that a "flow of money" would be constrained using the same code as energy flows. But maybe, this would rather be something for solph 2.)

Maxhi77:
Thank you for the feedback.

To specify the offset or cost per unit of a custom attribute, I used the following:

`custom_attributes={“space”: {“cost”: 1, “offset”: 5}},`

The old version used instead:

`custom_attributes={"space": 1}`

This is relevant at the moment the constraint is set. If I put a wrapper around the old constraint, it only works if

1.  I check the structure of the custom attributes and look for cost or not. I don't like this idea. 
2.  It would be easier to simply change the structure of the custom_attributes to the new one. 
3. Keep the old restriction and write a note that the new structure must be used in the future.

p-snft:
> This is relevant at the moment the constraint is set. If I put a wrapper around the old constraint, it only works if ...

I see the point. As this is not about about arguments to the function only, a classical wrapper will not do. Maybe, just write a deprecation warning for the old constraint, telling that the new one should be used in the future.

PR #1177, opened 2025-03-13
Bring back the _storage_balance_first_rule.

maltefritz:
**General description**

This pull request aims to fix the issue #1176.

To fix the error @jfreissmann and I bring back the `_storage_balance_first_rule`. In addition to this method, we also replace the `storage_content` at the first time step with the `init_content` variable in the `_inv_storage_init_content_fix_rule`.

<details><summary>GenericInvestmentStorageBlock</summary>

```python

def _inv_storage_init_content_fix_rule(block, n):
    """Constraint for a fixed initial storage capacity."""
    return block.init_content[n] == n.initial_storage_level * (
        n.investment.existing + block.invest[n, 0]
    )

...

def _storage_balance_first_rule(block, n):
    """
    Rule definition for the storage balance of every storage n
    for the first time step.
    """
    expr = 0
    expr += block.storage_content[n, 0]
    expr += (
        -block.init_content[n]
        * (1 - n.loss_rate[0]) ** m.timeincrement[0]
    )
    expr += (
        n.fixed_losses_relative[0]
        * (n.investment.existing + self.invest[n, 0])
        * m.timeincrement[0]
    )
    expr += n.fixed_losses_absolute[0] * m.timeincrement[0]
    expr += (
        -m.flow[i[n], n, 0] * n.inflow_conversion_factor[0]
    ) * m.timeincrement[0]
    expr += (
        m.flow[n, o[n], 0] / n.outflow_conversion_factor[0]
    ) * m.timeincrement[0]
    return expr == 0

self.balance_first = Constraint(
    self.INVESTSTORAGES, rule=_storage_balance_first_rule
)
```

</details>

### DIFF
--- a/src/oemof/solph/components/_generic_storage.py
+++ b/src/oemof/solph/components/_generic_storage.py
@@ -1676,15 +1676,42 @@ class GenericInvestmentStorageBlock(ScalarBlock):
 
             def _inv_storage_init_content_fix_rule(block, n):
                 """Constraint for a fixed initial storage capacity."""
-                return block.storage_content[
-                    n, 0
-                ] == n.initial_storage_level * (
+                return block.init_content[n] == n.initial_storage_level * (
                     n.investment.existing + block.invest[n, 0]
                 )
 
             self.init_content_fix = Constraint(
                 self.INVESTSTORAGES_INIT_CONTENT,
                 rule=_inv_storage_init_content_fix_rule,
+            )
+
+            def _storage_balance_first_rule(block, n):
+                """
+                Rule definition for the storage balance of every storage n
+                for the first time step.
+                """
+                expr = 0
+                expr += block.storage_content[n, 0]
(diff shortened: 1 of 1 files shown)


fwitte:
The tests are not yet working here, there is not `storage_content` attribute apparently.

jfreissmann:
Seems to be a problem when using `self.TSAM_MODE`. We'll need to take a look how this should be integrated for that, if at all. Maybe @p-snft has an idea?

p-snft (changes requested):
I think, this is not in line with the way the time steps are defined now. See [#1180](https://github.com/oemof/oemof-solph/pull/1180) for the way I think it should be.

p-snft:
Closed in favor of #1180.

(See https://github.com/oemof/oemof-solph/issues/1176#issuecomment-2745292107)

PR #1181, opened 2025-03-18
Add a new `Results` class

p-snft:
Instances of the class allow access to a model's results by converting the model's variables to pandas `DataFrame`s. Minimal design requirements are:

- allow access by variable name via attribute reference or subscription,
- convert to a `DataFrame` lazily, i.e. only convert on first access and
- convert only once.

The current implementation fulfils these requirements.


src/oemof/solph/_models.py
```diff
@@ -441,3 +444,46 @@ def relax_problem(self):
...
+    #   attributes of `model.solver_results` in order to make `Results`
+    #   instances returnable by `model.solve` and still be backwards
+    #   compatible.
+    def __init__(self, model: Model):
+        # TODO: Disambiguate colliding variable names.
+        self.variables = {
+            str(variable).split(".")[-1]: variable
+            for vardata in model.component_data_objects(Var)
+            for variable in [vardata.parent_component()]
+        }
```
p-snft, line 461: I think we could replace the dictionary comprehension by creating an empty dictionary and adding items afterwards. This way, it was possible to find name collisions.

```suggestion
        self.variables = {}
        for vardata in model.component_data_objects(Var):
            for variable in [vardata.parent_component()]:
                key = str(variable).split(".")[-1]
                if key not in self.variables:
                    self.variables[key] = variable
                else:
                    raise ValueError(
                        f"Variable name defined multiple times: {key}"
                        + f"(last time in '{variable}')"
                    )
```

p-snft:
Should we flag this experimental and add it to dev already before it's fully done?

src/oemof/solph/_models.py
```diff
@@ -441,3 +444,46 @@ def relax_problem(self):
         relaxer._apply_to(self)
 
         return self
+
+
+class Results:
```
p-snft, line 449: As this in not depending on `solph.Model`, I see this class going to processing.

src/oemof/solph/_models.py
Bachibouzouk: I am not a big fan of list/dict comprehension for more than one iteration loop (unless the operation is super obvious and simple), I like the for loop :). On the plus side it makes debugging  easier

Bachibouzouk: What is the usecase where the variable names are defined multiple times ?

p-snft: You can have the same variable name defined in multiple blocks. At the dev meeting, we decided to flatten this (e.g. "flow" instead of "SimpleFlowBlock.flow"). However, nobody prevents you from defining "MyOwnBlock.flow" as well, which then collides with the other variable.

p-snft: The change also revealed another problem: One variable was added thousands of times. (The current version avoids this, but still iterates over the same variable as often as before.)

src/oemof/solph/_results.py
```diff
@@ -0,0 +1,97 @@
...
+SPDX-FileCopyrightText: Patrik Schönfeldt <patrik.schoenfeldt@dlr.de>
+
+SPDX-License-Identifier: MIT
+
+"""
+
+import warnings
+from functools import cache
+
+import numpy as np
```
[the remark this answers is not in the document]
Bachibouzouk: Don't linter already exist, which can catch unused inputs? They probably consume less energy than the copilot no?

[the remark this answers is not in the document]
p-snft: Microsoft just pushes Copilot on everybody. We didn't ask for it.

p-snft:
I'll merge the current state to have an alpha release including it.

PR #1182, opened 2025-04-01
Rename storage content variables

p-snft:
I stumbled over intra/inter so often that I want to have clearly distinguishable names.

### DIFF
--- a/src/oemof/solph/components/_generic_storage.py
+++ b/src/oemof/solph/components/_generic_storage.py
@@ -349,10 +349,10 @@ class GenericStorageBlock(ScalarBlock):
         The variable of storage s and timestep t can be accessed by:
         `om.GenericStorageBlock.storage_content[s, t]`
 
-    storage_content_intra
+    storage_content_timepoint
         Storage content for every storage and timestep of typical periods
         (only used in TSAM-mode). The variable of storage s and timestep t can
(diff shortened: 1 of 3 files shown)


MaxHiDLR:
I see the similarity in both words. For my part I was fine with the names, since they describe them well according to the paper.

Alternative name suggestions with some theoretical background:
Intra-storage content is more like a storage delta that occurs during a typical period and this delta is the same for all typical periods. It can therefore be negative. The inter storage content is the storage content between two typical periods.

intra_storage_content -- intra_storage_delta -- low_layer_storage_delta
inter_storage_content -- intermediate_storage_content -- high_layer_storage_content

henhuy:
and they have the same length...
I like the "delta" suggestion, but would also vote for keeping intra and inter in variable name to keep them aligned with the paper. Maybe it would be sufficient to use `intra_storage_delta` and `inter_storage_content`?

p-snft:
Thanks for the suggestions. It is really just that "intra" and "inter" sound very similar. I remember that we had to be very careful with our pronunciation at the last dev meetings.

 I really like `_delta` because it also indicates the variable can be negative. It's not obvious unless you see that `intra_storage_delta = 0` at the beginning of the periods. So, it needs to allow negative values to discharge.

Issue #1179, opened 2025-03-14
InvestmentFlowBlock does not add ep_cost & offset to objective if Investment object is NonConvex


jfreissmann:
**Short description:**
While probing the `Investment` object for the Issue #1172 and PR #1173, it seems like another issue came to the attention of @maltefritz and me: When e.g. a `Converter` has a `solph.Investment` as the `nominal_capacity` with `ep_cost`, `offset` and `nonconvex` set, while the `Flow` itself is **not** set to be NonConvex (e.g. when the `min` parameter should be enforced), the `ep_cost` and `offset` are not added to the objective function. See below for an example of a `Converter` producing the error:

```python
hp_expensive = solph.components.Converter(
    'heat pump expensive',
    inputs={b_el: solph.Flow()},
    outputs={
        b_th: solph.Flow(
            nominal_capacity=solph.Investment(
                ep_costs=100,
                offset=2000,
                maximum=1,
                minimum=0,
                nonconvex=solph.NonConvex()
            ),
            variable_costs=1
        )
    },
    conversion_factors={b_el: [3]}
)
```

**To Reproduce:**
You can find a full example in the gist below:

https://gist.github.com/jfreissmann/a7165f9af517a1a8518a70e809a20186

The objective function produces from the energy system model above looks like this:

```
min
objective:
+50 InvestmentFlowBlock_invest(heat_pump_cheap_heat_bus_0)
+1 flow(heat_pump_expensive_heat_bus_0)
+1 flow(heat_pump_cheap_heat_bus_0)
```

It does not include the `ep_cost` or the `offset` of the `hp_expensive` Converter. Therefore, the solution of the optimization is to build the `hp_expensive` Converter for free and only pay its `variable_cost` to cover the heat demand. This also can be observed via the `'sequences'` of the `'heat bus'` node. The `'scalars'` do not even exist.

**Setup:**
 - OS: Windows 11
 - Python 3.11
 - oemof.solph dev branch latest

jfreissmann:
The issue seems to stem from the fact, that I set the `nonconvex` parameter of the `solph.Investment` object to `solph.NonConvex`. The `InvestmentFlowBlock` class seems to expect the parameter to be either `True` or `False`, as can be seen in its `Set` definitions:

```python
self.CONVEX_INVESTFLOWS = Set(
    initialize=[
        (g[0], g[1])
        for g in group
        if g[2].investment.nonconvex is False
    ]
)

[...]

PR #1173, opened 2025-03-11
Fix bug where InvestNonConvexFlowBlock does not eliminate Investment offset when no investment is made

maltefritz:
**General description**

This pull request aims to fix the issue #1172 

To fix the bug @jfreissmann and I expand the `_create_sets` method in the `_invest_non_convex_flow_block`. We decitde to name the sets linear and offset not to repeat the convetion of convex and nonconvex. But you can feel free to change those.

<details><summary>_create_sets method</summary>

```python
self.LINEAR_INVEST_NON_CONVEX_FLOWS = Set(
    initialize=[
        (g[0], g[1])
        for g in group
        if g[2].investment.nonconvex is None
    ]
)

self.OFFSET_INVEST_NON_CONVEX_FLOWS = Set(
    initialize=[
        (g[0], g[1])
        for g in group
        if isinstance(g[2].investment.nonconvex, NonConvex)
    ]
)
```

</details>

After that we expand the `_investvar_bound_rule` to differentiate between the linear and offset flows. In addition, we create the `invest_status` variable in alingment with the `GenericInvestmentStorage` class:

<details><summary>_create_variables method</summary>

```python
def _create_variables(self):
        r"""
        Status variable (binary) `om.InvestNonConvexFlowBlock.status`:
            Variable indicating if flow is >= 0 indexed by FLOWS

        :math::`P_{invest}` `InvestNonConvexFlowBlock.invest`
            Value of the investment variable, i.e. equivalent to the nominal
            value of the flows after optimization.

        :math::
            `Y_{invest_status}(i,o,p)` `InvestNonConvexFlowBlock.invest_status`
            Binary variable representing whether or not an investment is made. 

        :math::`status\_nominal(i,o,t)` (non-negative real number)
            New paramater representing the multiplication of `P_{invest}`
            (from the <class 'oemof.solph.flows.InvestmentFlow'>) and
            `status(i,o,t)` (from the
            <class 'oemof.solph.flows.NonConvexFlow'>)
            used for the constraints on the minimum and maximum
            flow constraints.

        .. automethod:: _variables_for_non_convex_flows
        """
   
        ...

        def _investvar_bound_rule(block, i, o, p):
            """Rule definition for bounds of the invest variable."""
            if (i, o) in self.LINEAR_INVEST_NON_CONVEX_FLOWS:
                return (
                    m.flows[i, o].investment.minimum[p],
                    m.flows[i, o].investment.maximum[p],
                )
[...]

PR #1184, opened 2025-04-02
Fix InvestNonConvexFlowBlock set definition by nonconvex parameter

jfreissmann:
The current set definitions of the `InvestNonConvexFlowBlock`, as implemented by our PR #1173, check wether the argument passed to the `nonconvex` parameter is an instance of the `solph._options.NonConvex` class. This is not the intended usage of the `nonconvex` parameter of the `solph._options.Investment` class. This PR aligns the set definition to check for boolean values `True` or `False`. A test added in the PR mentioned above also contains the erroneous usage, which is fixed as well.

Additionally, @maltefritz and I added a type check of the `nonconvex` argument, as its misaligned usage in contrast to the Flow's `nonconvex` parameter can lead to wrong usage staying undetected. A test for the check is added as well.

This should also help resolve Issue #1179.

### DIFF
--- a/src/oemof/solph/_options.py
+++ b/src/oemof/solph/_options.py
@@ -173,7 +173,7 @@ class Investment:
         if not isinstance(self.nonconvex, bool):
             e5 = (
                 "The `nonconvex` parameter of the `Investment` class has to be"
-                + f"of type boolean, not {type(self.nonconvex)}."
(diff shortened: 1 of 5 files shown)


jfreissmann:
The newly added type check for the `nonconvex` parameter seems to be worthwhile, as some other false usages of it in solphs test suite emerged as well 😄

PR #1186, opened 2025-04-22
Consider storage_costs for storage with capacity investment in v0.5

### LINKED ISSUE #1185: storage_costs are not considered for GenericInvestmentStorageBlock
The parameter "storage_costs" seems to have no effect on the objective of a GenericInvestmentStorageBlock, while a GenericStorageBlock considers it. 

**Expected behavior:** 
A GenericInvestmentStorageBlock and a  GenericStorageBlock with the same nominal storage capacity value give the same objective value. 

**Minimum (Non) Working Example:**
In the following example, comment/uncomment the nominal storage capacity lines (21ff.) to switch between Investment and fixed capacity storage. storage_costs is the only nonzero cost component. The objective after optimization is 0 for the investment case and 60 for the fixed one. 

```
import numpy as np
import pandas as pd
from matplotlib import pyplot as plt

from oemof import solph


def main():
    idx = pd.date_range("1/1/2023", periods=12, freq="h")
    es = solph.EnergySystem(timeindex=idx, infer_last_interval=False)

    bel = solph.Bus(
        label="bel"
    )
    es.add(bel)

    battery = solph.components.GenericStorage(
        label="battery",
        # nominal_storage_capacity=10,
        nominal_storage_capacity=solph.Investment(
            ep_costs=0,
            minimum=10,
            maximum=10,
        ),
        inputs={
            bel: solph.Flow(
                nominal_value=1,
            )
        },
        outputs={
            bel: solph.Flow(
                nominal_value=1,
            )
        },
        storage_costs=12 * [1],
        initial_storage_level=0.5,
    )
    es.add(battery)

    model = solph.Model(es)

    model.solve(solver="gurobi", solve_kwargs={"tee": True})

    print("Objective value: ", model.objective())


if __name__ == "__main__":
    main()
```

 -   OS: macOS 15.3
 -   Python version 3.11.10
 -   solph version: 0.6.0a3

p-snft:
Fixes #1185 (but with inconsistent definition of storage content indexing).

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -15,7 +15,7 @@ jobs:
 
     steps:
       - name: Check out Git repository
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
 
       - name: Set up Python
         uses: actions/setup-python@v5
--- a/.github/workflows/packaging.yml
+++ b/.github/workflows/packaging.yml
@@ -22,7 +22,7 @@ jobs:
         python-version: ["3.10"]
 
     steps:
-    - uses: actions/checkout@v1
+    - uses: actions/checkout@v3
     - name: Set up Python ${{ matrix.python-version }}
       uses: actions/setup-python@v5
       with:
--- a/.github/workflows/tox_checks.yml
+++ b/.github/workflows/tox_checks.yml
@@ -31,7 +31,7 @@ jobs:
       - name: Install LaTeX
         run: sudo apt install dvipng rubber texlive-latex-extra
       - name: Git clone
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
 
       - name: Set up Python ${{ env.default_python || '3.10' }}
         uses: actions/setup-python@v5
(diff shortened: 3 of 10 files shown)


p-snft:
The Problem mentioned in https://github.com/oemof/oemof-solph/issues/1185#issuecomment-2822158359,

> Note that this differs from your statement (60, value for v0.5). The fix for v0.5 ([#1186](https://github.com/oemof/oemof-solph/pull/1186)) revealed that the storage content for the last time point is not fixed.

originates from the constraint imposing the storage content limited is using `TIMEINDEX` instead of `TIMEPOINTS`:
https://github.com/oemof/oemof-solph/blob/8835a9af7faf9a14c5ad8ef344177bfbb87d6e04/src/oemof/solph/components/_generic_storage.py#L1668-L1672

The `TIMEINDEX` is using `TIMESTEPS` but is pretty much hard-wired to the v0.5 investment mode. Thus, I fear it's not worth fixing this problem in the stable version.

Issue #1169, opened 2025-02-28
Setting underlying structure to implement Facade


Bachibouzouk:
Follow up from the discussion which happened at [2025-02 oemof-meeting](https://github.com/oemof/oemof/wiki/Meeting-2025.02)

Variante A

![Image](https://github.com/user-attachments/assets/baf7a549-e259-471b-aeb1-6cf92cd12873)

`inputs` and `outputs` attributes are stripped from the class `Node`, `Node` as known in v0.5  is renamed after `AtomicNode` and `SubNetwork` is added 

This would require 2 (easy) releases in oemof.network and 1 in oemof.solph

Pros
- 

Cons
- might be confusing that what used to be Node is now AtomicNode and Node is still there but just higher in the hierarchy
- 2 releases in oemof.network which might 
- might break older version of oemof as network version is not upperbound

Variante B

![Image](https://github.com/user-attachments/assets/6e247a8a-1ff0-4b2a-b791-55619fa6b2ff)

Pros
- less releases to make to perform the change

Cons
- Using "Abstract" in the naming

p-snft:
I also see "Option C", to have `AbstractNode` and `AtomicNode` with explicit name prefixes.

Bachibouzouk:
We are tending towards option B because it breaks less things from the existing structure

p-snft:
Solved in a more generic way in https://github.com/oemof/oemof-network/pull/60.

PR #1193, opened 2025-05-06
 Add facade in oemof.solph

Bachibouzouk:
*   What functionality does it implement?
  This is related to the discussion in #1169, after a discussion with @uvchik we decided it might be worth testing approach B. We will link the related PR in `oemof.network`
  *   Where is it located?
  A new `_facade.py` file in `oemof/solph/`
  *   How does the API look?
  This is tentatively described in the `examples/facade/facade.py` commited file


Bachibouzouk:
At the moment the code does not work, this is why I set the PR to draft

src/oemof/solph/_facades.py
```diff
@@ -0,0 +1,33 @@
...
+
+class Facade(SubNetwork):
+    # attributes
+    def __init__(self,*args,**kwargs):
+        self.facade_type = None
+
+    def add_subnode(self,*args):
+        """Add a subnode to the subnetwork"""
+        for sub_component in args:
+            sub_component.label = self._sub_component_labelling(sub_component.label)
```
Bachibouzouk, line 24: Apparently cannot set label once it is set for renaming components in subnetwork
 - let labels be overwritten in oemof-solph
 - workaround with custom_properties "sub_label"

p-snft: Labels are meant to be unique and constant identifiers at the level of `oemof.network`. Why do the labels have to be changed?

Bachibouzouk: Because we wanted to label the components within a subnetwork in a specific way, but we found a solution, we do not change the labels, instead we use a method to label the subnetwork components right upon creation

examples/facade/facade.py
```diff
@@ -0,0 +1,180 @@
...
+        self.el_bus = el_bus
+
+    def build_subnetwork(self):
+        internal_bus = Bus(custom_properties={"sub_label": "internal_bus"})
+        self.add_subnode(internal_bus)
+
+        feedin = Converter(
+            inputs={self.el_bus: Flow(variable_costs=self.feedin_tariff)},
+            outputs={internal_bus: Flow()},
+            label=self.sub_component_labelling("feedin_converter"),
```
Bachibouzouk, line 57: @p-snft - here we label the subnetwork node using a Facade method. We think we need to have this naming convention to be able to bookkeep the nodes for postprocessing

Bachibouzouk: Our aim here is to manage to create a Facade class by tempering minimally with the existing code. This creation process will help the question raise and will prompt discussion about more elegant implementations.

p-snft: This sounds similar to the way we did it in MTRESS:

```python
class AbstractComponent:  # pretty much the MTRESS subnetwork

    (...)

    def create_solph_node(self, label: str, node_type: Callable, **kwargs):
        """Create a solph node and add it to the solph model."""
        _full_label = tuple(self.create_label(label))

        if label in self._solph_nodes:
            raise KeyError(
                f"Solph component named {_full_label} already exists"
            )

        _node = node_type(label=_full_label, **kwargs)

    def create_label(self, label: str) -> list[str]:
        """Return a unique label based on the identifier."""
        return self.identifier + [label]

    @property
    def identifier(self) -> list[str]:
        """Return identifier."""
        identifier = []
        highest_level = self
        while highest_level is not None:
            identifier.append(highest_level.name)
            highest_level = highest_level._nesting_element

        identifier.reverse()
        return identifier
```

(In reality split into [AbstractComponent](https://github.com/mtress/mtress/blob/f9afc0a77a448f2c90b293e66274ef107fb8e505/src/mtress/_abstract_component.py#L46-L55) and [NamedElement](https://github.com/mtress/mtress/blob/f9afc0a77a448f2c90b293e66274ef107fb8e505/src/mtress/_interfaces.py#L19).)

Bachibouzouk:
With the HierachicalLabel and its depth attribute, it is easy to plot nested SubNetworks and by changing in its `__str__()` method one can change how the SubNetwork labels should be rendered

With . between the levels: 

nested_facade_with_dot

Seeing the whole tuple:

nested_facade

p-snft:
I really liked casting things to `HierachicalLabel`s in the background, as it makes code easier to read. Can we revert https://github.com/oemof/oemof-solph/pull/1193/commits/4e0029879f164c90ab62fe6d13e76393c0246ede?

Bachibouzouk:
> I really liked casting things to `HierachicalLabel`s in the background, as it makes code easier to read. Can we revert [4e00298](https://github.com/oemof/oemof-solph/commit/4e0029879f164c90ab62fe6d13e76393c0246ede)?

For the moment oemof-visio is not able to display properly if the nodes  (which are `Node` instance) do not use `HierachicalLabel`. This is the reason why I needed to write them explicitly as the basic component inherit from `Node` and not `AtomicNode`

PR #1195, opened 2025-05-27
Feature/result dict wrapper

p-snft:
This adds a drop-in replacement for `processing.results` that uses the new `solph.Results` class. It means to provide backwards compatibility while radically simplifying the processing code.


Bachibouzouk:
> `NonConvexFlowBlock` and `InvestNonConvexFlowBlock`) define the same variable name

Do you mean `status` and `status_nominal` ?

p-snft:
I just implemented the approach in #1202.

p-snft:
What is now missing is just the support for experimental features (TSAM and multi-periods).

PR #1197, opened 2025-06-04
Fix a couple of issues in the updated documentation

fwitte:
- Missing link for the OffsetConverter in the components overview added
- Reordered the components section to be more intuitive (from general/basic components to more specialized ones)
- Shorten line length in the changed sections and fix lists indendations


fwitte:
The errors in the docs are broken hard-coded links (pointing to the documentation itself) instead of using references. However, they seem to be outdated anyway, not sure what to do with them.

- One is pointing to the groupings model, which is referenced from the results page (we deleted the groupings page, we may just eliminate that paragraph in the results page). Can you decide what to do here @p-snft, @uvchik?
- Two are from very early releases linking component docs, I will fix that.

fwitte:
> In my opinion, we can stop mentioning groupings in results.rst. Maybe, these are something for an advanced tutorial?

Completed by [77fd1df](https://github.com/oemof/oemof-solph/pull/1197/commits/77fd1dfddfeb1319a1daf6efc45d5870bcf63f70)

fwitte:
> Two are from very early releases linking component docs, I will fix that.

Completed by [ae125fd](https://github.com/oemof/oemof-solph/pull/1197/commits/ae125fdf1f66d0d3e4439ba844a550523c48422a)

Issue #1199, opened 2025-06-04
Deprecate parameter "time_intervals"


p-snft:
Mark `EnergySystem` initialisation giving time_intervals as deprecated. The new `Results` object expects a time index and there is actually no real benefit in not giving one.

To make this work, also the default of `infer_last_interval` is set to to False. (Inferring requires a frequency, which is not present for numeric indexes.)

PR #1200, opened 2025-06-04
Feature/allow numeric timeindex

p-snft:
Implements #1199 in the way that the following lines will now produce the same result:

```Python
EnergySystem(time_incement=[1,1,1,2,3,1])  # deprecated
EnergySystem(timeindex=[0,1,2,3,5,8,9])  # new way
EnergySystem(timeindex=[12,13,14,15,17,20,21])  # also possible (initial value is arbitrary)
```

Of course, using pandas.DatetimeIndex still works.

Todo: Add to whatsnew.

### DIFF
--- a/src/oemof/solph/_energy_system.py
+++ b/src/oemof/solph/_energy_system.py
@@ -102,16 +102,6 @@ class EnergySystem(es.EnergySystem):
             groupings = []
         groupings = GROUPINGS + groupings
 
-        if not (
-            isinstance(timeindex, pd.DatetimeIndex)
-            or isinstance(timeindex, type(None))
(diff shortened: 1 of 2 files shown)


p-snft:
Everything that is failing is related to the outdated absolute links in the docs (see #1197).

uvchik:
Okay, too late 😄 

Thank you, it makes the code more simple and one can easily transfer the old increments to the new index.

```python
>>> import numpy as np
>>> increment = [1, 1, 1, 2, 3, 1]
>>> np.cumsum([0] + increment)
[0 1 2 3 5 8 9]
```

PR #1202, opened 2025-06-05
Fix/result object same name vars

p-snft:
This changes the results object to be able to handle variables with the same name (present in different Blocks) as long as the indexes are disjunct.

Analysis of the situation:
* We have the convention that one Entity (Node or Flow) is only covered by one Block.
* The status of a Flow is defined either in NonConvexFlowBlock or in InvestNonConvexFlowBlock. As the function   _create(self, group=None) is not called by the Constructor, the variable exists only once per Flow.


p-snft:
As the result object is not used yet, this change reduces coverage. I'd like to accept this looking at #1195, which replaces the old processing with one based on the Result object fixed here.

PR #1204, opened 2025-06-12
Remove unmaintained and non-working code

p-snft:
This PR removes large chunks of code that is hard to maintain (and mostly abandoned unmaintained). As the code is partly covered by (integration-) tests, this will reduce the measured test coverage. However, there are no unit-tests deleted that are checking features that still exist. I removed the tests in particular as they include the (removed) `SinkDSM` component. As it is an integration test checking results of a complete energy system, they are hard to adjust. Also, the new numbers (about 80 %) are more honest than the old ones when we interpret them as unit test coverage.

Removed are:
* Experimental component `SinkDSM` (unmaintained and bad code quality).
* `ElectricalLine` for voltage angle simulation (unmaintained, broken).
* Integration tests for multi-period model (hard to understand, partly disused).


lensum:
I'm surprised that `SinkDSM` is part of this. Though I do not use it myself, I always assumed that people are using it for their research. If I'm not mistaken, there were also some presentations using it at the second to last oemof meeting...
From the `ElectricalLine`, on the other hand, I haven't heard anything in years.

Considering this, would it make sense to clean-up the code of the `SinkDSM` component and write better tests for it? Maybe we can find some volunteers for this at the next oemof meeting?

p-snft:
I asked @jokochems as the original author of `SinkDSM` if he could improve it and he did not find time. He agreed with the removal under the given conditions.

In fact, dropping it does no mean it cannot be re-added. I do not see somebody maintaining it at the moment, in particular as the module is very hard to maintain (code duplication, long file, complex working principle). We had the unmaintained `ElectricalLine` without tests in the repository for some time and it was eventually broken for several releases. Thus I consider removing a component cleaner than having it long-term unmaintained and untested - potentially in a broken state.

p-snft:
@uvchik suggested to directly add a new PR to re-add the `SinkDSM`.

p-snft:
> > @uvchik suggested to directly add a new PR to re-add the `SinkDSM`.
> 
> Then why delete it in the first place?

This way, we make sure we do not have any unverified code in a release. If we just keep it and leave it untested, chances are high that somebody will use the code to produce wrong results. (I believe results are correct at the moment, but as there are no unit tests, I won't guarantee for this after changes.)

lensum:
> > > @uvchik suggested to directly add a new PR to re-add the `SinkDSM`.
> > 
> > 
> > Then why delete it in the first place?
> 
> This way, we make sure we do not have any unverified code in a release. If we just keep it and leave it untested, chances are high that somebody will use the code to produce wrong results. (I believe results are correct at the moment, but as there are no unit tests, I won't guarantee for this after changes.)

That makes sense, thank you for the explanation!

PR #1208, opened 2025-08-19
Revision/cleanup ci

p-snft:
We have a lot of outdated code from legacy CI test pipelines (like Travis). This aims to simplify testing.

Tested python versions are now defined in a single file, depending on where the tests are run:

- CI pipeline: .github/workflows/tox_pytests.yml (runs default tox env)
- Offline tests: tox.ini

Changes are analogue to https://github.com/oemof/oemof-network/pull/57


p-snft:
On the long run, I would remove all (external) services that offer analysis that is not run offline (using tox). But except for badges, this is a GitHub configuration thing and nothing to be changed in the repository. (I already removed the config for Scrutinizer, though.)

In my opinion, our tests should not rely on external services. If you successfully test offline, the online test should yield the same results.

tox.ini
```diff
@@ -138,5 +131,16 @@ deps =
...
+basepython = {env:TOXPYTHON:python3}
+setenv =
+    {[testenv]setenv}
+usedevelop = true
+commands =
+    {posargs:pytest --cov --cov-report=term-missing -vv}
+deps =
+    {[testenv]deps}
+    pytest-cov
+
```
fwitte, line 144: This uses whatever `py3` version the user has installed?

p-snft: Yes.

fwitte: Then double approve! Very nice addition!

PR #1209, opened 2025-09-01
Implement `filter`ing columns of `Results` `DataFrame`s

gnn:
Since the `DataFrame` of a `Results` variable can have a lot of columns, not all of which might be of interest for a user, this PR implements an API to easily filter specific columns in order to only retain those of interest.

NOTE: This PR description is not complete and I will write a few more bullet points with desired and implemented features later.

### DIFF
--- a/src/oemof/solph/_models.py
+++ b/src/oemof/solph/_models.py
@@ -455,6 +455,13 @@ class Filters(dict):
         result.update(filters)
         return result
 
+    @staticmethod
+    def nofilter(*args, **kwargs):
+        return True
+
+    def __getitem__(self, variable: str) -> pd.DataFrame | pd.Series:
+        return self.get(variable, self.nofilter)
+
 
 class Results:
     filters = Filters(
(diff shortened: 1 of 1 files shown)


src/oemof/solph/_models.py
```diff
@@ -446,7 +447,31 @@ def relax_problem(self):
         return self
 
 
+class Filters(dict):
+    def updating(
+        self, filters: dict[str, Callable[[object], bool]]
+    ) -> "Filters":
+        result = Filters(self)
```
Bachibouzouk, line 454: Why is this needed? Rather than `self.update(filters)` ?

Bachibouzouk (commented):
It would be great to have an example of usage

src/oemof/solph/_models.py
gnn: I tried to explained that in the corresponding [commit message](https://github.com/oemof/oemof-solph/commit/9182ba5cda5d9ddd45bc2a6b8118e26a87b76808).  I agree though, that both, `updating` and `Filters` could use a docstring which would contain some of what is written in the mentioned commit message. Would you mind giving it a read and if that doesn't answer your questions I'll have a better understanding on what to change once I'll write those docstrings.

PR #1215, opened 2025-09-17
Add subnetwork functionnality

Bachibouzouk:
Replaces #2147


p-snft (approved):
Released updated oemof.network, works now.

Issue #1013, opened 2023-11-15
GenericStorage not creating InvestmentFlows when used with invest_relation_input_output


lensum:
When using the GenericStorage with an investment optimization approach like this:
```
electric_storage = cmp.GenericStorage(
            label="electric_storage",
            inputs={el_bus: solph.Flow()},
            outputs={el_bus: solph.Flow()},
            loss_rate=0.01,  # based on 25kW for 2.5MW self consumption
            nominal_storage_capacity=solph.Investment(
                ep_costs=elstor_epc, existing=1200, maximum=2100
            ),
            invest_relation_input_capacity=0.9375,
            invest_relation_input_output=1,
            inflow_conversion_factor=0.9,
            outflow_conversion_factor=0.8 / 0.9,  # overall efficiency of 0.8
            initial_storage_level=0,
            balanced=False,
            ),
        )
```
I get `KeyError: 'Index \'("<oemof.solph.components._generic_storage.GenericStorage: \'electric_storage\'>", "<oemof.solph.buses._bus.Bus: \'el_bus\'>", 0)\' is not valid for indexed component \'InvestmentFlowBlock.total\''`

I assume that the error might be in the internal function `_set_flows`. I defined the `self.invest_relation_input_capacity` and `self.invest_relation_input_output`, but the `_set_flows` function only considers `self.invest_relation_output_capacity` for the output flows, meaning that my output flows don't get converted into `InvestmentFlow` objects.
If I use `invest_relation_output_capacity=0.9375` instead of the `invest_relation_input_output`, the model starts to solve.

p-snft:
Shouldn't the `Flow`s be created with `nominal_value=Investment()` to allow for investments in any case?

p-snft:
Maybe, it makes sense to rewrite this thing anyway. I'd consider replacing the fixed relations between the investments by a limit to the c-rate. It would be an internal property that exists independent from the nominal values of the flows.

jokochems:
Yeah, you are right about locating this to the `_set_flows` method, @lensum , I think.
* The case of `invest_relation_input_output` is not accounted for here. In my view, a check needs to be added in `_set_flows`.
* Also, for the other cases, the type of `flow.investment` is evaluated. I think, this needs to be `flow.nominal_value` instead according to the latest structural changes.

p-snft:
[...]

PR #1217, opened 2025-09-22
Bugfix/genericstorage relations

lensum:
##   Describe your pull request as transparent as possible
This PR fixes the bug where the `invest_relation_input_output` parameter of the `GenericStorage` is not adhered to in the resulting models.

## Related issues?
The related issue is: https://github.com/oemof/oemof-solph/issues/1013

## Share your knowledge: Insights/Remarks
The reason for the "misbehavior" was, that the autogeneration for InvestmentFlows did not check the keyword. During the bugfixing, it became increasingly difficult to find a solution that would allow for all the different ways, the GenericStorage is currently used (and the corresponding implicitly expected behavior).
**Therefore, I decided that it is now required to set the `Investment` objects in the `input`, `output` and `nominal_capacity` of the storage explicitly!** 
If `invest_relation_input_capacity`, `invest_relation_output_capacity` or `invest_relation_input_output` are used, checks are performed to make sure that all required `Investment` objects are set and an `AttributeError` is thrown otherwise. Imho, this allows for more fine-grained control of the behaviour of the energy system. However, it means that behavioural patterns of users need to adjust.

### DIFF
--- a/src/oemof/solph/components/_generic_storage.py
+++ b/src/oemof/solph/components/_generic_storage.py
@@ -216,6 +216,7 @@ class GenericStorage(Node):
         # --- END ---
 
         self.nominal_storage_capacity = None
+        self.investment = None
         self._invest_group = False
         self.invest_relation_input_output = sequence(
             invest_relation_input_output
@@ -280,7 +281,7 @@ class GenericStorage(Node):
     def _check_storage_for_investment(self):
         """Checks the storage for an investment object (i.e. if investment
         into the capacity is possible)"""
-        return hasattr(self, "investment")
+        return isinstance(self.investment, Investment)
 
     def _check_invest_relations(self):
         """Checks if the passed invest_relation keywords fit the
@@ -354,7 +355,7 @@ class GenericStorage(Node):
             )
             raise AttributeError(e2)
         if (
-            hasattr(self, "investment")
+            self.investment
             and self.fixed_losses_absolute.max() != 0
             and self.investment.existing == 0
             and self.investment.minimum.min() == 0


lensum:
If anyone could point out to me what kind of problem Black has with the marked sections, I would appreciate that. I find it kind of cryptic.

src/oemof/solph/components/_generic_storage.py
p-snft, line 1: I don't get why this diff is so large. Did you change the line endings? Really hard to review, I will skip this for now and revisit it later.

p-snft (commented):
This is API breaking but it's also a fix. I would weave it through. However, before merge, at least two things are missing:

- [ ] Document changes in changelog.
- [ ] Review of _generic_storage.py. (The diff shows all lines as changed, which means it hard to identify the real changes.)

src/oemof/solph/components/_generic_storage.py
```diff
@@ -56,19 +58,19 @@ class GenericStorage(Node):
         Absolute nominal capacity of the storage, fixed value or
         object describing parameter of investment optimisations.
     invest_relation_input_capacity : numeric (iterable or scalar) or None, :math:`r_{cap,in}`
-        Ratio between the investment variable of the input Flow and the
+        Ratio between the investment variable of the input flow and the
```
p-snft, line 61: I think these had capital letters to refer to the class `Flow`.

p-snft (commented):
Now, that the line endings are fixed, there are just a few open points.

src/oemof/solph/components/_generic_storage.py
p-snft: I also have no clear preference. I just wanted to make sure that it is not changed before we have a common understanding of the previous way. In my option, 2 and 3 are cleaner than 1.

PR #1219, opened 2025-10-27
Revision/rename set nonconvexflow

p-snft:
The new name reflects the fact that `INVEST_NON_CONVEX_FLOWS` is not a subset of the set `NON_CONVEX_FLOWS`, that is now called `FIXED_CAPACITY_NONCONVEX_FLOWS`. Originally (before solph 0.5), this was implicit as investment and binary status could not be combined. Now, the name was misleading.
    
As this only affects expert users, we do not consider this API breaking.


p-snft:
Closed in favour of more complete #1220.

PR #1231, opened 2025-11-19
Harmonise usage of "custom_..." keywords

p-snft:
In oemof.network, every Entity has custom_properties, which is a dictionary containing user-defined information. Until now, this was used for Nodes but not for Flows. Those instead implemented their own way, named custom_attributes. This keyword generates class attributes and left the (also existing) custom_properties attribute empty.

This adds custom_properties to the class Flow, maps custom_attributes to custom_properties, and deprecates custom_attributes.

### DIFF
--- a/docs/whatsnew/v0-6-2.rst
+++ b/docs/whatsnew/v0-6-2.rst
@@ -7,8 +7,8 @@ API changes
 New features
 ############
 
-* Class `Flow` now supports `custom_properties`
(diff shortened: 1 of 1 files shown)


jfreissmann (approved):
The PR definetly makes sense and all changes look fine to me. The code to be deprecated in the future is highlighted and the additional tests make sure the behavior is predictable until >= 0.7. In my opinion, this can be merged.

Issue #843, opened 2022-05-09
nominal_value=1 bounds flow variable without min/max/fix being set


lensum:
**Describe the bug**
If the attribute `nominal_value=1` is set in a flow without setting either one of `min`/`max`/`fix`, the flow variable seems to be bounded to 1.

**To Reproduce**
1. Set up an energysystem model
2. Create a sink with fixed demand > 1, fed from a transformer
3. Set `nominal_value=1` in the transformer outflow
4. Try to solve the resulting model

**Expected behavior**
- A raised warning if `nominal_value` is set without `min`/`max`/`fix` being set
- Documentation of default values for `min`/`max` arguments

**Desktop (please complete the following information):**
 - oemof.solph version: 0.4.5

**Additional context**
Might be related to: https://github.com/oemof/oemof-solph/issues/818

jnnr:
This is because the default value of max is 1. https://github.com/oemof/oemof-solph/blob/7e341fdd041c38da8dfbb95867bb3a5bbc9bbe15/src/oemof/solph/network/flow.py#L182 We could tell about this in the docstring.

p-snft:
We should just state that ’max’ has a default value of 1. When we touch the docstring, we should also delete

https://github.com/oemof/oemof-solph/blob/v0.5/src/oemof/solph/flows/_flow.py#L90-L93

Which is no longer correct.

Issue #1157, opened 2025-02-18
Investment to bidirectional Flow fails


p-snft:
**Describe the bug**

Investment to bidirectional Flow leads to an infeasibility.

**To Reproduce**

See script at https://gist.github.com/p-snft/6106ec8727c78928d2b541afe5ef7501

**Expected behavior**

Capacity should be optimised (to set minimum) if investment problem is set.

**Additional context**

I suggested to do it like this to a colleague of mine without testing if it really works, because I was sure it would do. It does not.

lensum:
In https://github.com/oemof/oemof-solph/pull/1233 @p-snft and I agreed on @p-snft `s following proposal as an initial step on this topic: 

> In my opinion, it makes sense to phase out the argument bidirectional. I would add a FutureWarning telling that bidirectional=True is an alias to set min=-1 and that a negative minimum is an experimental feature. If there is an explicitly set min(t)<0 for any t, there should be an ExperimentalFeatureWarning. (I am undecided if we should warn twice when bidirectional is used.)

This is also implemented in https://github.com/oemof/oemof-solph/pull/1233

PR #1233, opened 2025-11-20
Clarify the default values and behaviour of `min` and `max` parameters in the `flow` class

lensum:
This PR implements changes to the API documentation and `__init__` function of the `flow` class. By this, the default behaviour (and values) of the `min` and `max` parameter/attribute should be clarified to the user.

Fixes: https://github.com/oemof/oemof-solph/issues/843

### DIFF
--- a/docs/whatsnew/v0-6-2.rst
+++ b/docs/whatsnew/v0-6-2.rst
@@ -4,8 +4,8 @@ v0.6.2
 API changes
 ###########
 
-* The `Flow` parameter now raises a `FutureWarning`, as it is to be removed
-  in a following version.
+* The `Flow` parameter `bidirectional` now raises a `FutureWarning`,
+  as it is to be removed in a following version.
 
 New features
 ############


p-snft:
> 1.) Currently, we implicitly set the `min` attribute to `-1` if `bidirectional=True` is passed to a `Flow`-instance. I've currently replicated this behaviour in the PR, but I'm in favor of putting out some kind of warning for the user. As this is related to #1157, it is probably best discussed there (will reply soon).

In my opinion, it makes sense to phase out the argument `bidirectional`. I would add a `FutureWarning` telling that `bidirectional=True` is an alias to set `min=-1` and that a negative minimum is an experimental feature. If there is an explicitly set `min(t)<0`  for any `t`, there should be an `ExperimentalFeatureWarning`. (I am undecided if we should warn twice when `bidirectional` is used.)

```python
if np.array(min).min() < 0:  # Cannot use min without numpy because of we use that as an argument name.
    warnings.warn("Using negative values for Flow.min is experimental.", ExperimentalFeatureWarning)
```

> 2.) While extending the tests, I've found one duplicate test (`./tests/test_flows/test_flow_class.py/test_fix_sequence_non_nominal()` is possibly a duplicate of `./tests/test_solph_network_classes.py/test_attributes_needing_nominal_capacity_get_it()`) and one test that does nothing at all (`./tests/test_flows/test_flow_class.py/test_source_with_full_load_time_max()`). Should I just delete them?

Please add comments instead (like `# same as test_attributes_needing_nominal_capacity_get_it()` and `# does not test anything`). Cleaning up the test - and giving them a comprehensible structure - is definitely on the to-do list. Maybe something for the developer meeting.

One additional point which I want to make explicit as the point came up in my comment: The names `min` and `max` shadow the build-in functions in and max. I would be very much in favour of renaming them (to `minimum` and `maximum`) to be able to write `min(minimum)` without any workarounds.

docs/whatsnew/v0-6-2.rst
```diff
@@ -4,6 +4,9 @@ v0.6.2
 API changes
 ###########
 
+* The `Flow` parameter now raises a `FutureWarning`, as it is to be removed
+  in a following version.
```
p-snft:
```suggestion
* The `Flow` parameter `bidirectional` now raises a `FutureWarning`,
  as it is to be removed in a following version.
```

p-snft (approved):
Makes code more understandable and maintainable. Thank you.

docs/whatsnew/v0-6-2.rst
lensum: TYSM for fixing this! That would have been confusing otherwise...

Issue #1237, opened 2025-12-09
Improve API of Results


p-snft:
The API of `solph.Results` feels a bit like a dictionary. With that in mind, there are two possible improvements:

* [x] results["objective"] does not exist (but only results.objective). The former is clearly the dictionary like way. (It also is API compatible with the old meta_results.) Fixed in #1239.
* [x] Allow checking for membership `in` a `Results` instance. Implemented in #1239.
* [ ] The function `to_df` could be renamed to `get`. If we add a default value, usage can be similar to the one of a dict.
    https://github.com/oemof/oemof-solph/blob/171bedcf6b3e54f226459df78d3b6b56a15eb8ca/src/oemof/solph/_results.py#L82
* [x] Just overwriting `__getattr__` might not be the best idea. Here is a short example:
    ```python
    results = Results(model)
    hasattr(results, "non_existing")  # results in a KeyError
    ```
    Fixed in #1239.
* [ ] What to do with `Results.timeindex`, which is redundant and not a result?

PR #1238, opened 2025-12-10
Improve API of Results

p-snft:
* Adds dict-like access for all properties. This achieves API compatibility to the old meta_results, which had `meta_results["objective"]`.
* Adds unit test to reveal the problem described in #1237.
* Removes `Results.__getattr__(..)` to fix it, for consistency, also removes other properties, that now have dict-like access. (Another reason to remove this second option is that user code will be more consistent if we gite one way for access.)


gnn:
As discussed, I'd close this in favor of PR #1239.

p-snft:
Closed in favour of #1239.

Issue #1234, opened 2025-11-26
Rename `min` and `max` attributes in the `Flow` class


lensum:
As pointed out by @p-snft in https://github.com/oemof/oemof-solph/pull/1233#issuecomment-3574448685, the `min` and `max` attributes of the `Flow` class overwrite the basic python functionality `min` and `max`. It would probably make more sense to rename them to `minimum` and `maximum`, respectively.

> One additional point which I want to make explicit as the point came up in my comment: The names min and max shadow the build-in functions in and max. I would be very much in favour of renaming them (to minimum and maximum) to be able to write min(minimum) without any workarounds.

PR #1244, opened 2025-12-23
Fix/rename min and max attributes in the flow class

AntonellaGia:
Hey, 

I tried to fix the issue https://github.com/oemof/oemof-solph/issues/1234

I renamed the attributes max and min to maximum and minimum and added a FutureWarning. 

I am not sure if I updated the docs correctly. Can somebody check this? Thank you


lensum (approved):
Thank you for the effort you've put into this! The changes you've made look good to me. I really like the way you've handled possible incompatibilities if (e.g.) `min` and `minimum` are set at the same time.

I did not check if there are any mentions of `min` or `max` in the documentation that you might have missed, but I believe that if there are any, these will be found in the following weeks and can be mended then.

p-snft (approved):
The code changes look flawless. I think, we should have short talk about good commit messages, though. Maybe at our next dev meeting. (I will fix that in the merge commit, but it would have been nice to see the reason for these changes in the commit message.)

PR #1259, opened 2026-02-14
Revision/improve results api

### LINKED ISSUE #1237: Improve API of Results

p-snft:
Resolves #1237

### DIFF
--- a/src/oemof/solph/_results.py
+++ b/src/oemof/solph/_results.py
@@ -80,7 +80,6 @@ class Results:
             | self._economy.keys()
         )
 
-    @cache
     def get(
         self,
         variable: str,
--- a/src/oemof/solph/_results.py
+++ b/src/oemof/solph/_results.py
@@ -11,7 +11,6 @@ SPDX-License-Identifier: MIT
 
 import warnings
 from collections.abc import Hashable
-from functools import cache
 
 import pandas as pd
 from oemof.tools.debugging import ExperimentalFeatureWarning
--- a/src/oemof/solph/_results.py
+++ b/src/oemof/solph/_results.py
@@ -127,9 +127,9 @@ class Results:
             index_type = tuple(dataset.index_set().subsets())[-1].name
             match index_type:
                 case "TIMEPOINTS":
-                    rv.index = self.timeindex
+                    rv.index = self._model.es.timeindex
                 case "TIMESTEPS":
-                    rv.index = self.timeindex[:-1]
+                    rv.index = self._model.es.timeindex[:-1]
                 case _:
                     rv.index = rv.index.get_level_values(-1)
         else:
@@ -148,6 +148,7 @@ class Results:
         if df is None:
             raise KeyError(f"Key '{variable}' not in Results.")
         return df
+
     #  --- END ---
 
     @staticmethod
@@ -255,6 +256,7 @@ class Results:
 
         return df_opex
 
+    # --- BEGIN: The following code can be removed for versions >= v0.7 ---
     @property
     def timeindex(self):
         """Returns timeindex of energy system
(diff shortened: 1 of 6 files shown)


src/oemof/solph/_results.py
```diff
@@ -80,8 +79,11 @@ def keys(self):
             | self._economy.keys()
         )
 
-    @cache
```
Bachibouzouk, line 83: Why isn't cache kept? I had understood it was helping processing the results faster.

src/oemof/solph/_results.py
```diff
@@ -80,8 +79,11 @@ def keys(self):
...
-    def to_df(self, variable: str) -> pd.DataFrame | pd.Series:
+    def get(
+        self,
+        variable: str,
+        value: any = None,
```
Bachibouzouk: As I understand `value` it is a default value when no variable was found. I would be for a variable name which describes this better, such as `default_value` or `value_if_not_existing`

src/oemof/solph/_results.py
```diff
@@ -236,15 +254,23 @@ def _calc_variable_costs(self):
...
     @property
     def timeindex(self):
         """Returns timeindex of energy system
 
         Returns:
             float: time index of the model
         """
+        warnings.warn(
+            "Results.timeindex will be removed in a future version."
+            + " Use index of particular results instead.",
```
Bachibouzouk: I am not sure I understand what is meant by "index of particular results"

src/oemof/solph/_results.py
```diff
@@ -80,8 +79,11 @@ def keys(self):
             | self._economy.keys()
         )
 
-    @cache
-    def to_df(self, variable: str) -> pd.DataFrame | pd.Series:
+    def get(
```
Bachibouzouk, line 82: The value of the `variable` argument can be anything or are they meant to be "invest", "flow", etc.. 
It would be helpful to have the information somewhere what the most current values are, for example in the docstring

Bachibouzouk (commented):
Thanks for preparing this PR! I have a few comments to improve readability, nothing blocking a merge though.

src/oemof/solph/_results.py
p-snft: Cache does only work if we restrict results to be hashable. (This includes the tests I implemented. They did not work with cache.)

src/oemof/solph/_results.py
p-snft: This was because I found a description of the API of plain Pythons dictionary using that name. However, current Python seems to use `key` and `default`. I changed our names accordingly.

src/oemof/solph/_results.py
p-snft: I changed the wording. The idea is that every result DataFrame or Series will have the matching timeindex anyway.

PR #1264, opened 2026-02-20
Fix Furo error when displaying ToC of result tutorial.

jfreissmann:
In the [results handling](https://oemof-solph.readthedocs.io/en/stable/basic_concepts/results.html#) section of the documentation Furo raises an error when displaying the table of contents in the main text (because it thinks its redundant information). This PR implements the Furo suggested fix to avoid the error while still keeping the ToC.


p-snft (approved):
Thanks a lot! Do you want to add this to the change log before merge?

PR #1265, opened 2026-02-20
Fix content level of time-index-tsam-tutorial.

maltefritz:
Fix conent level of "time indexes and temporal aggregation" tutorial in the documentation. I changed the underline style of the .rst file so that only the main tutorial heading is displayed in the sidebar.


p-snft:
Thanks. I already realised the problem and fixed it.

PR #1273, opened 2026-03-10
Maximum charging capacity depending on the storage level

uvchik:
Currently, the charging of a storage unit can only be limited by a constant value. However, when charging batteries, this power decreases as the SOC increases. To address this, a second constraint is to be introduced that additionally limits the charging power (see figure).

grafik

The figure shows the original limit in red and the new secondary condition in blue. This results in the yellow line for the charging power. In the current draft, this is to be defined by the following factors.

The red line is defined by the flow, as before.

```python
...
inputs={my_bus: Flow(nominal_capacity=5)}
...
```

The blue line is defined by the intersection with the red line on the one hand and by the end point at a storage level of 100% on the other.

The intersection with the red line defines the proportional storage level up to which the charging power is constant and would have a value of 0.3 in the figure, because at 30% SOC the second constraint begins to take effect (green line)

The charging power at 100% storage level is the point at which the straight line is aligned (orange line). The value is the proportion of the original charging power. This means that the value here is 0.2, because 5*0.2=1 and that is the charging power at the end (orange line).

**Please help to find good names for the two new attributes:**

1. state of charge until the charging power is constant
2. charging power at an SOC of 100%

The API could look like this:

```python
    solph.components.GenericStorage(
        label="storage",
        inputs={bel: solph.Flow(35)},
        outputs={bel: solph.Flow(40)},
        nominal_capacity=1000,
        constant_soc_until=0.4,
        fraction_of_charge_at_full_storage=0.2,
    )
```

The equation follows the basic linear equation: $y = a*x + b$

$a = -\frac{P_{\max} \cdot (1-f_{end})}{E_{\max} \cdot (1-f_{lim})}$

$b = P_{\max} \cdot \frac{1 - f_{end} \cdot f_{lim}}{1-f_{lim}}$

$f_{lim}$ is the limit soc until the charing power is constant and $f_{end}$ is the fraction of charging power that remains at the end.

In the figure below are the results of an artifical system, using the storage definition above. The green line is the SOC the blue line charging and the brown line discharging. You can find the example to create this plot in the example section "storage_soc_dependent_charging".  

grafik


uvchik:
Next steps:

* [x] Fixing the tests/checks
* [x] Writing new tests
* [x] Documentation
* [x] Doc-Strings

How to deal with Investment and TSAM usage?

p-snft:
Thanks for the initiative. I've seen this implemented using custom constraints quite a number of times already.

Looking at the proposed changes, It would be consistent to also move the old limits (`invest_relation_input_capacity`) from the `Flow`s to internal constraints. What we have are maybe (using to terms from batteries) `c_rate_limits`. And we could have a descriptor object/dataclass (similar to `Investment`), that then defines `maximum_charging`, `maximum_discharging`, and `saturation_charging`.

Alternatively, the SOC depending c-rate would be realised using a `ConvexHullConverter` (we didn't have time to discuss this at the last user meeting) that is defined separately and linked through parent/subnode structures with predefined additional constraints linking the two components.

uvchik:
We moved the limit for dis-/charging the storage from the component to the `Flows()` a while ago. There are arguments for both, keeping or moving, but I do not want to discuss this within this PR, because this is a general discussion.

Furthermore, I think the combination of a constant limit and a decreasing limit is a good idea to come closer to the real curve.

I have no idea how to deal with TSAM, because I have not used/implemented it so far. Could anyone help?

I think it would be a good idea to revise the GenericStorage component in general, but I would like to do it in a new PR and after merging this one, 

I like `saturation_charging` for the second parameter.

p-snft:
> Furthermore, I think the combination of a constant limit and a decreasing limit is a good idea to come closer to the real curve.
> I think it would be a good idea to revise the GenericStorage component in general, but I would like to do it in a new PR and after merging this one,

The idea was that we could avoid double work if we plan to have something similar outside the storage anyway. But you are right, waiting poses some risk that things won't happen.

> I have no idea how to deal with TSAM, because I have not used/implemented it so far. Could anyone help?

I am currently reworking the time index in #1272, and I fear that this might lead to another hard to resolve merge conflict. I would suggest that you raise a `NotImplementedError`.

uvchik:
In that case I would add tests and docs and then ask again for a review.

lensum:
That's a cool new feature!

> I think it would be a good idea to revise the GenericStorage component in general, but I would like to do it in a new PR and after merging this one

I agree. I thought about it multiple times now, but haven't gotten around to it yet. Maybe we should start an issue for that, where we collect the (more specific) things that need to be done?

Concerning the first parameter: what do you think of `constant_charging_storage_level`? It is quite long, but other parameters of the `GenericStorage` that refer to the SOC are also named `...storage_level`, so that should be included to minimize misunderstandings. Also, it wouldn't be too specifically geared towards batteries (we don't have to dive into CC-CV charging protocols).

uvchik:
soc_dependent_charging

I think most things are done but we still have to discuss about the name and the symbole.

 $f_{lim}$ - soc until the charing power is constant - `constant_soc_until`
 $f_{saturation}$ - fraction of charging power at saturation - `fraction_saturation_charging`
 
 
 The failing test is caused by the pandas documentation, that is down at the moment.

p-snft:
Please note that we have an example on a custom-made (Pyomo level) saturating storage. It would make sense to mention the new feature in that example (or to completely remove it).

https://github.com/oemof/oemof-solph/blob/4922d48525b9f4e8735f90853da4eff0e51a6722/examples/flexible_modelling/saturating_storage.py

PR #1274, opened 2026-03-12
Feature/return results

p-snft:
*  Model.solve() now returns `Results`.
* As the Results object assumes that the model is feasible, the return type now depends on the feasibility of the model. Thus, I introduced a warning that suggests a phase out of the option "allow_nonoptimal".

Old API (still possible):
```
model = Model(energy_system)
model.solve()
results = Results(model)
```

New API:
```
model = Model(energy_system)
results = model.solve()
```

I think, at some point we should exclusively describe the new API, as it prevents the following error:

Never worked:
```Python
model = Model(energy_system)
# not solve the model but directly:
results = Results(model)  # needs to be solved, first
```

But I think that is something for a separate PR after we have released an alpha with this status.

### DIFF
--- a/docs/whatsnew/v0-6-4.rst
+++ b/docs/whatsnew/v0-6-4.rst
@@ -4,6 +4,8 @@ v0.6.4
 API changes
 ###########
 
+* `Model.solve()` now directly returns the optimisation `Results`. (For some
+  time, they will be backwards compatible with the old meta_results.)
 
 New features
 ############
@@ -30,6 +32,7 @@ Contributors
 
 * Jonas Freißmann
 * Lena Rosin
+* Patrik Schönfeldt
 
 Acknowledgements
 ################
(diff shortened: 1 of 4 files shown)


uvchik:
For experts it should be possible to avoid the `Exception` and be able to analyse the `Model` Object.

Everything else is fine.

p-snft:
> For experts it should be possible to avoid the `Exception` and be able to analyse the `Model` Object.

I agree with the second part but don't think that this disallows an Exception. However, the `FutureWarning` is not core to this PR and it can easily be introduced at a later point in time – in case we have consensus to phase out `allow_nonoptimal`. Thus, I propose to merge this without the warning.

uvchik (approved):
I think it is a good idea to diretly return the `Results` object. Thank you for the code.

PR #1275, opened 2026-03-16
Replace processing.results by Results

p-snft:
If we want to make `solph.Results` the new default, we should already prepare phasing out `processing.results`. 

- [x] Replace in documentation (see [RTD](https://oemof-solph.readthedocs.io/en/revision-disuse-processing/index.html))
- [x] Replace in tutorials
- [x] Replace in examples (partly)

As the tests are not meant to be a training resource or documentation, I won't go through all of those. Also, I think it is better to release this with remaining examples in the old style than having even the tutorials unchnaged for a long time. (Solving #1279 is required to update at least some of the examples.)


p-snft:
@uvchik: With 7cdcf06ab13094a4f0877bd9f1e678722ed3b63f, I simplified the duals example. However, I see two issues:
1. The example is not really self-explanatory. (To me, it looks like some Pyomo magic.)
2. We do not have the duals in the result object.

What do you think? Is this actually solph related or rather something Pyomo specific?

PR #1278, opened 2026-03-26
Filter and combine result columns

gnn:
Implement a proposal for an API that allows filtering the columns of `DataFrame`s in `oemof`'s `Results` objects. Having a stable API would also allow predefining filters enabled by default or filters which are expected to be used often.

Also implement a proposal for an API that allows specifying callback functions that manipulate the `DataFrame`s in different ways than just filtering.


uvchik:
I added a function "hide_and_rename" to the examples to show what I need. For me it is already fine to have this function but we should think about a way to make the functionality of this function available for others in a simple way (callbacks, filters etc.)

p-snft:
I accidentally created a new use case for filtering: https://github.com/oemof/oemof-solph/pull/1311

(That PR suggests to replace the `Link` logic by two nested `Converter`s that do the same thing. but of course, that changes the result plotting.)

PR #1280, opened 2026-04-13
Fix NonConvex flows tsam disaggreation + route invest_status as period scalar

jdwit:
First off, thanks for this project; it is great to work with!

I ran into an issue while using NonConvex investments together with tsam time clustering. I think btw that combination can be quite common since NonConvex makes problems expensive fast, and clustering can keep solving time manageable.

## Problem

The tsam disaggregation loop in `_disaggregate_tsa_result` sliced each flow's long-format frame by position, assuming exactly one per-timestep variable per flow. That is true for plain `Flow` and convex `InvestmentFlowBlock`, because their extras (`invest`, `total`, `old`, etc.) are period scalars and already get stripped via `PERIOD_INDEXES` before the loop runs. 

However NonConvex is different: there are extra variables that are per-timestep `status`, `status_nominal`, etc. They live alongside `flow` in the same frame. The old loop silently dropped these extra values from the disaggregated output.

## Fix

- Group each flow's frame by `variable_name` before slicing, so the per-period math runs once per variable against its own row positions.

- Route `invest_status` through `PERIOD_INDEXES` - it is a period scalar rather than a per-timestep sequence, and was being iloc-sliced incorrectly.

- Add three regression tests in `tests/test_tsam_processing.py`

BTW: The contributing guide mentions contributors should add themselves to `AUTHORS.rst` - is that necessary for this PR?

### DIFF
--- a/docs/whatsnew/v0-6-4.rst
+++ b/docs/whatsnew/v0-6-4.rst
@@ -38,6 +38,7 @@ Contributors
 * Jonas Freißmann
 * Lena Rosin
 * Patrik Schönfeldt
+* Jelmer de Wit
 
 Acknowledgements
 ################
--- a/docs/whatsnew/v0-6-4.rst
+++ b/docs/whatsnew/v0-6-4.rst
@@ -22,6 +22,8 @@ Documentation
 Bug fixes
 #########
 
+* Fix TSAM result disaggregation for ``NonConvex`` and ``InvestNonConvex``
+  flows, which silently dropped per-timestep variables from the output
 
 Other changes
 #############
--- a/src/oemof/solph/processing.py
+++ b/src/oemof/solph/processing.py
@@ -36,7 +36,7 @@ from oemof.solph.components._generic_storage import GenericStorage
 from ._plumbing import _FakeSequence
 from .helpers import flatten
 
-PERIOD_INDEXES = ("invest", "total", "old", "old_end", "old_exo")
+PERIOD_INDEXES = ("invest", "total", "old", "old_end", "old_exo", "invest_status")
 
 
 def get_tuple(x):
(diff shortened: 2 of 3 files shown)


p-snft:
Thank you for the fix. Unfortunately, I might not be able to review it before Friday. Thus, let me already answer to your following question:

> BTW: The contributing guide mentions contributors should add themselves to `AUTHORS.rst` - is that necessary for this PR?

You can see this as an offer. I just aligned the text with our common practice. (Adding the name in the whatsnew file is typically done and highly appreciated. By not adding your name although we asked for it, you agree that we distribute your code without naming you.)

src/oemof/solph/processing.py
```diff
@@ -491,28 +491,42 @@ def _disaggregate_tsa_result(df_dict, tsa_parameters):
...
-                        flow_k, tsa_period["segments"], k
-                    )
-                disaggregated_flow_frames.append(flow_k)
-            period_offset += tsa_period["timesteps"] * len(
-                tsa_period["occurrences"]
-            )
-        ts = pd.concat(disaggregated_flow_frames)
-        ts.timestep = range(len(ts))
+        if flow_dict[flow].empty:
+            continue
```
p-snft: I think, it is impossible that a flow has no assigned variables, so this statement is always False.

docs/whatsnew/v0-6-4.rst
p-snft, line 42: I'd appreciate if you added your name at least here.

p-snft (changes requested):
Thanks for fixing this.

I think accepting this fix makes sense although we are planning to replace `solph.processing` in the future, are rewriting the pathway planning code, and both, TSAM support and pathway planning are experimental. The features are actively used and the replacements will not be ready too soon. But honestly, I did not go into details for suggestions how the fix could be improved. (There is just one thing I stumbled upon.)

docs/whatsnew/v0-6-4.rst
jdwit: I added my name here. When I end up making a more substantial contribution I'll add it to `AUTHORS.rst`. 😄

jdwit:
> Thanks for fixing this.
> 
> I think accepting this fix makes sense although we are planning to replace `solph.processing` in the future, are rewriting the pathway planning code, and both, TSAM support and pathway planning are experimental. The features are actively used and the replacements will not be ready too soon. But honestly, I did not go into details for suggestions how the fix could be improved. (There is just one thing I stumbled upon.)

You mention TSAM support is experimental but IMO that integration is very valuable for this framework. When working with longer time-series/larger problems time clustering quickly becomes essential. This fix may be useful to have as a reference when replacing `solph.processing`?

p-snft:
> > Thanks for fixing this.
> > I think accepting this fix makes sense although we are planning to replace `solph.processing` in the future, are rewriting the pathway planning code, and both, TSAM support and pathway planning are experimental. The features are actively used and the replacements will not be ready too soon. But honestly, I did not go into details for suggestions how the fix could be improved. (There is just one thing I stumbled upon.)
> 
> You mention TSAM support is experimental but IMO that integration is very valuable for this framework. When working with longer time-series/larger problems time clustering quickly becomes essential. This fix may be useful to have as a reference when replacing `solph.processing`?

It is experimental in the sense that we know there is a lot of room for improvement. I do not see it removed. And, I agree that your fix can serve as a reference. (Particularly, the tests will help that we do not accidentally drop functionality.)

PR #1286, opened 2026-04-24
Feature/arithmetic investment

p-snft:
The option `solph.Investment` now has the multiplication operator (`*`) defined. Imagine, you have a PV capacity of 500 W/panel, now you can give the investment costs per panel without having to calculate the specific costs: `nominal_capacity=Investment(ep_costs=200)  # 200 €/kW` and `nominal_capacity=0.5*Investment(ep_costs=100)  # 0.5 kW/panel, 100 €/panel`.

PR #1287, opened 2026-04-28
Revision/clean fake sequence

p-snft:
As #1286 might require some discussion, I already want to add these refactorings.

Changes:
* `_FakeSequence` of `None` will no longer be created. Instead, `None` will be used directly.
* Arithmetic operators are defined for the `_FakeSequence`, as not having `None` contents allows this easily.


uvchik (approved):
Thank you for your contribution. It has bothered me for a long time that, in a sequence, we only checked the first element, as this could be caused by a broken sequence as well.

PR #1288, opened 2026-04-29
Fix some typos and adjust outdated docs to new OffsetConverter behavior

fwitte:
These are minor changes to docs only: Some texts of the `OffsetConverter` docs had typos and/or were specific to the older implementation.


p-snft (approved):
Indeed, the information you replaced was outdated. Thanks for the fix.

PR #1294, opened 2026-05-18
Update contributor list

p-snft:
The `CITATION.cff` was outdated and not reflecting the current lost of contributors. I updated it and summarised a lot of people that did not give their name into one entry. Before merging, I want two things to be discussed:

The `AUTORS.rst` is a duplication of the content in a different format. What to do?
1. Remove `AUTORS.rst` and include (literal json) `CITATION.cff` in the readme.
2. Also update `AUTORS.rst`.

To allow significant new contributors to appear in "their" release, it might make sense to reorder based on newer contributions. What do you think?


lensum:
1. Sorry for taking so long to reply!
2. Summarizing unnamed developers seems like a good idea to me - thank you for doing that!
3. Generally, I don't see a point for duplicate information. For the [documentation page where the AUTHORS.rst is used](https://oemof-solph.readthedocs.io/en/latest/authors.html), maybe a simple link to the CITATION.cff would be the easiest replacement option? Otherwise I would prefer some kind of automatic creation of AUTHORS.rst from CITATION.cff in the docs pipeline - but I have no idea how to manage that.
4. Re-ordering the list with every release seems like an unnecessary amount of work and changes to me. Though I understand the wish to highlight more (recently) active maintainers... But they are already highlighted in the `whatsnew` section. So I'm leaning towards "not reordering".

lensum (commented):
I checked the names before and after the update and it seems that two names got dropped:
- Birgit Schachler
- Julian Endres

Was this deliberately? And if so: what was the rational for that?

birgits:
You can drop me from the developers list, that's fine :) I haven't worked with oemof in years..

p-snft:
> I checked the names before and after the update and it seems that two names got dropped:
> 
> * Birgit Schachler
> * Julian Endres
> 
> Was this deliberately? And if so: what was the rational for that?

Thanks for checking. It was not my intention to remove anybody. I just went through the list GitHub provides, the one that sorts by the number of current lines of code. I guess that the contributions by @birgits and @nailend have been overwritten by new contributions. While this is technically correct in  terms of copyright, I think that oreintation given by older contributions is still honorable, thus contributions should not fade out. @gnn brought another suggestion to handle the order: We could just sort alphabetically. That way, the order never gets outdated.

To me, rendering an AUTHORS.rst from the CITATION.cff in the CI pipeline sound like a good plan. Thanks for the suggestion.

jfreissmann:
Since this is still open, I took the freedom to add ORCID identifiers for @maltefritz and me. If you @p-snft think this is an inappropriate PR to do so, feel free to revert the commit. If not, maybe other can add themselves as well (or we create a separate PR for all of us that wish to add their ORCID).

p-snft:
> Since this is still open, I took the freedom to add ORCID identifiers for @maltefritz and me.

It's fine to add IDs, but I won't wait for them. What I am waiting for, however, is feedback (including a formal review) about the restructuring. I think, except for urgent cases, we should only merge after review.

jfreissmann (approved):
I was able to reproduce the intended behavior (dynamically creating AUTHORS.rst when building the docs) locally. In my opinion, this can be merged.

The only thing we could add back in again would be the removed link to the GitHub Contributors page. But I wouldn't insist on it being there.

p-snft:
> The only thing we could add back in again would be the removed link to the GitHub Contributors page. But I wouldn't insist on it being there.

To my knowledge, GitHub counts contributions based on or analogue to the latest git blame for every line. Thus, it will stop naming a contributor when all the lines somebody contributed are changed by somebody else. (I observed contributors being dropped and this is the best explanation I found.) As I think, this should not happen, I kept old contributors in our list and removed the link.

PR #1295, opened 2026-05-28
Configure Dependabot version updates

p-snft:
This adds a minimal Dependabot config, that checks pip and github workflow versions and creates PRs if updates are recommended by the bot.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,12 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://docs.github.com/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
+
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+  - package-ecosystem: "pip"
+    directory: "/" # Location of package manifests
+    schedule:
+      interval: "weekly"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -13,3 +13,4 @@ updates:
     directory: "/" # Location of package manifests
     schedule:
       interval: "weekly"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,12 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://docs.github.com/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
+
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
```
jfreissmann, line 8: I think, every `package-ecosystem` entry needs its own `directory` and `schedule` definitions, so they should be copy-pasted here as well
```suggestion
  - package-ecosystem: "github-actions"
    directory: "/"
    schedule:
      interval: "weekly"
```

.github/dependabot.yml
```diff
@@ -0,0 +1,12 @@
...
+  - package-ecosystem: "pip"
+    directory: "/" # Location of package manifests
+    schedule:
+      interval: "weekly"
```
jfreissmann, line 16: We could also explicitly set `target-branch: "dev"`, but it will be selected anyway if it is the default branch. If we ever think this could change, setting `target-branch` could make sense, otherwise we can omit it.

```suggestion
      interval: "weekly"
      target-branch: "dev"
```

jfreissmann (approved):

jfreissmann:
I think `target-branch` needs to be on the same indentation level as `directory` and `schedule` and would also be placed in both `package-ecosystem`s.

p-snft:
> I think `target-branch` needs to be on the same indentation level as `directory` and `schedule` and would also be placed in both `package-ecosystem`s.

Thanks for fixing this. I just inserted it where you put your comment.

PR #1302, opened 2026-06-04
Use  setter/getter make sure attributes which are sequences are correctly assigned

### LINKED ISSUE #1260: Casting to series is inconsistent
**Describe the bug**
In the init, we cast constant values to iterables. However, the interface to overwrite the values does not.

**To Reproduce**
Steps to reproduce the behaviour:
1.  Construct a converter with a constant conversion factor (e.g. `conversion_factors={node: 1}`)
2.  Overwrite the conversion factor `converter.conversion_factors[node] = 1`
3.  From the syntax, you would expect that you just have overwritten the value with the same value. However:
4.  Construction of the problem fails with "int is not subscriptable".

**Expected behavior**
Setting at and after construction should be equivalent.

Bachibouzouk:
Fixes #1260

### DIFF
--- a/docs/whatsnew/v0-6-5.rst
+++ b/docs/whatsnew/v0-6-5.rst
@@ -37,6 +37,7 @@ Contributors
 * Jelmer de Wit
 * Patrik Schönfeldt
 * Pierre Francois Bachibouzouk
+* Stephan Günther
 
 Acknowledgements
 ################
--- a/docs/whatsnew/v0-6-5.rst
+++ b/docs/whatsnew/v0-6-5.rst
@@ -18,6 +18,9 @@ Bug fixes
 
 * Fix TSAM result disaggregation for ``NonConvex`` and ``InvestNonConvex``
   flows, which silently dropped per-timestep variables from the output.
+* Fix inconsistency in the assignment of sequences using scalar variables;
+  e.g. `flow = Flow(variable_costs=5)` worked, but `flow.variable_costs = 5`
+  failed when constructing the model.
 
 Other changes
 #############
@@ -33,7 +36,7 @@ Contributors
 
 * Jelmer de Wit
 * Patrik Schönfeldt
-
+* Pierre Francois Bachibouzouk
 
 Acknowledgements
 ################
--- a/tests/test_plumbing.py
+++ b/tests/test_plumbing.py
@@ -85,8 +85,8 @@ def test_sequence_dict():
 
 def test_Apply_descriptor():
     class TestClass:
-        x = Apply()
-        y = Apply("y")
+        x = Apply(sequence)
+        y = Apply(converter=sequence, default="y")
 
         def __init__(self, x):
             self.x = x
@@ -108,7 +108,7 @@ def test_Apply_descriptor():
 
 def test_using_Apply_to_convert_to_int():
     class TestClass:
-        x = Apply(type_converter=int)
+        x = Apply(converter=int)
 
         def __init__(self, x):
             self.x = x
--- a/tests/test_plumbing.py
+++ b/tests/test_plumbing.py
@@ -106,20 +106,26 @@ def test_Apply_descriptor():
     assert (i2.y == [1, 2]).all()
 
 
-def test_using_Apply_to_convert_to_int():
+def test_using_Apply_to_convert_to_numeric():
     class TestClass:
-        x = Apply(converter=int)
+        i = Apply(converter=int)
+        f = Apply(converter=float)
 
         def __init__(self, x):
-            self.x = x
+            self.i = x
+            self.f = x
 
     i1 = TestClass(1)
-    assert isinstance(i1.x, int)
-    assert i1.x == 1
+    assert isinstance(i1.i, int)
+    assert isinstance(i1.f, float)
+    assert i1.i == 1
+    assert i1.f == 1
 
     i2 = TestClass(2.0)
-    assert isinstance(i2.x, int)
-    assert i2.x == 2
+    assert isinstance(i2.i, int)
+    assert isinstance(i2.f, float)
+    assert i2.i == 2
+    assert i2.f == 2
 
 
 def test_sequence():
--- a/tests/test_plumbing.py
+++ b/tests/test_plumbing.py
@@ -12,7 +12,8 @@ import pandas as pd
 import pytest
 
 from oemof.solph._plumbing import _FakeSequence
-from oemof.solph._plumbing import SequenceProperty
+from oemof.solph._plumbing import SequenceDict
+from oemof.solph._plumbing import ConvertingProperty
 from oemof.solph._plumbing import sequence
 from oemof.solph._plumbing import valid_sequence
 
@@ -61,10 +62,23 @@ def test_fake_sequence():
         assert (seq2 == np.array(2 * [84])).all()
 
 
+def test_sequence_dict():
+    d0 = SequenceDict()
+    d0["a"] = 5
+    assert isinstance(d0["a"], _FakeSequence)
+    assert d0["a"][0] == 5
+
+    d1 = SequenceDict({"b": [1, 2, 3], 3: 20})
+    assert isinstance(d1["b"], np.ndarray)
+    assert isinstance(d1[3], _FakeSequence)
+    assert (d1["b"] == [1, 2, 3]).all()
+    assert d1[3][0] == 20
+
+
 def test_sequence_property():
     class TestClass:
-        x = SequenceProperty()
-        y = SequenceProperty("y")
+        x = ConvertingProperty()
+        y = ConvertingProperty("y")
 
         def __init__(self, x):
             self.x = x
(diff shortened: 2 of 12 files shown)


p-snft (approved):
Interesting, that I can approve my own code.

tests/test_plumbing.py
```diff
@@ -60,6 +62,58 @@ def test_fake_sequence():
...
+            if x is not None:
+                self.y = x
+
+    i0 = TestClass(None)
+    assert i0.x is None
+    assert i0.y == "y"
+
+    i1 = TestClass(1)
+    assert i1.x[0] == 1
+    assert i1.y[42] == 1
```
Bachibouzouk, line 102: Is it because of Hitchhiker's guide to the galaxy?

Bachibouzouk:
> Interesting, that I can approve my own code.

And I thus cannot approve the PR I reviewed because I created it in the first place. I do approve it though :)

gnn:
So, I read the changes and I quite like them, but I got a few comments.
I like the fact that `SequenceProperty` got generalized to `ConvertingProperty` but the same generalisation can be applied to `SequenceDict`. The conversion function can be an argument do `SequenceDict`. That would change the constructor to not be a drop in replacement for a regular `dict`, but I don't think that's a problem, because as it's been used through a descriptor, we know exactly how the constructor is called and can cater for that.
There's also a "bug" in `SequenceDict` because `dict.update` doesn't go through `__setitem__`. Properly subclassing `dict` is a bit fidgety, but I've got an implementation from a previous project lying around, so I hope that I can check what else can be missed when doing this and whip up a prototype implementation that also implements the proposal above on Tuesday.
I also have a few other implementation remarks. But since it's a bit late (actually early) right now, I hope it's OK if I continue writing them down on Tuesday, too.

p-snft:
> I like the fact that `SequenceProperty` got generalized to `ConvertingProperty` but the same generalisation can be applied to `SequenceDict`. The conversion function can be an argument do `SequenceDict`. That would change the constructor to not be a drop in replacement for a regular `dict`, but I don't think that's a problem, because as it's been used through a descriptor, we know exactly how the constructor is called and can cater for that.

IMHO, the `SequenceDict` is pretty special anyway. We have a special container that casts its members to a special datatype. When generalising this to a dictionary using any callable, the question would arise why can't it be a list or whatever else. Thus, I stopped right away and just implemented what we really need.

 For the `ConvertingProperty`, I particularly liket that I could use a standard datatype for testing. So, the test is really a unit test and does not rely on `sequence(...)`.

> There's also a "bug" in `SequenceDict` because `dict.update` doesn't go through `__setitem__`. Properly subclassing `dict` is a bit fidgety, but I've got an implementation from a previous project lying around, so I hope that I can check what else can be missed when doing this and whip up a prototype implementation that also implements the proposal above on Tuesday. I also have a few other implementation remarks. But since it's a bit late (actually early) right now, I hope it's OK if I continue writing them down on Tuesday, too.

I found that the [Python doc](https://docs.python.org/3/library/collections.html) suggests to subclass `collections.UserDict` "for easier dict subclassing". I change my code accordingly.

p-snft:
Only link checks fail (because of timeouts).

gnn:
> > I like the fact that `SequenceProperty` got generalized to `ConvertingProperty` but the same generalisation can be applied to `SequenceDict`. The conversion function can be an argument do `SequenceDict`. That would change the constructor to not be a drop in replacement for a regular `dict`, but I don't think that's a problem, because as it's been used through a descriptor, we know exactly how the constructor is called and can cater for that.
> 
> IMHO, the `SequenceDict` is pretty special anyway. We have a special container that casts its members to a special datatype.

It's correct that the `SequenceDict` is only used in very special circumstances, namely `conversion_factors`, I think. But IMHO that means that the `SequenceDict` could also handle the other code that's unique to this situation, like setting defaults and handling missing keys. So when creating a descriptor class specifically for conversion factors, I'd say go all the way and try to really make it specific for this case.
But that's not the reason, I originally mentioned generalising `SequenceDict` to a `ValueConverter`. My original reasoning had something to do with readability at the use site, but I now realized, that `SequenceDict` so well named, that my argument doesn't apply to this. I also managed to implement a generalized solution, but I think it's so "complicated" that it's actually a good idea to keep `SequenceDict` around.

```python
def ApplyToValues(
    converter,
    __setitem__=lambda self, key, value: self.data.update(
        {key: self.converter(value)}
    ),
):
    """Create `dict`s that apply `converter` to values upon insertion.

    Create a class that can be used as a drop-in replacement for a
    `dict`, but which calls `converter` on values before they are
    inserted and stores the result.

    """

    return type(
        f"ApplyToValues",
        (UserDict,),
        {"__setitem__": __setitem__, "converter": staticmethod(converter)},
    )
```


> For the `ConvertingProperty`, I particularly liket that I could use a standard datatype for testing. So, the test is really a unit test and does not rely on `sequence(...)`.

That directly touches upon the other comments I have for `ConvertingProperty`, because it's concept actually generalises beyond data types to arbitrary functions, meaning that you don't have to convert the type of the value passed in. For example if we ever try to put unit awareness into code and don't encode different units via different types, `ConvertingProperty` could also be used to normalise values to standard units without changing their types. For this reason, I'd advocate to changing `type_converter` to just `converter`.
I'd also advocate for swapping the order of `default` and `converter`, because `converter` is the principal argument that always has to be supplied and `default` is a secondary argument which isn't really necessary. This ties into `converter` not defaulting to `sequence`. I understand that the consistent conversion to `sequence`s is the reason for implementing this, but whoever reads the code where `ConvertingProperty` is used might not be aware of this and has no idea to what incoming values are converted, and from the context of just seeing `x = ConvertingProperty()`, it wouldn't be obvious to me what a sensible default would be. For that reason, I would argue for `converter` to not have a default at all and be a mandatory explicit argument.
I'd also change the name of `ConvertingProperty`. The `Property` suffix isn't really necessary. If we really want to keep that information alive, I'd prefer a `properties` module over the `Property` suffix, but I really don't think that's necessary. Most libraries I know that employ descriptors don't just give their descriptors descriptive names, like `field` for the built-in `dataclass`es or `mapped_column` and `relationship` from SQLAlchemy. So I suggest renaming `ConvertingProperty` to `Apply`. That way, when seeing `x = Apply(sequence)` or `x = Apply(SequenceDict)` it's easier to understand what's happening without reading the documentation or the descriptor code.
Last but not least, there's no need to use the `__set_name___` machinery to create an underscore prefixed attribute on the instance using the descriptor class. The data can actually be stored in a dictionary on the descriptor.

```python
class Apply:
    """Apply `converter` to values assigned to this attribute.

    Whenever a `value` is assigned to an attribute that uses this
    descriptor, what's actually stored is the result of
    `converter(value)`, e.g.:

    >>> class Inverter:
    ...     inverted = Apply(lambda x: -x)
    ...
    >>> one = Inverter()
    >>> one.inverted = 1
    >>> one.inverted
    -1

    """

    nodefault = object()

    def __init__(self, converter, default=nodefault):
        self.converter = converter
        self.default = default
        self.data = {}
        self.name = "unknown attribute"

    def __get__(self, obj, objtype=None):
        if (id(obj) not in self.data) and (self.default is self.nodefault):
            raise AttributeError(
                f"attribute '{self.name}' of '{objtype.__name__}'"
                " object accessed before being assigned a value"
            )
        return self.data[id(obj)]

    def __set__(self, obj, value):
        self.data[id(obj)] = self.converter(value)

    def __set_name__(self, owner, name):
        self.name = name
```

I know, I'm still using `__set_name__` but only to get a better error message. :) Using a dictionary keeps all the data in the descriptor instance and doesn't accidentally clash with an underscored attribute created if someone maybe tries to combine `property` and `Apply`.


> > There's also a "bug" in `SequenceDict` because `dict.update` doesn't go through `__setitem__`. Properly subclassing `dict` is a bit fidgety, but I've got an implementation from a previous project lying around, so I hope that I can check what else can be missed when doing this and whip up a prototype implementation that also implements the proposal above on Tuesday. I also have a few other implementation remarks. But since it's a bit late (actually early) right now, I hope it's OK if I continue writing them down on Tuesday, too.
> 
> I found that the [Python doc](https://docs.python.org/3/library/collections.html) suggests to subclass `collections.UserDict` "for easier dict subclassing". I change my code accordingly.

I was about to argue against using `UserDict`, but then I played around with it a bit and learned something new. I did not know, that `UserDict` automatically called `__setitem__` even when `update` is used. Together with the fact that the other appeal of `UserDict`s is, that their contents are accessible via a `data` attribute, the `SequenceDict` code can be simplified to

```python
class SequenceDict(UserDict):
    """Convert values to `sequence`s upon insertion.

    A drop-in replacement class for `dict`s, that calls `sequence` on
    values before they are inserted and stores the result.

    """

    def __setitem__(self, key, value):
        self.data[key] = sequence(value)
```

Please excuse the fact that I didn't yet manage to create commits from this code into. If that's OK, create a branch with proposal commits that you can cherry pick onto this branch.


Lastly, even though the following point isn't contended by me anymore, I had the first paragraph written before I realized that so this is more in the spirit of sharing thoughts:

> When generalising this to a dictionary using any callable, the question would arise why can't it be a list or whatever else.

For me, that this is because the implementation of Python's built-in container classes makes it rather hard to generalize arbitrary containers. For example it would be hard to implement one descriptor class that generalizes both, `dict`s and `list`s, because just overriding `__setitem__` isn't enough and which other methods also have to be overridden depends on which specific container type one wants to emulate.
Although, with the "new knowledge" of how `UserDict` works, there might be some interesting solutions here, so I'm going to play around a bit more to learn more about what's possible in Python.

docs/whatsnew/v0-6-5.rst
```diff
@@ -33,7 +36,7 @@ Contributors
 
 * Jelmer de Wit
 * Patrik Schönfeldt
-
+* Pierre Francois Bachibouzouk
```
p-snft, line 41:
```suggestion
* Pierre Francois Bachibouzouk
* Stephan Günther
```

PR #1303, opened 2026-06-10
Simplify fetching results from pyomo

Bachibouzouk:
This is re-factoring the class Results to be clearer and cleaner.


src/oemof/solph/_results.py
p-snft:
```suggestion
default: Any = None,
) -> pd.DataFrame | pd.Series | None:
```

src/oemof/solph/_results.py
```diff
@@ -148,15 +148,18 @@ def get(
...
-            match index_type:
-                case "TIMEPOINTS":
-                    rv.index = self._model.es.timeindex
-                case "TIMESTEPS":
-                    rv.index = self._model.es.timeindex[:-1]
-                case _:
-                    rv.index = rv.index.get_level_values(-1)
+            if rv.empty is True:
+                rv = default
+            else:
```
p-snft:
```suggestion
            if not rv.empty:
```

p-snft (commented):
Thanks for these simplifications. Now that I see it, I would simplify `get` a bit further by early return statements. With your addition, we now have two lines reading `rv = default` but both times, the next thing to happen will just be `return rv`. Instead, we could just `return default` in the end, if nothing has been returned, yet.

p-snft:
I wanted to click on "ready for review" but then the page loaded the running checks and "close pull request" was right in the same spot. XD

src/oemof/solph/_results.py
```diff
@@ -48,32 +49,32 @@ def __init__(self, model: ConcreteModel):
...
+                self._variables[key][occurence] = variable
+            else:
+                # Only left option should be
+                # self._variables[key][occurence] == variable.
+                # Iterated over the same thing twice.
+                # Case for debugging purposes.
+                # We should avoid useless iterations.
+                pass
 
         # adss additional keys for the calculation of opex and capex
```
p-snft:
```suggestion
        # adds additional keys for the calculation of opex and capex
```

p-snft (approved):
Thank you! I just simplified this a bit further, and it's now ready.

PR #1305, opened 2026-06-11
Feature/fake sequence arithmetic

p-snft:
Implements arithmetic operators (+, -, *, /) for `_FakeSequence`.

Both, `__add__` and `__mul__` also have the right hand operation defined, as the operator commutes. For `__truediv__` and `__sub__`, however, I think it makes sense that you cannot divide by a sequence or subtract a sequence from arbitrary other objects.


src/oemof/solph/_plumbing.py
```diff
@@ -183,6 +198,106 @@
                 "Length needs to be defined for casting to numpy."
             )
 
+    def all(self):
+        return bool(self.value)
+
+    def any(self):
+        return bool(self.value)
+
+    def __bool__(self):
```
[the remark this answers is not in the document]
p-snft: This is aligned with numpy arrays.

tests/test_plumbing.py
```diff
@@ -18,48 +18,252 @@
...
-        assert seq.size is None
+    with pytest.raises(ValueError, match="Length needs to be defined"):
+        seq.to_numpy()
+    assert (seq.to_numpy(length=5) == np.array(5 * [42])).all()
 
-        assert seq.max() == 42
-        assert seq.min() == 42
-        assert seq.value == 42
-        assert seq.sum() == np.inf
+    assert len(seq) == 0
```
uvchik: Shouldn't it be `None` instead of zero? For me, the value is simply undefined.

tests/test_plumbing.py
```diff
@@ -18,48 +18,252 @@
...
 
-        assert str(seq) == "[42.0, 42.0, ..., 42.0]"
+    seq.size = 2
+    assert seq.size == 2
+    assert len(seq) == 2
```
uvchik: If the size is "2", then the following test should fail, shouldn't it?

`assert seq[5] == 42`

uvchik (commented):
Thank you for this PR, it addresses an important issue.

I would use the sequence-function if I do not know whether the input is scalar oder list-like. So I would expect the tests to cover both cases e.g.

```python
seq1 = sequence(42)
seq2 = sequence([42, 43, 44])
other1 = 42
other2 = [42, 43, 44]

for other in [other1, other2]:
    for seq0 in [seq1, seq2]:
        assert (seq0 <= other).any()
        assert (seq0 == other).any()
        assert (seq0 >= other).any()
```

In case of all we have to distinguish somehow or use the same value everywhere.

tests/test_plumbing.py
p-snft: This is because I want to be compatible with Python: The Python documentation specifies that `__len__` must return a non-negative integer (≥0). I also put it to the changelog: „_FakeSequences with no defined length will now evaluate to have `len(fake_sequence) == 0`, as len is defined to return integers only.”

When we want to adhere to that, the only other option would be raising an error.

tests/test_plumbing.py
p-snft: I see the size of a `_FakeSequence` as something fluid. The number gives something like a "preferred length", it technically still is defined everywhere (in contrast to being overwritten by a `np.array`.) Thus, I would leave it as it is.

PS: The old behaviour was to use `_FakeSequence.size` to track the last (virtual) element that was accessed. This made the status of the object rely on something that looks like a reading access, thus we dropped that behaviour.

p-snft:
> Thank you for this PR, it addresses an important issue.
> 
> I would use the sequence-function if I do not know whether the input is scalar oder list-like. So I would expect the tests to cover both cases e.g.
> 
> ```python
> seq1 = sequence(42)
> seq2 = sequence([42, 43, 44])
> other1 = 42
> other2 = [42, 43, 44]
> 
> for other in [other1, other2]:
>     for seq0 in [seq1, seq2]:
>         assert (seq0 <= other).any()
>         assert (seq0 == other).any()
>         assert (seq0 >= other).any()
> ```
> 
> In case of all we have to distinguish somehow or use the same value everywhere.

In `test_fake_sequence_compare`, this is done but with explicit comparison to an `integer`, a `list`, another `_FakeSequence`, and a `np.array`. There is a problem however, that having  a `np.array` as the LHS will rely on the `_FakeSequence` to be cast to a `np.array`, which is only possible if it has a finite length. The only solution probably is implementing our own functional comparison that always puts the `_FakeSequence` first, instead of supporting the ordinary comparison operators (e.g. `_FakeSequence.le(a, b)` instead of `a <= b`).

tests/test_plumbing.py
uvchik: Return a zero seems strange to me, "inf" would be okay but it is not integer. Then raising an error would be the solution e.g. something like this.

```python
TypeError: A naive sequence has no len()
```

tests/test_plumbing.py
uvchik: I don't like the idea of a fluid size. Maybe we should say, that a `_FakeSequence` has no length! This would make the error message above more clear and if you need to have a size you do not need a `_FakeSequence` you can just use an array with `seq.to_array(8760)`

p-snft:
A bit more of discussion with Uwe lead to the idea that the `_FakeSequence` might work without a length. This consists of two aspects:

1. Once we really want a length, we could convert it to a numpy array instead of setting a length.
2. Operators use `sequence` and in the form:
```python
def __operator__(self, other):
    return sequence(self._value.__operator__(other))
```

This would solve the inconsistency we have at the moment that casting is not always possible. (Casting to lists will no longer be possible at all.) To tackle the problem with numpy array as left hand side of the operators, instead it would be possible to cast `_FakeSequence` to a scalar. (This second idea needs verification if it really works in all cases.)

uvchik:
That means that if I multiply a `_FakeSequence` by a ‘float’, it remains an `_FakeSequence`; if I multiply the  `_FakeSequence` by an `Array`, I get an `Array`; and if I multiply it by a `Series`, I also get an `Array`, right?

I agree with that, but I think it would be slightly better if the `Series` also resulted in a `Series`.

src/oemof/solph/_plumbing.py
```diff
@@ -188,65 +177,72 @@ class _FakeSequence:
...
 
     def __mul__(self, other):
-        return sequence(self._value * other, length=self._length)
+        return sequence(self.value * other)
 
     __rmul__ = __mul__
 
     def __truediv__(self, other):
-        return 1 / other * self
+        return 1 / other * self.value
```
juncienf: I don't know if this is correct: 1 / other * self.value
And, do we need the reflected version?

tests/test_plumbing.py
```diff
@@ -167,18 +267,18 @@ def test_valid_sequence():
     with pytest.raises(ValueError):
         valid_sequence(np_array, 1337)
 
-    fake_sequence = _FakeSequence(42)
-    assert valid_sequence(fake_sequence, 5)
-    assert len(fake_sequence) == 5
+    # fake_sequence = _FakeSequence(42)
```
juncienf: I need review for "test_valid_sequence". I don't know if this is valid for the new _FakeSequence

src/oemof/solph/_plumbing.py
p-snft: I originally did it this way as it can reuse the (already defined) multiplication operator. With the new logic, probably
```suggestion
        return sequence(self.value / other)
```
is an easier solution. (Honestly, I'd have to test if `__rtruediv__` is needed for our purposes. I think, it is okay to not implement it, as we also have no explicit `__radd___`, etc.)

tests/test_plumbing.py
p-snft: Honestly, this is now a design decision. As the `_FakeSequence` can no longer have a size, if feel like `test_valid_sequence` is no longer needed, but I have to check that closer.

src/oemof/solph/_plumbing.py
juncienf: Included your suggestion and decided to add `__rtruediv__` (Adding it won't hurt)

tests/test_plumbing.py
juncienf: I made some adjustments. Let me know if you'd like to include something else.

juncienf:
All the tests I had in mind are ready. Waiting for review now (@p-snft)

p-snft:
> That means that if I multiply a `_FakeSequence` by a ‘float’, it remains an `_FakeSequence`; if I multiply the `_FakeSequence` by an `Array`, I get an `Array`; and if I multiply it by a `Series`, I also get an `Array`, right?
> 
> I agree with that, but I think it would be slightly better if the `Series` also resulted in a `Series`.

In my opinion, this has to be decided separately. We have the existing convention that `sequence` casts everything to a numpy array. I agree that this might be changed, but now – as `_FakeSequence` just uses `sequence` – it can be changed directly in `sequence`.

src/oemof/solph/components/_offset_converter.py
```diff
@@ -256,74 +253,6 @@ def __init__(
     def constraint_group(self):
         return OffsetConverterBlock
 
-    # --- BEGIN: To be removed for versions >= v0.7 ---
```
p-snft, line 259: I asked @fwitte about this and he agreed that we might remove this without neccesarily increasing the version counter. The current API is there for over a year now and far more understandable.

src/oemof/solph/_plumbing.py
```diff
@@ -53,8 +52,11 @@ def sequence(iterable_or_scalar, length=None):
     if iterable_or_scalar is None:
         return None
     if isinstance(iterable_or_scalar, _FakeSequence):
-        iterable_or_scalar.size = length
-        return iterable_or_scalar
+        if length is not None:
+            return np.full(shape=length, fill_value=iterable_or_scalar.value)
```
uvchik, line 59: I woul add a sentence to the docstring to explain this feature and maybe an example as the one below to make illustrate it.   

```python
    >>> a = np.array([5] * 10)
    >>> a
    array([5, 5, 5, 5, 5, 5, 5, 5, 5, 5])
    >>> b = 5
    >>> sequence(sequence(a), length=10)
    array([5, 5, 5, 5, 5, 5, 5, 5, 5, 5])
    >>> sequence(sequence(b), length=10)
    array([5, 5, 5, 5, 5, 5, 5, 5, 5, 5])
```

uvchik (commented):
I got the impression that there are changes that do not belong to this PR. I will not be able to review that without more details. Maybe someone else could review that.

Thank you for the contribution. For me it looks fine, now. Just one little comment.

src/oemof/solph/_plumbing.py
p-snft: I agree that the docstring can be more descriptive. This PR was not focusing on this but just touched `sequence` because it relied on `_FakeSequence` allowing to have a length.

The code example you show here will never see the line you comment (because of `if isinstance(iterable_or_scalar, _FakeSequence)`.

uvchik (approved):

PR #1306, opened 2026-06-12
name disaggregated tsa storage soc storage_content

jdwit:
When working with results from a model that uses tsam clustering, I noticed an inconsistency in column naming: non-clustered results use `storage_content` for tracking state of charge and clustered results use the column name `soc`.

This should be consistent and results structure should be stable no matter if using temporal clustering or not.

I think currently the `views` module already misses it in the current version and this will be fixed by this PR (see `views.py:401-402`, I think since they are currently named `soc`, results are silently dropped).


p-snft (approved):
Thank you for submitting this fix!

PR #1307, opened 2026-06-18
Add HiGHS solver

### LINKED ISSUE #855: Support of HiGHS solver
@sasa821 contacted the devs of HiGHS to ask whether they would make it work with pyomo. They replied that a pyomo-HiGHS interface is being developed as it was wished by both teams. However it is not ready yet. The best one can do is to use pyomo to write a out the model as an .mps file and then call HiGHS from the command line or from python.

On pyomo, the only open issue mentionning HiGHS is https://github.com/Pyomo/pyomo/issues/2094

This would be a great improvement if we can use HiGHS over cbc, and both @sasa821 and I are motivated to try it out, would anyone else also be interested to help?

uvchik:
I want to use the HiGHS solver with oemof-solph (see #855 for earlier discussions)

## Installation

`pip install highspy` found here: https://pyomo.readthedocs.io/en/6.9.5/getting_started/solvers.html

## Example

I started with a simple example from the oemof-solph example directory and changed the solver name to `appsi_highs`, accoding to https://or.stackexchange.com/a/12688

```
solver = SolverFactory("appsi_highs")
```

In oemof-solph solph it looks like this
```
model.solve(solver="appsi_highs")
```

## Adapting the `Model` class

This collides with the `solver_io` parameter, so I removed it for "appsi"-solver.

## Stucked

I got stuck with the following error. **Can anybody help?**

```python
  File "...oemof-solph\src\oemof\solph\_models.py", line 485, in solve
    solver_results = opt.solve(self, **solve_kwargs)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "...my_env\Lib\site-packages\pyomo\contrib\appsi\base.py", line 1607, in solve
    if hasattr(model, 'dual') and model.dual.import_enabled():
                                  ^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'import_enabled'
```

### DIFF
--- a/examples/storage_investment/v1_invest_optimize_all_technologies.py
+++ b/examples/storage_investment/v1_invest_optimize_all_technologies.py
@@ -220,7 +220,7 @@ def main(optimize=True):
 
     # if tee_switch is true solver messages will be displayed
     logging.info("Solve the optimization problem")
-    om.solve(solver="appsi_highs", solve_kwargs={"tee": True})
+    om.solve(solver="highs", solve_kwargs={"tee": True})
 
     ##########################################################################
     # Check and plot the results
@@ -231,8 +231,8 @@ def main(optimize=True):
 
     electricity_bus = solph.views.node(results, "electricity")
 
-    meta_results = solph.processing.meta_results(om)
-    pp.pprint(meta_results)
+    # meta_results = solph.processing.meta_results(om)
+    # pp.pprint(meta_results)
 
     # returns a pandas Series with all scalar values (investment, total) of
     # components connected to the electricity bus
--- a/examples/storage_investment/v1_invest_optimize_all_technologies.py
+++ b/examples/storage_investment/v1_invest_optimize_all_technologies.py
@@ -220,7 +220,7 @@ def main(optimize=True):
 
     # if tee_switch is true solver messages will be displayed
     logging.info("Solve the optimization problem")
-    om.solve(solver="cbc", solve_kwargs={"tee": True})
+    om.solve(solver="appsi_highs", solve_kwargs={"tee": True})
 
     ##########################################################################
     # Check and plot the results
--- a/examples/storage_investment/v1_invest_optimize_all_technologies.py
+++ b/examples/storage_investment/v1_invest_optimize_all_technologies.py
@@ -231,8 +231,6 @@ def main(optimize=True):
 
     electricity_bus = solph.views.node(results, "electricity")
 
-    # meta_results = solph.processing.meta_results(om)
-    # pp.pprint(meta_results)
 
     # returns a pandas Series with all scalar values (investment, total) of
     # components connected to the electricity bus
--- a/src/oemof/solph/_models.py
+++ b/src/oemof/solph/_models.py
@@ -471,11 +471,17 @@ class Model(po.ConcreteModel):
         if tc == appsi.base.TerminationCondition.optimal:
             appsi_results.solution_loader.load_vars()
             if self.dual is not None:
-                for c, val in appsi_results.solution_loader.get_duals().items():
-                    self.dual[c] = val
+                try:
+                    for c, val in appsi_results.solution_loader.get_duals().items():
+                        self.dual[c] = val
+                except RuntimeError:
+                    pass  # duals not available for MIP models
             if self.rc is not None:
-                for v, val in appsi_results.solution_loader.get_reduced_costs().items():
-                    self.rc[v] = val
+                try:
+                    for v, val in appsi_results.solution_loader.get_reduced_costs().items():
+                        self.rc[v] = val
+                except RuntimeError:
+                    pass  # reduced costs not available for MIP models
             logging.info(
                 f"Optimization successful with condition: {tc.name}"
             )
--- a/src/oemof/solph/_models.py
+++ b/src/oemof/solph/_models.py
@@ -444,24 +444,53 @@ class Model(po.ConcreteModel):
         solve_kwargs=None,
         cmdline_options=None,
     ):
-        msg = (
-            "Using the 'HiGHS'-solver is experimental in solph.\n Using "
-            "options and accessing the meta results might be different."
-        )
-        warnings.warn(msg)
+        if solve_kwargs is None:
+            solve_kwargs = {}
+        if cmdline_options is None:
+            cmdline_options = {}
+
         opt = appsi.solvers.Highs()
-        results = opt.solve(self)
-        if results.termination_condition.value == 5:
+        opt.config.load_solution = False
+
+        if solve_kwargs.get("tee"):
+            opt.config.stream_solver = True
+
+        opt.highs_options = cmdline_options
+
+        appsi_results = opt.solve(self)
+        tc = appsi_results.termination_condition
+
+        solver_results_dict = {
+            "termination_condition": tc.name,
+            "best_feasible_objective": appsi_results.best_feasible_objective,
+            "best_objective_bound": appsi_results.best_objective_bound,
+        }
+        self.es.results = solver_results_dict
+        self.solver_results = solver_results_dict
+
+        if tc == appsi.base.TerminationCondition.optimal:
+            appsi_results.solution_loader.load_vars()
+            if self.dual is not None:
+                for c, val in appsi_results.solution_loader.get_duals().items():
+                    self.dual[c] = val
+            if self.rc is not None:
+                for v, val in appsi_results.solution_loader.get_reduced_costs().items():
+                    self.rc[v] = val
             logging.info(
-                f"Optimization successful with condition: "
-                f"{results.termination_condition.name}"
+                f"Optimization successful with condition: {tc.name}"
             )
+            return Results(self)
         else:
-            warnings.warn("Optimization ended with an unclear status ")
-            print(results)
-        self.solver_results = results
-        # results = Results(self)  # ToDo: This does not work!!
-        return results
+            msg = (
+                f"The solver did not return an optimal solution. "
+                f"Instead the optimization ended with\n"
+                f"       - termination condition: {tc.name}"
+            )
+            if allow_nonoptimal:
+                warnings.warn(msg, UserWarning)
+                return solver_results_dict
+            else:
+                raise RuntimeError(msg)
 
     def solve(
         self,
--- a/src/oemof/solph/_models.py
+++ b/src/oemof/solph/_models.py
@@ -21,6 +21,7 @@ from logging import getLogger
 
 from oemof.tools import debugging
 from pyomo import environ as po
+from pyomo.contrib import appsi
 from pyomo.core.plugins.transform.relax_integrality import RelaxIntegrality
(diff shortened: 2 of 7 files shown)


jfreissmann:
At [maltefritz/owp_milp_optimization](https://github.com/maltefritz/owp_milp_optimization), we use a somewhat hacky approach similarly to the one below. The config options that are commented out don't really work for us (e.g. we don't get a solver log stream to the console, but only a log file works). Maybe this can help push you in the right direction.

```python
from pyomo.contrib import appsi

# ...

opt = appsi.solvers.Highs()
opt.config.mip_gap = self.param_opt['MIPGap']
opt.config.logfile = logpath
if self.param_opt['TimeLimit'] is not None:
    opt.config.time_limit = self.param_opt['TimeLimit']
# opt.config.stream_solver = True
# opt.highs_options['output_flag'] = True
# opt.highs_options['log_to_console'] = True
try:
    results = opt.solve(self.model)
except RuntimeError:
    results = appsi.solvers.highs.HighsResults(opt)

```

uvchik:
Thank you, @jfreissmann. Actually, I found an old implementation using a specific "solve_highs" method within the Model class.  It may lead to additional maintainance effort. It works with the old `processing_results` method but not with the new Results-object. I had hoped that, with the idea above, the results API would be more similar to the current results API, since the SolverFactory is used. To implement this properly in Solph, `solve_highs` must return a valid Results object, even if this requires some additional code. It won’t work without a Results object, as everything else is built on top of it. Otherwise, it will always remain at a ‘hacky’ level.

fwitte:
> Thank you, @jfreissmann. Actually, I found an old implementation using a specific "solve_highs" method within the Model class. It may lead to additional maintainance effort. It works with the old `processing_results` method but not with the new Results-object. I had hoped that, with the idea above, the results API would be more similar to the current results API, since the SolverFactory is used. To implement this properly in Solph, `solve_highs` must return a valid Results object, even if this requires some additional code. It won’t work without a Results object, as everything else is built on top of it. Otherwise, it will always remain at a ‘hacky’ level.

I asked claude code to come up with a solution, looks promising: [feature/highs-compatibility](https://github.com/oemof/oemof-solph/tree/feature/highs-compatibility)
should we merge the branch into here @uvchik?

fwitte:
To test it you just need to run any of the examples with `solver="highs"` instead of whatever the default is.

uvchik:
@fwitte  Thank you. Go ahead 😄

fwitte:
> @fwitte Thank you. Go ahead 😄

done, we should add some tests I guess, and some explanations as to how exactly that implementation works

examples/storage_investment/v1_invest_optimize_all_technologies.py
```diff
@@ -231,8 +231,8 @@ def main(optimize=True):
 
     electricity_bus = solph.views.node(results, "electricity")
 
-    meta_results = solph.processing.meta_results(om)
-    pp.pprint(meta_results)
+    # meta_results = solph.processing.meta_results(om)
+    # pp.pprint(meta_results)
```
p-snft: I'd consider "commenting out" stuff pre-version-control habit. Please never commit something like this.

```suggestion
```

src/oemof/solph/_models.py
```diff
@@ -471,7 +533,20 @@ def solve(
...
             cmdline_options = {}
-        opt = SolverFactory(solver, solver_io=solver_io)
+        if solver == "highs":
+            return self.solve_highs(
+                solver_io=solver_io,
+                allow_nonoptimal=allow_nonoptimal,
+                solve_kwargs=solve_kwargs,
+                cmdline_options=cmdline_options,
+            )
+
```
p-snft, line 592: We should be more explicit here that we branch into a "traditional" code path and one for highs. (The current implementation just continues if it falls through the check.)

```suggestion
        if solver == "highs":
            return self._solve_highs(
                solver_io=solver_io,
                allow_nonoptimal=allow_nonoptimal,
                solve_kwargs=solve_kwargs,
                cmdline_options=cmdline_options,
            )
        else:
            return self._solve_traditional(
                solver=solver,
                solver_io=solver_io,
                allow_nonoptimal=allow_nonoptimal,
                solve_kwargs=solve_kwargs,
                cmdline_options=cmdline_options,
            )

    def _solve_traditional(
        self,
        solver,
        solver_io,
        allow_nonoptimal,
        solve_kwargs,
        cmdline_options,
    ):
```

tests/test_models.py
```diff
@@ -16,59 +16,229 @@
 from pyomo.opt.results import SolverResults
 
 from oemof import solph
+from oemof.solph._results import Results
```
p-snft: In tests, I'd try not to unnecessarily do underscored imports. In this case using `solph.Results` below should be the cleanest way. (The whole test is very verbose with modules, so I even think that `from oemof.solph import Results` would be kind of inconsistent.)

tests/test_models.py
```diff
@@ -16,59 +16,229 @@
...
 
 
-def test_infeasible_model():
+# ---------------------------------------------------------------------------
+# Shared energy-system factories
+# ---------------------------------------------------------------------------
+
+
+def _make_infeasible_es():
+    """Source capacity (4) < sink demand (5) → infeasible."""
```
p-snft: I know, we have utf8. Still, I'd prefer to only use "easy to type" characters.

```suggestion
    """Source capacity (4) < sink demand (5) -> infeasible."""
```

(Same comment applies below.)

src/oemof/solph/_models.py
```diff
@@ -436,6 +437,67 @@ def results(self):
...
+        self,
+        solver_io="lp",
+        allow_nonoptimal=False,
+        solve_kwargs=None,
+        cmdline_options=None,
+    ):
+        if solve_kwargs is None:
+            solve_kwargs = {}
+        if cmdline_options is None:
+            cmdline_options = {}
```
p-snft: In my opinion, `_solve_highs` should not be called by the user. So:

```suggestion
    def _solve_highs(
        self,
        solver_io,
        allow_nonoptimal,
        solve_kwargs,
        cmdline_options,
    ):
```

(If there are still points to discuss, I am also willing to approve with some technical dept if we include an `ExperimentalFeatureWarning` here.)

src/oemof/solph/_models.py
```diff
@@ -471,7 +533,20 @@ def solve(
...
+        if solver == "highs":
+            return self.solve_highs(
+                solver_io=solver_io,
+                allow_nonoptimal=allow_nonoptimal,
+                solve_kwargs=solve_kwargs,
+                cmdline_options=cmdline_options,
+            )
+
+        if "appsi" in solver:
+            solver_io = {}
```
p-snft: This is untested.

p-snft (changes requested):
Neat! I am astonished, how fast this was developed.

Still, there are (minor) quality issues. Most of them are commented inline, except one:
While this runs without problems on my Linux machines, I get "Is this a writeable TextIOBase object?" and a `WinError` when trying this using Windows. (If this is a general compatibility problem, documenting it might be enough. If there is an easy solution, it should be documented.)

At last, don't forget to add this to the changelog.

src/oemof/solph/_models.py
fwitte: It was a relict of the initial commit here, has been removed

tests/test_models.py
fwitte: Applied

fwitte:
Interesting question if this works: should highspy ship as standard dependency and be recommended as default solver? Would make installation very much easier.

fwitte:
> Neat! I am astonished, how fast this was developed.
> 
> Still, there are (minor) quality issues. Most of them are commented inline, except one: While this runs without problems on my Linux machines, I get "Is this a writeable TextIOBase object?" and a `WinError` when trying this using Windows. (If this is a general compatibility problem, documenting it might be enough. If there is an easy solution, it should be documented.)
> 
> At last, don't forget to add this to the changelog.

Can't reproduce the error on my windows machine...

fwitte:
Claude however was able to reproduce it...

So I guess some kind of intermediate fix here and a bug report to pyomo?

## HiGHS Windows Compatibility Issue

### How to Reproduce

Run any `model.solve(solver="highs")` call using `pythonw.exe` (the
console-less Windows Python interpreter) instead of `python.exe`:

```
.venv\Scripts\pythonw.exe your_script.py
```

This simulates any Windows environment without a real console window:
GUI applications, scripts launched via double-click with `.pyw` extension,
some IDE configurations, etc.

### Error
```
AttributeError: 'NoneType' object has no attribute 'flush'
File "pyomo/common/tee.py", line 381, in _enter_impl
stream.flush()
```


### Root Cause

In console-less Windows processes, Python sets `sys.stdout` and `sys.stderr`
to `None` (there is no console to write to). This is standard Python/Windows
behavior.

`pyomo.contrib.appsi.solvers.Highs.solve()` — which our `_solve_highs()`
calls — wraps the entire solve in a `capture_output(capture_fd=True)` context
manager. That context manager unconditionally calls `stream.flush()` on both
`sys.stdout` and `sys.stderr` without a `None` guard, crashing before the
solver even starts.

**This is a pyomo bug**, not a HiGHS or oemof-solph bug. The traditional
`SolverFactory` path (used for CBC, GLPK, etc.) does not have this problem.

### Proposed Fix (in oemof-solph)

Before calling `opt.solve(self)` in `_solve_highs()`, temporarily replace
`None` streams with a `StringIO` sink and restore them afterward. This is ~5
lines, fully reversible, and has zero impact on normal console environments.

### Affected Environments

- Scripts run with `pythonw.exe`
- `.pyw` files launched from Windows Explorer
- Possibly some IDE embedded consoles that redirect or suppress
  `sys.stdout`/`sys.stderr`

uvchik:
> done, we should add some tests I guess, and some explanations as to how exactly that implementation works

I started to review and adapt your draft, but in the meantime, you (or Claude) had written so much new code that I couldn't keep up anymore. That's why I handed the task over to you, @fwitte, since my approach was to make only limited code changes at first, also with a view to future maintenance. That's also why I removed the duals, because they aren't currently included in the Results object at all. For me, that would mean creating a new PR to make the duals and rc available for all solvers. My approach can now be found in #1308.

p-snft:
> **This is a pyomo bug**, not a HiGHS or oemof-solph bug. The traditional `SolverFactory` path (used for CBC, GLPK, etc.) does not have this problem.

Would you please link the Pyomo bug report? (If there is none, we should report it. And as you name the cause, it would make sense to directly suggest a fix.)

p-snft:
> I get "Is this a writeable TextIOBase object?" and a `WinError` when trying this using Windows.

I checked different combinations of opating systems and pyomo versions. The problem only occured using pyomo<6.10 on Windows. So, the problem is fixed when setting `pyomo>=6.10.0` as the lower bound for pyomo.

uvchik:
Closed in favour of PR #1308

PR #1308, opened 2026-06-19
Add highs solver (altternative)

### LINKED ISSUE #855: Support of HiGHS solver

uvchik:
This an alternative approach to #1307 

* I tried to reduce the additional code to keep maintainance low.
* I revised the `check_example` code, to check all examples with "cbc" and "highs"

Most example work with both solver. As there is a lot of movement in #1307 I stop by now.

investment_with_minimal_invest.minimal_invest.py failed
storage_investment.v2_invest_optimize_only_gas_and_storage.py failed
storage_investment.v3_invest_optimize_only_storage_with_fossil_share.py failed
storage_investment.v4_invest_optimize_all_technologies_with_fossil_share.py failed
time_series_aggragation.invest_optimize_all_technologies_using_mp_and_tsam.py failed

5 of 33 examples failed with highs!
******* TEST RESULTS - summary ***********************************
      total failed
cbc      33      0
highs    33      5


fwitte:
@uvchik, I fixed in the example scripts some left over hard-coded solver names and pushed the changes. Currently, there are some models still failing for me because the meta_results are called, which are not compatible with the way highs results are stored. This would be a fix, that let's highs solver return meta results in a way that is compatible with how SolverFactory API solvers did:

```python
def meta_results(om, undefined=False):
    """
    Fetch some metadata from the Solver. Feel free to add more keys.

    Valid keys of the resulting dictionary are: 'objective', 'problem',
    'solver'.

    om : oemof.solph.Model
        A solved Model.
    undefined : bool
        By default (False) only defined keys can be found in the dictionary.
        Set to True to get also the undefined keys.

    Returns
    -------
    dict
    """
    meta_res = {"objective": om.objective()}
    
    # part for highs, here the es.results is a dictionary
    if isinstance(om.es.results, dict):
        sr = om.es.results
        lb = sr.get("best_objective_bound")
        ub = sr.get("best_feasible_objective")
        denom = abs(ub + lb) / 2 if (ub is not None and lb is not None) else None
        gap = abs(ub - lb) / denom if denom else 0.0
        meta_res["problem"] = {
            "Lower bound": lb,
            "Upper bound": ub,
            "MIPGap": gap,
        }
        meta_res["solver"] = {
            "Termination condition": sr.get("termination_condition"),
            "Wall time": sr.get("wallclock_time"),
        }
        return meta_res
    
    for k1 in ["Problem", "Solver"]:
        k1 = k1.lower()
        meta_res[k1] = {}
        for k2, v2 in om.es.results[k1][0].items():
            try:
                if str(om.es.results[k1][0][k2]) == "<undefined>":
                    if undefined:
                        meta_res[k1][k2] = str(om.es.results[k1][0][k2])
                else:
                    meta_res[k1][k2] = om.es.results[k1][0][k2]
            except TypeError:
                if undefined:
                    msg = "Cannot fetch meta results of type {0}"
                    meta_res[k1][k2] = msg.format(
                        type(om.es.results[k1][0][k2])
                    )
    meta_res["problem"]["MIPGap"] = abs(
        meta_res["problem"]["Upper bound"] - meta_res["problem"]["Lower bound"]
    ) / (
        abs(
            meta_res["problem"]["Upper bound"]
            + meta_res["problem"]["Lower bound"]
        )
        / 2
    )
    return meta_res
```

The fix for the windows compatibility @p-snft mentioned could be pulled from my last commit in #1307.
And, to make the examples with the duals work again (tuple_as_labels and dual_variable_example) you would need to bring back the snippet of code that was removed in your earlier commit.

uvchik:
@fwitte I like your code snippet above, even though, I’d view it as a temporary solution whilst we work towards the new implementation of meta results in the ‘Results’ class. Nevertheless, it does provide a good basis for the new implementation. I would start a new PR, when we merged this one.

uvchik:
...and thank you for the fixes in the examples ❤️

p-snft:
I like the following test in the other PR: https://github.com/oemof/oemof-solph/pull/1307/changes#diff-0adb667cd397bd56edce10eec11e1b10821740a10d72bd148b09645fc9108968R269

uvchik:
I think this PR is ready to merge.

p-snft (approved):
There were some minor issues that I just fixed:
1. You forgot the changelog entry.
2. You deactivated check-manifest instead of fixing the manifest.
3. You had some German text printed. 

In total, this PR is fine. Thanks for adding HiGHS. :)

PR #1311, opened 2026-07-02
Make Link parent Node of Converters

p-snft:
The `Link` is actually equivalent to two `Converter`s. Having subnodes, we can now implement it this way without loosing any functionality. For those, who want it more visually, here is the result plot of the transshipment example:

old_link

bundled_converters

PS: I came up with this idea while answering the [question at openmod, when Bus or Converter should be used](https://forum.openmod.org/t/bus-connection-patterns-in-oemof-solph-bus-link-and-converter/5831).

### DIFF
--- a/docs/whatsnew/v0-6-5.rst
+++ b/docs/whatsnew/v0-6-5.rst
@@ -37,6 +37,8 @@ Other changes
   by convention. Thus, it can now be reliably cast; not to an array but to
   a scalar float. We expect no effect for end users.
 * Updated respective examples and tutorials to use current TSAM API (v3).
+* The component ``Link`` now uses the nested Node capability and is internally
+  represented by two ``Converter`` s.
 
 Contributors
 ############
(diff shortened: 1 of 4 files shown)


Bachibouzouk (approved):
To me it is great if the Node-subnode structure can help delete code (less to maintain ^^) especially if this was just here to avoid breaking changes :)

Two question came to my mind as I looked the code, they might go beyond this PR but I want to have them written somewhere :
- Isn't there a way to change the displayed label of the link ? 
- How should oemof.visio diplay a link ? As a simple arrow with two heads between two busses? As a component? As two arrows with a single head connecting two busses?

p-snft:
> Isn't there a way to change the displayed label of the link ?

That would be a postprocessing hook and is on the to-do-list of @gnn. (I won't merge this before we the results look different.)

> How should oemof.visio diplay a link ? As a simple arrow with two heads between two busses? As a component? As two arrows with a single head connecting two busses?

I think, visio should use the implicit Flows of the latest network release and have no special handling for this particular component.

PR #1315, opened 2026-07-07
Fix casting _FakeSequence to number

### LINKED ISSUE #1314: float(_FakeSequence) might return non-floats
We just assumed that values stored in a `_FakeSequence` are floats, but this is not always the case.

https://github.com/oemof/oemof-solph/blob/205c59fd1a0e4724a1d458acafe82dd4b5c096dd/src/oemof/solph/_plumbing.py#L205-L206

p-snft:
We just assumed that values stored in a `_FakeSequence` are floats, but this is not always the case. Fortunately, this was easy to fix.

### DIFF
--- a/src/oemof/solph/_plumbing.py
+++ b/src/oemof/solph/_plumbing.py
@@ -14,6 +14,7 @@ SPDX-License-Identifier: MIT
 import warnings
 from collections import UserDict
 from collections import abc
+from numbers import Number
 
 import numpy as np
 
@@ -180,7 +181,7 @@ class _FakeSequence:
     42
     """
 
-    def __init__(self, value):
+    def __init__(self, value: Number):
         self._value = value
 
     def __getitem__(self, i):
--- a/src/oemof/solph/_plumbing.py
+++ b/src/oemof/solph/_plumbing.py
@@ -203,7 +203,10 @@ class _FakeSequence:
         return f"[{self._value}, {self._value}, ..., {self._value}]"
 
     def __float__(self):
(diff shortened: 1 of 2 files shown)


src/oemof/solph/_plumbing.py
```diff
@@ -203,7 +203,10 @@ def __repr__(self):
         return f"[{self._value}, {self._value}, ..., {self._value}]"
 
     def __float__(self):
-        return self._value
+        return float(self._value)
+
+    def __int__(self):
+        return int(self._value)
```
gnn, line 210: I'd use `int(float(self))` here. Adds a bit of overhead, but establishes the float representation as a numeric "reference" and reduces the sites that have to be changed, should the way `._value` is stored ever change.
But that's just a minor suggestion that you don't have to follow. Regardless of the decision, changes look good. :)

gnn (approved):

src/oemof/solph/_plumbing.py
p-snft: I see the point, but in my opinion, it's easier to understand the way it is. (Also, if float was the reference, why don't we cast to float when setting the value?)

One additional thought: You can initalise using strings. This currently leads to inconsistent behavour and some operations fail. I think this might even be better then just inconsistency with active prevention of the failure.
```Python
seq = _FakeSequence("5.5")
float(seq)
> 5.5
int(seq)
> (fails)
seq[5]
> "5.5" 
```

I just added a type hint to make clear what we expect.

Issue #946, opened 2023-06-15
limit_active_flow_count_by_keyword does not work with investments


p-snft:
As of solph v0.5, `NonConvex` and `Investment` can be combined. However, `limit_active_flow_count_by_keyword(...)` just iterates over flows that do not have an investment. This is unexpected behaviour.

In my opinion, `INVEST_NON_CONVEX_FLOWS` should be a subset of `NONCONVEX_FLOWS` but it seems it is not.

(See https://forum.openmod.org/t/the-constraint-additional-investment-flow-limit-not-working-with-investnonconvexflow/4054 for more information on how this started.)

jfreissmann:
Hey @p-snft , I stumbled upon this issue idependently when trying to disallow simultaneous charing and discharging a thermal storage, which can either have a defined fixed or optimized capacity depending on user selection, for [owp_milp_optimization](https://github.com/maltefritz/owp_milp_optimization) using one of the flow count limit constraints. I think, my TES model as well as the one described in the openmod thread you linked to are very reasonable and possibly common modeling tasks. Furthermore, while the `limit_active_flow_count` constraint at least raises a somewhat cryptic `KeyError` (as the intended Flow is not part of the `NonConvexFlowBlock` index set), the `limit_active_flow_count_by_keyword` constraint silently drops the intended Flow and the user gets no feedback at all.

The simplest fix would likely be the suggested approach in the openmod thread, which was iterating over both sets containing `.status` variables. Was the reason this was not implemented that a more fundamental design adjustment would be preferred (e.g., your suggestion of `INVEST_NON_CONVEX_FLOWS` being a subset of `NONCONVEX_FLOWS`)? Are there any other constraints affected as well? In my opinion, there should be a warning/error implemented soon to avoid the silent ingnoring of the constraint. Maybe the simple fix of iterating both sets would also be a best for now, instead of leaving it as is and waiting for a future "optimal" design approach.

PR #1319, opened 2026-07-24
Fix flow count limit for invest and nonconvex flows

jfreissmann:
This PR aims at fixing issue #946 by also iterating through `model.InvestNonConvexFlowBlock.INVEST_NON_CONVEX_FLOWS` when building the `limit_active_flow_count` and `limit_active_flow_count_by_keyword` constraints. I also adjusted the respective doc strings and added tests for solely nonconvex flows with investment and a mixed case with both purely nonconvex flows and nonconvex flows with investment.

This PR can be deleted, if another approach is intended in order to fix the current issue with these constraints not working for nonconvex flows with investment, e.g., when the set definition of `NON_CONVEX_FLOWS` & `INVEST_NON_CONVEX_FLOWS` fundamently changes.

### DIFF
--- a/src/oemof/solph/constraints/flow_count_limit.py
+++ b/src/oemof/solph/constraints/flow_count_limit.py
@@ -136,8 +136,20 @@ def limit_active_flow_count_by_keyword(
     --------
     limit_active_flow_count
     """
+    nonconvex_flows = []
+    if hasattr(
+        model.NonConvexFlowBlock, "FIXED_CAPACITY_NONCONVEX_FLOWS"
+    ):
+        nonconvex_flows += list(
+            model.NonConvexFlowBlock.FIXED_CAPACITY_NONCONVEX_FLOWS
+        )
+    if hasattr(model.InvestNonConvexFlowBlock, "INVEST_NON_CONVEX_FLOWS"):
+        nonconvex_flows += list(
+            model.InvestNonConvexFlowBlock.INVEST_NON_CONVEX_FLOWS
+        )
+
     flows = []
-    for i, o in model.NonConvexFlowBlock.FIXED_CAPACITY_NONCONVEX_FLOWS:
+    for i, o in nonconvex_flows:
         if keyword in model.flows[i, o].custom_properties:
             flows.append((i, o))
 
--- a/src/oemof/solph/constraints/flow_count_limit.py
+++ b/src/oemof/solph/constraints/flow_count_limit.py
@@ -88,10 +88,10 @@ def limit_active_flow_count(
         for ts in m.TIMESTEPS:
             lhs = 0
             for fi, fo in flows:
-                if (fi, fo) in invest_flows:
-                    lhs += m.InvestNonConvexFlowBlock.status[fi, fo, ts]
-                else:
+                if (fi, fo) not in invest_flows:
                     lhs += m.NonConvexFlowBlock.status[fi, fo, ts]
+                else:
+                    lhs += m.InvestNonConvexFlowBlock.status[fi, fo, ts]
             rhs = getattr(model, constraint_name)[ts]
             expr = lhs == rhs
             if expr is not True:
--- a/src/oemof/solph/constraints/flow_count_limit.py
+++ b/src/oemof/solph/constraints/flow_count_limit.py
@@ -79,10 +79,17 @@ def limit_active_flow_count(
     attrname_constraint = constraint_name + "_constraint"
 
     def _flow_count_rule(m):
+        if hasattr(m.InvestNonConvexFlowBlock, "INVEST_NON_CONVEX_FLOWS"):
+            invest_flows = m.InvestNonConvexFlowBlock.INVEST_NON_CONVEX_FLOWS
+        else:
+            invest_flows = []
         for ts in m.TIMESTEPS:
-            lhs = sum(
-                m.NonConvexFlowBlock.status[fi, fo, ts] for fi, fo in flows
-            )
+            lhs = 0
+            for fi, fo in flows:
+                if (fi, fo) in invest_flows:
+                    lhs += m.InvestNonConvexFlowBlock.status[fi, fo, ts]
+                else:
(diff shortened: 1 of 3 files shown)


src/oemof/solph/constraints/flow_count_limit.py
```diff
@@ -79,10 +81,17 @@ def limit_active_flow_count(
...
         for ts in m.TIMESTEPS:
-            lhs = sum(
-                m.NonConvexFlowBlock.status[fi, fo, ts] for fi, fo in flows
-            )
+            lhs = 0
+            for fi, fo in flows:
+                if (fi, fo) in invest_flows:
+                    lhs += m.InvestNonConvexFlowBlock.status[fi, fo, ts]
+                else:
+                    lhs += m.NonConvexFlowBlock.status[fi, fo, ts]
```
p-snft, line 92: I would invert the checks here, so that the if/else has not to be evaluated that often.

p-snft (changes requested):
I think it's a good idea to have this hotfix. Still, there at least a change log entry is missing. (Please also consider my inline comment.)

src/oemof/solph/constraints/flow_count_limit.py
jfreissmann: Did I understand you correctly that the condition should be ``if (fi, fo) not in invest_flows`` and handle the ``NonConvexFlowBlock`` case first, because it is more likely that a flow is just nonconvex?

p-snft: Sorry, my formulation was misleading. I am refining this using a concrete suggestion.

src/oemof/solph/constraints/flow_count_limit.py
```diff
@@ -79,10 +81,17 @@ def limit_active_flow_count(
     attrname_constraint = constraint_name + "_constraint"
 
     def _flow_count_rule(m):
+        if hasattr(m.InvestNonConvexFlowBlock, "INVEST_NON_CONVEX_FLOWS"):
+            invest_flows = m.InvestNonConvexFlowBlock.INVEST_NON_CONVEX_FLOWS
+        else:
+            invest_flows = []
```
p-snft, line 87:
```suggestion
```
The set should always be there if the Block is present (and you do not need a local copy).

src/oemof/solph/constraints/flow_count_limit.py
```diff
@@ -79,10 +81,17 @@ def limit_active_flow_count(
...
         for ts in m.TIMESTEPS:
-            lhs = sum(
-                m.NonConvexFlowBlock.status[fi, fo, ts] for fi, fo in flows
-            )
+            lhs = 0
+            for fi, fo in flows:
+                if (fi, fo) not in invest_flows:
+                    lhs += m.NonConvexFlowBlock.status[fi, fo, ts]
+                else:
+                    lhs += m.InvestNonConvexFlowBlock.status[fi, fo, ts]
```
p-snft, line 94:
```suggestion
            lhs = 0
            for fi, fo in m.NonConvexFlowBlock.FIXED_CAPACITY_NONCONVEX_FLOWS:
                lhs += m.NonConvexFlowBlock.status[fi, fo, ts]
            for fi, fo in m.InvestNonConvexFlowBlock.INVEST_NON_CONVEX_FLOWS:
                lhs += m.InvestNonConvexFlowBlock.status[fi, fo, ts]
```

src/oemof/solph/constraints/flow_count_limit.py
```diff
@@ -129,8 +139,20 @@ def limit_active_flow_count_by_keyword(
...
+    if hasattr(
+        model.NonConvexFlowBlock, "FIXED_CAPACITY_NONCONVEX_FLOWS"
+    ):
+        nonconvex_flows += list(
+            model.NonConvexFlowBlock.FIXED_CAPACITY_NONCONVEX_FLOWS
+        )
+    if hasattr(model.InvestNonConvexFlowBlock, "INVEST_NON_CONVEX_FLOWS"):
+        nonconvex_flows += list(
+            model.InvestNonConvexFlowBlock.INVEST_NON_CONVEX_FLOWS
+        )
```
p-snft, line 152:
```suggestion
     nonconvex_flows += list(
        model.NonConvexFlowBlock.FIXED_CAPACITY_NONCONVEX_FLOWS
    )
    nonconvex_flows += list(
        model.InvestNonConvexFlowBlock.INVEST_NON_CONVEX_FLOWS
    )
```

Both should always be there. (But there is no unified set, so creating the unification makes sense.)

p-snft (commented):
Here are some concrete suggestions. I hope this is more precise than what I did before.

src/oemof/solph/constraints/flow_count_limit.py
jfreissmann: Thank you for the more in-depth suggestions. I will finish them shortly.

However, when I tried to implement the unified set without the ``hasattr`` guard, it seems that it can fail with an ``AttributeError``, because, while the pyomo blocks always exists, the flow sets are not created if the ``group`` is empty (``_create_sets(group=None)`` returns early). I created a minimal reproduceable example in the following github gist:

https://gist.github.com/jfreissmann/53f160c3f48463ea47e33573da92551d

Let me know, if I am missing anything.

p-snft: My bad. I believed we would create empty sets in that case - which would be cleaner. I suggest to change `XYZBlock._create(group)` from the early return to something like:
```
    def _create(self, group=None):
        if group is None:
            self._create_sets([])
        else:
            self._create_sets(group)
            self._create_variables()
            self._create_constraints()
```

p-snft (approved):
I just add the proposed changes upon merge. Thanks.

jfreissmann:
Sorry @p-snft, you were quicker than me. Thank you for finishing up this PR.

PR #1323, opened 2026-08-27
Add meta data to results object

uvchik:
Here’s a draft for sharing the meta results.

What I’d like to see is:

- A "meta" key for the meta results as a pandas series
- A "solver_results" key for the objects—for people who want to dig around in them—but I’d prefer to hide that one.

Right now, though, the keywords are being phased out as single keys. I don't like that.

In general, we can still discuss the keywords I chose for the meta results—I didn't spend much time on them, so there might still be some mapping errors. My first priority is the concept itself, and only then do I focus on the keywords within the meta results.

### DIFF
--- a/examples/simple_dispatch/temporary_meta_example.py
+++ b/examples/simple_dispatch/temporary_meta_example.py
@@ -1,0 +1,163 @@
+import os
+import pprint
+
+import matplotlib.pyplot as plt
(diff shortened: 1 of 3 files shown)


uvchik:
In addition, we should talk about backward compatibility, but I would wait until we've agreed on the other issues.

PR #1324, opened 2026-08-31
Make ExtractionTurbineCHP a generic component

uvchik:
In the long run, this new component could replace the "ExtractionTurbineCHP" class.

Basically, it follows a more generic approach in which two output flows are dependent on one another. If one changes, the other changes too. The dependencies are defined by two boundary conditions, each of which specifies the ratio of the flows to the input flow and thus to one another.

grafik

In the normal state, the two points must be different. If this is not the case, an error is raised. This applies to both individual values and time series. This makes sense, since otherwise the Converter could just as well be used. In that case, only the minimum necessary constraints are created. For the rare case where time series are used and the values happen to be equal at only a few individual points in time, the parameter "allow_equal_states" can be set to "True". In that case, it is checked at every point in time whether the points are equal, and only then are the additional constraints created. However, this check of the individual values is only carried out if "allow_equal_states" is set to "True" to save time in the "normal" run.

* I would suggest replacing ExtractionTurbineCHP and only keeping the Extraction Turbine in the documentation, to help people.
* Of course, we should still discuss the name of the new component.

If you generally like the concept, I will add a FutureWarning, error messages and tests, and finish documenting the component.

### DIFF
--- a/examples/variable_chp/variable_chp.py
+++ b/examples/variable_chp/variable_chp.py
@@ -110,7 +110,7 @@ def write_lp_file(model):
     model.write(lp_filename, io_options={"symbolic_solver_labels": True})
 
 
-def main(optimize=True, solver="cbc"):
+def main(optimize=True, solver="cbc", new=True):
     # Read data file
     filename = os.path.join(os.getcwd(), "variable_chp.csv")
     try:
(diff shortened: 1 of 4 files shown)


uvchik:
You can use the [variable_chp](https://github.com/oemof/oemof-solph/blob/revision/make-extraction-turbine-chp-a-generic-component/examples/variable_chp/variable_chp.py) example to test the new component.

Issue #1318, opened 2026-07-22
Minimum uptime/downtime: Optional cyclical enforcement


jfreissmann:
Minimum uptime or downtime of the ``NonConvex`` class are great options to avoid on-off cycling of e.g. Converters. The current implementation enforces these constraints within the assessment period, which makes sense for one-off simulations. However, this can lead to a "bypass" of the constraints, if the simulation period is supposed to be a representative one, which "repeats" itself (e.g. when calculating LCOH from annuities and repeating constant operating cost from the rep. simulation period). This can be marginal in case of a year, but if you're simulating, e.g., days or 72h periods (see also TSA), the effect can be quite consequential. In any case, allowing for "wrapping/cyclic" minimum uptime or downtime should only be optional.

Maybe, a not too laborious approach could be working with modular division. See this example for minimum uptime (mu), for each timestep (t) in the assessment period (T):

$$(status[t] - status[(t-1) \mod T]) \cdot mu  \leq  \sum_{n=0}^{mu-1} status[(t+n) \mod T]$$

What do you think? I know, this also needs to consider ``initial_status`` parametrization.

p-snft:
We face the same questions with the storage (solved) and ramping constraints (not addressed, yet). As we have a global time axis anyway, I'd suggest to solve it globally. So, instead of having using `status[t+1]` we could have a global mapping like `status[EnergySystem.timestep_after(t)]`. This would also simplify the TSAM code.

Issue #1102, opened 2024-08-14
Initial_status of Flow ignored by default


p-snft:
For a flow with `solph.NonConvex(initial_status=0)`, this information is ignored, as `minimum_downtime=0` by default. I would expect the status to be 0 in the initial time step.

PR #1325, opened 2026-09-01
Enforce minimum_uptime/minimum_downtime over the first time steps

dylanpulver:
`minimum_uptime` / `minimum_downtime` are silently not enforced over the first time
steps, so the solver returns schedules that violate them. Both rules in
`src/oemof/solph/flows/_shared.py` were built only for `t > first_flexible_timestep`
(`_shared.py:217` and `:271`). That value is not a property of either constraint:
`NonConvex.__init__`
(`_options.py:317-320`) sets it to `minimum_downtime[0]` when `initial_status == 0`
and to `minimum_uptime[0]` otherwise. It says how many initial `status` values
`NonConvexFlowBlock` fixes, and it comes from only one of the two minimum times — so
when they differ, the rule for the *other* one is skipped over time steps where a
real transition can happen. Neither rule covered `t = 0` at all, although
`initial_status` already supplies the pre-horizon status to the startup and shutdown
rules just below (`_shared.py:317` and `:345`).

Two runs on current `dev` (10 h, `nominal_capacity=10`, `minimum=0.5`, CBC 2.10.13):

| `NonConvex` | `variable_costs` | `dev` status / obj. | patched |
|---|---|---|---|
| `initial_status=0, minimum_downtime=2, minimum_uptime=4` | `[1,1,-10,1,...]` | `[0,0,1,0,...]`, -100.0 | `[0,0,1,1,1,1,0,...]`, -85.0 |
| `initial_status=1, minimum_uptime=2, minimum_downtime=4` | `[1,1,10,-10,1,...]` | `[1,1,0,1,1,0,...]`, -85.0 | `[1,1,1,1,0,...]`, -40.0 |

One time step of uptime against `minimum_uptime=4`, and one of downtime against
`minimum_downtime=4`.

### The change

Build both rules for every time step and take the status before the horizon from
`initial_status`, the convention the startup/shutdown rules already used. That is
factored into `_previous_status()` and reused in all four rules, so the constraint
rules no longer read `first_flexible_timestep`. Over the fixed time steps the added
rows hold for any binary `status`, so the documented `initial_status` behaviour is
unchanged. Both docstrings are updated to match.

Worth noting: the obvious smaller fix does not hold. Relaxing the guard to
`first_flexible_timestep <= t` raises `KeyError: status[..., -1]` at `t = 0`, and
guarding that as `max(1, first_flexible_timestep) <= t` closes the mismatched-minimums
hole but still leaves `t = 0` unconstrained.

Partly overlaps #1102 — a `minimum_uptime`-only flow with `initial_status=0` now
honours the uptime when it starts in the first time step, without changing any
default. Not #1318, which asks for optional cyclical wrap-around; this is only about
the start of the horizon.

Four tests in `tests/test_flows/test_non_convex_flow.py` asserting literal flow
series. All four fail on unpatched `dev`; the two `t = 0` ones still fail against the
guard-only variant above. Full suite 286 passed / 2 skipped before, 290 /
2 after, nothing else changed. `flake8 src tests`, `isort --check-only --profile
black`, `black --check` on the touched files and `sphinx-build -W -b html` are clean.

Changelog note and name added to `docs/whatsnew/v0-6-5.rst` — CONTRIBUTING points at
`next_version.rst`, which no longer exists, so please redirect me if that is wrong.
This can change results for existing models combining `initial_status` with
`minimum_uptime`/`minimum_downtime`, in the direction of the documented behaviour;
flagged as such in the changelog.

Written with AI assistance; the schedules above are solver output, not inference.

### DIFF
--- a/src/oemof/solph/flows/_shared.py
+++ b/src/oemof/solph/flows/_shared.py
@@ -189,6 +189,19 @@ def variables_for_non_convex_flows(block):
         )
 
 
+def _previous_status(block, i, o, t):
+    """Status of a flow in the time step before `t`.
+
+    Before the first time step of the optimisation horizon, the flow is
+    assumed to have been in its `initial_status`. This is the same
+    convention that the startup and shutdown constraints use.
+    """
+    m = block.parent_block()
+    if t > m.TIMESTEPS.at(1):
+        return block.status[i, o, t - 1]
+    return m.flows[i, o].nonconvex.initial_status
+
+
 def _min_downtime_constraint(block):
     r"""
     .. math::
@@ -197,15 +210,11 @@ def _min_downtime_constraint(block):
         \leq t_{down,minimum} \
         - \sum_{n=0}^{t_{down,minimum}-1} Y_{status}(t+n) \\
         \forall t \in \textrm{TIMESTEPS} | \\
-        t \neq \{0..t_{down,minimum}\} \cup \
-        \{t\_max-t_{down,minimum}..t\_max\} , \\
+        t \neq t\_max , \\
         \forall (i,o) \in \textrm{MINDOWNTIMEFLOWS}.
-        \\ \\
-        Y_{status}(t) = Y_{status,0} \\
-        \forall t \in \textrm{TIMESTEPS} | \\
-        t = \{0..t_{down,minimum}\} \cup \
-        \{t\_max-t_{down,minimum}..t\_max\} , \\
-        \forall (i,o) \in \textrm{MINDOWNTIMEFLOWS}.
+
+    where :math:`Y_{status}(-1) := Y_{status,0}`, the `initial_status`
+    of the flow.
     """
     m = block.parent_block()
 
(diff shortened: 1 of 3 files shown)


src/oemof/solph/flows/_shared.py
```diff
@@ -251,32 +256,24 @@ def _min_uptime_constraint(block):
...
     def _min_uptime_rule(_, i, o, t):
         """
         Rule definition for min-uptime constraints of non-convex flows.
         """
-        if (
-            m.flows[i, o].nonconvex.first_flexible_timestep
-            < t
-            < m.TIMESTEPS.at(-1)
-        ):
+        if t < m.TIMESTEPS.at(-1):
```
p-snft: This is not needed any more (could have been removed before but was forgotten).

src/oemof/solph/flows/_shared.py
```diff
@@ -251,32 +256,24 @@ def _min_uptime_constraint(block):
...
             # We have a 2D matrix of constraints,
             # so testing is easier then just calling the rule for valid t.
```
p-snft: Comment does not apply any more.

p-snft (commented):
Hi @dylanpulver, thanks for your suggestion. Moving some logic to a function really makes things easier. I think, our code could be improved further (e.g. making `first_flexible_timestep` obsolete is now in reach), but as this PR also fixes a bug, I won't delay the merge for this. (Still, as it's not super urgent, here are some suggestions.)

About the copyright note and changelog entry:
1. `next_version.rst` in `CONTRIBUTING.rst` is meant to be read as a placeholder for whichever version is planed to be next.
2. If you want, you can also add your name to the `SPDX-FileCopyrightText`. (Not all files have that section, yet, but some do.)
3. There are some (now) unnecessary or even misleading lines that can/should be removed. Please see inline comment. (I just marked the lines once, they are there for both, uptime and downtime.)

p-snft (approved):
Thanks for the fix. The PR looks very good now.

Issue #1128, opened 2024-10-17
storage_level_constraint.py in input constraint missing t+1


Maxhi77:
https://github.com/oemof/oemof-solph/blob/551fa97bc5ebf618d6686207f01ad9c3cf5ed46c/src/oemof/solph/constraints/storage_level.py#L120C1-L137C49

Hello,

in storage_level_constraint.py the input constraint is probably missing a t+1 for the storage content, like done in the output constraint.

Status Quo:`m.GenericStorageBlock.storage_content[storage_component, t]` 
--> Improvement: `m.GenericStorageBlock.storage_content[storage_component, t +1]`

p-snft:
In #1327, I added a comment why this is by design.

PR #1327, opened 2026-09-11
Improve description of storage_level_constraint

p-snft:
There is a nuanced difference in the input and output part of the `storage_level_constraint`. This should be clearer, now.

Solves #1128

### DIFF
--- a/src/oemof/solph/constraints/storage_level.py
+++ b/src/oemof/solph/constraints/storage_level.py
@@ -60,6 +60,9 @@ def storage_level_constraint(
             r"""
             .. math::
                 y_n \le E(t) / E_n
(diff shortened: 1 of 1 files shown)


jfreissmann (approved):
I think, this is a helpful addition for the respective contraint doc strings. In my opinion, the general description of the main constraint could be improved, as it was somewhat hard for me to fully understand what the `storage_level_constraint` does, but that is out of scope for this PR. Finally, do you want to add something to the change log or is the change to small for an entry?

p-snft:
This was meant to be just a quick "fix" showing that there is no error. I will merge and hope for more improvements to come.

PR #1328, opened 2026-09-11
Revision/reuse compliance

p-snft:
This makes the repository adhere to the latest REUSE standard and adds tests (GitHub CI and tox) to validate we don't loose it again.

### DIFF
--- a/docs/_files/code/plots_grids.py
+++ b/docs/_files/code/plots_grids.py
@@ -1,3 +1,11 @@
+# -*- coding: utf-8 -*-
+
+"""
+SPDX-FileCopyrightText: oemof association (oemof e.V.)
+
+SPDX-License-Identifier: MIT
+"""
+
 import os
 import matplotlib.colors as mcolors
 import matplotlib.pyplot as plt
--- a/examples/activity_costs/activity_costs.py
+++ b/examples/activity_costs/activity_costs.py
@@ -34,9 +34,7 @@ This example requires oemof.solph (at least v0.6.4), install by:
 
     pip install oemof.solph>=0.6.4
 
-License
--------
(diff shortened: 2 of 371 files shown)


docs/_files/code/plots_grids.py
lensum: Just for my understanding: Why aren't the contributors named here?

examples/activity_costs/activity_costs.py
lensum: Why is the License Identifier included in the docstring and not in the comment (like the Copyright notice)?

lensum: This applies to multiple examples, but not all of them.

lensum:
I noticed that the `AUTHORS.rst` is missing. But it seems that this is on purpose and that it's auto-generated from the `CITATION.cff?`, correct?

p-snft:
> I noticed that the `AUTHORS.rst` is missing. But it seems that this is on purpose and that it's auto-generated from the `CITATION.cff`, correct?

Yes. We started this some time ago, as `CITATION.cff` contains all the information needed for the `AUTHORS.rst` and syncing the two manually was tedious and unreliable.

jfreissmann:
It took a while to scroll through everything, but it looks good to me. Is there any particular reason some files (e.g. docs/_files/code/plots_grids.py) use "oemof association (oemof e.V.)" instead of "oemof e.V. and contributors" that most others use? After your explanation I would positively review this PR.

p-snft:
The original idea was that for more "simple" scripts, part of which are barely copyrightable, I would only name the association. However, I see now that this is rather confusing. It's now consistent, thank you.

jfreissmann:
Totaly get that. Thank you for harmonizing it. This looks good to me.

jfreissmann (approved):

